PR #3, opened 2021-02-23
To allow disable omnicompl for specific ft.

Shane-XB-Qian:
// when `let g:LSP_24x7_Complete = v:false`,
some compl from lsp maybe not a good way / user like,
or whatever user may like his/her preferred omnifunc which not override by this plugin.

```vim
	let lspServers = [
		\     {
		\	'filetype': ['javascript', 'typescript'],
		\	'omnicompl': v:true,
		\	'path': '/usr/local/bin/typescript-language-server',
		\	'args': ['--stdio']
		\     },
		\     {
		\	'filetype': 'python',
		\	'omnicompl': v:false,
		\	'path': '/usr/local/bin/pyls',
		\	'args': ['--check-parent-process', '-v']
		\     }
		\   ]
```

To allow user can disable omnicompl for specific ft.
- adding a 'omnicompl' option in configuration.
- default (or if not set) 'omnicompl' is true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ To register a LSP server, add the following lines to your .vimrc file:
 		\	'filetype': ['javascript', 'typescript'],
 		\	'path': '/usr/local/bin/typescript-language-server',
 		\	'args': ['--stdio']
-		\     }
+		\     },
 		\     {
 		\	'filetype': 'sh',
 		\	'path': '/usr/local/bin/bash-language-server',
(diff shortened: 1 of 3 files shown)


Shane-XB-Qian:
- refined the commit/pr description.
- missed edited one previous existed 'indent', redraw that one.

// so force pushed a new commit.

Shane-XB-Qian:
thx review and refine.

PR #4, opened 2021-02-25
Let user be able to customize cfg for hoverpopup

Shane-XB-Qian:
- when hover popup showed up:
    * set a 'b:lsphoverpopupid' var as popupid to hover popup
    * set a 'b:lsphovermarkupkind' var as hover markup kind
- user can customize setup according to those if needed if so.

### DIFF
--- a/autoload/handlers.vim
+++ b/autoload/handlers.vim
@@ -117,7 +117,7 @@ def s:processSignaturehelpReply(lspserver: dict<any>, req: dict<any>, reply: dic
   var text = sig.label
   var hllen = 0
   var startcol = 0
-  if sig->has_key('parameters')
+  if sig->has_key('parameters') && result->has_key('activeParameter')
     var params_len = sig.parameters->len()
     if params_len > 0 && result.activeParameter < params_len
       var label = sig.parameters[result.activeParameter].label
@@ -297,7 +297,7 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
     if reply.result.contents->empty()
       return
     endif
-    hoverText->add(reply.result.contents)
+    hoverText->extend(reply.result.contents->split("\n"))
   else
     ErrMsg('Error: Unsupported hover contents (' .. reply.result.contents .. ')')
     return
@@ -865,6 +865,10 @@ def s:processWorkspaceSymbolReply(lspserver: dict<any>, req: dict<any>, reply: d
   var r: dict<dict<number>>
   var symName: string
 
+  if reply.result->type() != v:t_list
+    return
+  endif
+
   for symbol in reply.result
     if !symbol->has_key('location')
       # ignore entries without location information
@@ -1153,7 +1157,7 @@ export def ProcessMessages(lspserver: dict<any>): void
 							.. emsg .. ")")
 	endif
       endif
-    elseif msg->has_key('id')
+    elseif msg->has_key('id') && msg->has_key('method')
       # request message from the server
       lspserver.processRequest(msg)
     elseif msg->has_key('method')


Shane-XB-Qian:
to allow or let user have a interface to setup customized option or highlight etc..
// so far, the hover popup looks a bit poor... :-)

what do you think?

Shane-XB-Qian:
* set hover popup ft as 'lsphover'
// force pushed a new commit.

to be uniformed if so, and then user can dev he/she preferred syntax/options if so if needed,
or/and would not impact anything if just keep the default (kept current hover as it looked like).

// i think this would be better, pls review/refine, or withdraw/ignore if not-like... :-)

Shane-XB-Qian:
* and to fix - hoverReply: type mismatch #6

yegappan:
Thanks for the patch. It will be useful if the plugin, by default, highlights the items in the hover popup. This is not currently implemented. Can you add some help text to describe how a user can use the filetype to highlight the hover popup?

Shane-XB-Qian:
description added. a35dff8

Shane-XB-Qian:
solved code/commit conflict with your last commit, now should be clear and direct to merge if ok. thx.

autoload/handlers.vim
```diff
@@ -302,7 +306,12 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
     ErrMsg('Error: Unsupported hover contents (' .. reply.result.contents .. ')')
     return
   endif
-  hoverText->popup_atcursor({moved: 'word'})
+  hoverWinid = hoverText->popup_atcursor({moved: 'word'})
+  if hoverWinid > 0
+    call setbufvar(winbufnr(hoverWinid), 'lsphoverpopupid', hoverWinid)
```
yegappan, line 311: There is no need to store the window ID as a buffer variable as the user can always get it using the bufwinid() function.

autoload/handlers.vim
```diff
@@ -265,21 +265,25 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
...
+        hoverMarkupKind = 'plaintext'
       elseif reply.result.contents.kind == 'markdown'
-	hoverText = reply.result.contents.value->split("\n")
+        hoverText = reply.result.contents.value->split("\n")
+        hoverMarkupKind = 'markdown'
       else
-	ErrMsg('Error: Unsupported hover contents type (' .. reply.result.contents.kind .. ')')
-	return
+        ErrMsg('Error: Unsupported hover contents type (' .. reply.result.contents.kind .. ')')
+        return
```
yegappan, line 282: Note that the indentation style is to use a mix of spaces and tabs with 'shiftwidth' set to 2, 'tabstop' set to 8 and 'softtabstop' set to 2.

doc/lsp.txt
```diff
@@ -331,6 +331,21 @@ accepts a list of LSP servers with the above information.
 :LspHover		Show the documentation for the symbol under the cursor
 			in a popup window.
 
+			Var 'b:lsphoverpopupid' and 'b:lsphovermarkupkind' as
+			hover popup id and its markup kind ("plaintext" or
+			"markdown" or "") added when hover popup showed.
```
yegappan, line 336: When the hover popup contains markdown text, the markdown characters also will be displayed in the buffer. These characters need to be processed and hidden. So it is not sufficient to just highlight the characters in the hover popup. The hover popup buffer contents needs to be updated based on the markdown syntax.  I am not yet sure how this will work. Instead of expecting another plugin or the user to do this, it will be useful if the LSP plugin itself handles this.

autoload/handlers.vim
Shane-XB-Qian: perhaps yes, but looks that would not easy, i tried to 'au filetype lsphover' to get it (or whatever user would try to get by this way, i guess would), the bufwinid(bufnr()) is not same value like b:lsphoverpopupid.
anyway, by 'b:lsphoverpopupid' looks would be clear and direct...

doc/lsp.txt
Shane-XB-Qian: if it's a concern, actually can just get rid of first-line '```' and last-line '```' (if it is markdown markup kind), then treat it as plaintext at all.
// but i prefer one day someone would  dev a common syntax for 'lsphover' ft, just like others in vim runtime/syntax.

autoload/handlers.vim
Shane-XB-Qian: yea, looks it is the same indentation style of vim source (c/c++) code. i used to set 'ts' to 4, and/or same value like sw/sts...
// i can reset this to your preferred style, or let me do it in next pr? i plan to offer a 'lsp#serverStatus' function to workaround #5.

Shane-XB-Qian: or actually i am a little confused, if looking into other functions or e.g 'lsp#gotoDefinition()'
```vim
  var lspserver: dict<any> = s:lspGetServer(ftype)
  if lspserver->empty()
    ErrMsg('Error: LSP server for "' .. ftype .. '" filetype is not found')
    return
  endif
```
looks many places it's using ts 2 and et yes.

doc/lsp.txt
Shane-XB-Qian: i mean here now i/we just offerred an interface to user, then he/she can customize something as he/she preferred;
as for you said if the LSP plugin itself handled this, i think it's good if can, but now, perhaps we can just give interface as first step.

PR #7, opened 2021-02-27
issue_5

Shane-XB-Qian:
~~adding vimenter for lsp#addFile() too.~~
~~- otherwise lsp server would Not be kicked off if lsp#addServer was added in vimenter.~~
- using 'lsp#serverStatus(&ft)' to check for specific ft, and/or can register into stl as well if needed.


Shane-XB-Qian:
d96da5a5accf307d15aeddefa8994007ae074ff8 is the correct one for this pr.
// another one is for hover, which you have merged yet, sorry if confused.

PR #8, opened 2021-02-28
disable auto showSignature

Shane-XB-Qian:
- if g:LSP_24x7_Complete is false

// auto show signature looks a bit delay, not work smoothly always..
// specially those 'inoremap' would override user's key map, perhaps not a good idea..
// so should be better to enable auto show signature Like auto completion Only IF g:LSP_24x7_Complete is true (default though).

### DIFF
--- a/autoload/handlers.vim
+++ b/autoload/handlers.vim
@@ -25,7 +25,7 @@ def s:processInitializeReply(lspserver: dict<any>, req: dict<any>, reply: dict<a
   # and then setup the below mapping for those buffers.
 
   # map characters that trigger signature help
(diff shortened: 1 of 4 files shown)


yegappan:
Thanks for the patch. As showing the function/method signature is different from the auto-completion functionality, it is a good idea to add another control variable. Can you add a new LSP_Show_Signature variable and use that to enable/disable this functionality instead of LSP_24x7_Complete?

Shane-XB-Qian:
sure, changed to `g:LSP_Show_Signature`, and default is true.

PR #20, opened 2021-11-11
Add : Possibility to echo signature in cmdline

indelog:
This add an `g:LSP_Echo_Signature` option which permit to see signature of a symbol echoed in cmdline rather than displayed in a popup window.

### DIFF
--- a/autoload/handlers.vim
+++ b/autoload/handlers.vim
@@ -120,15 +120,30 @@ def s:processSignaturehelpReply(lspserver: dict<any>, req: dict<any>, reply: dic
   if sig->has_key('parameters') && result->has_key('activeParameter')
     var params_len = sig.parameters->len()
     if params_len > 0 && result.activeParameter < params_len
-      var label = sig.parameters[result.activeParameter].label
+      var label = ''
+      if sig.parameters[result.activeParameter]->has_key('documentation')
+        label = sig.parameters[result.activeParameter].documentation
+      else
+        label = sig.parameters[result.activeParameter].label
+      endif
       hllen = label->len()
       startcol = text->stridx(label)
     endif
   endif
-  var popupID = text->popup_atcursor({moved: 'any'})
-  prop_type_add('signature', {bufnr: popupID->winbufnr(), highlight: 'LineNr'})
-  if hllen > 0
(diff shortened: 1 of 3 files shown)


doc/lsp.txt
```diff
@@ -215,7 +215,9 @@ accepts a list of LSP servers with the above information.
 			The popup is also automatically displayed in insert
 			mode after entering a symbol name followed by a
 			separator (e.g. a opening parenthesis). Unless if: >
-			let g:LSP_Show_Signature = v:false
+			let g:LSP_Show_Signature = v:false. If you set
+			g:LSP_Show_Signature = 1 the signature will be echoed
```
yegappan, line 219: This should be g:LSP_Echo_Signature = v:true?

autoload/handlers.vim
```diff
@@ -120,15 +120,30 @@ def s:processSignaturehelpReply(lspserver: dict<any>, req: dict<any>, reply: dic
...
       hllen = label->len()
       startcol = text->stridx(label)
     endif
   endif
-  var popupID = text->popup_atcursor({moved: 'any'})
-  prop_type_add('signature', {bufnr: popupID->winbufnr(), highlight: 'LineNr'})
-  if hllen > 0
-    prop_add(1, startcol + 1, {bufnr: popupID->winbufnr(), length: hllen, type: 'signature'})
+  if g:LSP_Echo_Signature
+    echon "\r\r"
```
yegappan, line 134: Will this result in a "Press Enter to continue" prompt, if the default 'cmdheight' is used?

autoload/handlers.vim
```diff
@@ -120,15 +120,30 @@ def s:processSignaturehelpReply(lspserver: dict<any>, req: dict<any>, reply: dic
   if sig->has_key('parameters') && result->has_key('activeParameter')
```
yegappan, line 120: Thanks for the patch.

doc/lsp.txt
indelog: Yes it's an error. Il'll fix this.

autoload/handlers.vim
indelog: It seem me no (i use vim8). The purpose of the `echo "\r\r"` it clear cmdline when we call `LspShowSignature` several time in a row (the permit to update the highlight of the current parameter).

PR #21, opened 2021-11-11
Add add function to get error count for the current buffer

indelog:
Can be used to display error count in status bar or with vim-airline.

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -504,7 +504,7 @@ enddef
 
 # get the count of error in the current buffer
 def lsp#errorCount():dict<number>
-  var res = {'E': 0, 'W': 0, 'I': 0, 'H': 0}
+  var res = {'Error': 0, 'Warn': 0, 'Info': 0, 'Hint': 0}
   var ftype = &filetype
   if ftype == ''
     return res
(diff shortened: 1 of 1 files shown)


autoload/lsp.vim
```diff
@@ -502,6 +502,42 @@ def lsp#showCurrentDiag()
   endif
 enddef
 
+# get the count of error in the current buffer
+def lsp#errorCount():dict<number>
+  var res = {'E': 0, 'W': 0, 'I': 0, 'H': 0}
```
yegappan, line 507: To make this readable, can you expand this to 'Error', 'Warn', 'Info' and 'Hint'?

autoload/lsp.vim
```diff
@@ -502,6 +502,42 @@ def lsp#showCurrentDiag()
...
+  var ftype = &filetype
+  if ftype == ''
+    return res
+  endif
+
+  var lspserver: dict<any> = s:lspGetServer(ftype)
+  if lspserver->empty()
+    return res
+  endif
+  if !lspserver.running
```
yegappan, line 517: Can you combine these two if statements into one?

```
if lspserver->empty() || !lspserver.running
  return res
endif
```

autoload/lsp.vim
```diff
@@ -502,6 +502,42 @@ def lsp#showCurrentDiag()
...
+    return res
+  endif
+
+  var bnr: number = bufnr()
+  if lspserver.diagsMap->has_key(bnr)
+      for item in lspserver.diagsMap[bnr]->values()
+          if item->has_key('severity')
+              if item.severity == 1
+                  res.E = res.E + 1
```
yegappan, line 526: This can be simplified to "res.E += 1".

indelog: I can't. If i do this y get an error : `Index with operation not supported yet`

autoload/lsp.vim
indelog: Done

autoload/lsp.vim
indelog: Done

PR #22, opened 2021-11-11
Add : possibility to get hover result in preview window

indelog:
This add an option `g:LSP_Hover_Reply_In_Preview` which permit to see hover result in a preview window.

### DIFF
--- a/autoload/handlers.vim
+++ b/autoload/handlers.vim
@@ -302,7 +302,7 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
     ErrMsg('Error: Unsupported hover contents (' .. reply.result.contents .. ')')
     return
   endif
-  if g:LSP_Hover_Reply_In_Preview
+  if g:LSP_Hover_In_Preview
       silent! pedit HoverReply
       wincmd P
       setlocal buftype=nofile
--- a/autoload/handlers.vim
+++ b/autoload/handlers.vim
@@ -302,7 +302,21 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
     ErrMsg('Error: Unsupported hover contents (' .. reply.result.contents .. ')')
     return
   endif
-  hoverText->popup_atcursor({moved: 'word'})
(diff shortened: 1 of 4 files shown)


autoload/handlers.vim
```diff
@@ -302,7 +302,21 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
     ErrMsg('Error: Unsupported hover contents (' .. reply.result.contents .. ')')
     return
   endif
-  hoverText->popup_atcursor({moved: 'word'})
+  if g:LSP_Hover_Reply_In_Preview
```
yegappan, line 305: Can you rename this setting to "g:LSP_Hover_In_Preview"?

doc/lsp.txt
```diff
@@ -333,7 +333,9 @@ accepts a list of LSP servers with the above information.
 
 						*:LspHover*
 :LspHover		Show the documentation for the symbol under the cursor
-			in a popup window.
+			in a popup window. If you prefer show documentation in
+			a preview window rather a popup set
```
yegappan, line 337: Can you change the text to "If you want to show the symbol documentation in the preview window instead of in a popup, set the g:LSP_Hover_In_preview variable to v:true?

autoload/handlers.vim
indelog: Done.

doc/lsp.txt
indelog: Done.

indelog:
I have modified option for show hover reply in preview window for use the new `lspOptions`.

Shane-XB-Qian:
though looks a bit tradition way :-) 
but if so previous discussed pr#4 which adding 'option' on hover popup now is not req, preview perhaps looks more clean.

Shane-XB-Qian:
line#341 : `      if !reply.result.contents.kind->empty()`

// E715: Dictionary required

Shane-XB-Qian:
and seems syntax would not immediately update when if ft (in/of this preview) was 'markdown' .........

indelog:
> line#341 : ` if !reply.result.contents.kind->empty()`
> 
> // E715: Dictionary required

I try to solve this with #25

PR #23, opened 2021-11-11
Fix util#LspUriToFile with handler#processDocSymbolReply

indelog:
When get handler#processDocSymbol, textDocument.uri does not begin with
'file://', so util#LspUriToFile truncate the file name.

### DIFF
--- a/autoload/util.vim
+++ b/autoload/util.vim
@@ -56,7 +56,8 @@ export def LspUriToFile(uri: string): string
     # MS-Windows URI
     uri_decoded = uri_decoded[8 : ]
     uri_decoded = uri_decoded->substitute('/', '\\', 'g')
-  else
+  # On GNU/Linux (pattern not end with `:`)
+  elseif uri_decoded =~? '^file:///\a'
     uri_decoded = uri_decoded[7 : ]
   endif
 


autoload/util.vim
```diff
@@ -56,7 +56,8 @@ export def LspUriToFile(uri: string): string
     # MS-Windows URI
     uri_decoded = uri_decoded[8 : ]
     uri_decoded = uri_decoded->substitute('/', '\\', 'g')
-  else
+  # On GNU/Linux (pattern not end with `:`)
+  elseif uri_decoded =~? '^file:///\a'
```
yegappan, line 60: How to handle URIs that begin with "file://"? What do we do for the default else case?

indelog: If i trust this page <https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?>, the correct form of uri for a local file must  be `file:///path` (or `file:/path` but i never see this used anywhere).  Normally the lsp server mus return `file:///` from but i'm not sure of anythings (i don't see explicit information here <https://microsoft.github.io/language-server-protocol/specifications/specification-current/>).

I think we do not need default else case. If `uri_decoded` do not match the previous else, it remains unchanged. Normally it's mean the uri of the file is in this form `/path/to/the/file.ext` (not begin with `file:///`). This is my case when i get a reply from *phpactor* (an lsp server for php)  after an `LspSymbolSearch`. In this case if we keep the default else (that do `uri_decoded = uri_decoded[7 : ]`), the files paths obtained are like `o/the/file.ext` rather `/path/to/the/file.ext` (the first seven car are truncated).

PR #25, opened 2021-11-19
Fix : Better ft detection for preview widow on hover reply

indelog:
Avoid ` E715: Dictionary required` if lsp server does not provide the kind of reply.

This try to solve the problem encountred by @Shane-XB-Qian  in https://github.com/yegappan/lsp/issues/22#issuecomment-970182311

### DIFF
--- a/autoload/handlers.vim
+++ b/autoload/handlers.vim
@@ -293,14 +293,17 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
   endif
 
   var hoverText: list<string>
+  var hoverKind: string
 
   if reply.result.contents->type() == v:t_dict
     if reply.result.contents->has_key('kind')
       # MarkupContent
       if reply.result.contents.kind == 'plaintext'
         hoverText = reply.result.contents.value->split("\n")
+        hoverKind = 'text'
       elseif reply.result.contents.kind == 'markdown'
         hoverText = reply.result.contents.value->split("\n")
+        hoverKind = 'markdown'
       else
         ErrMsg('Error: Unsupported hover contents type (' .. reply.result.contents.kind .. ')')
         return
@@ -335,7 +338,7 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
     wincmd P
     setlocal buftype=nofile
     setlocal bufhidden=delete
-    exe 'setlocal ft=' .. reply.result.contents.kind
+    exe 'setlocal ft=' .. hoverKind
     deletebufline(bufnr(), 1, '$')
     append(0, hoverText)
     cursor(1, 1)
--- a/autoload/handlers.vim
+++ b/autoload/handlers.vim
@@ -335,8 +335,15 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
       wincmd P
       setlocal buftype=nofile
       setlocal bufhidden=delete
-      if !reply.result.contents.kind->empty()
-          exe 'setlocal ft=' .. reply.result.contents.kind
+      # If lsp server provide kind of hover reply and this kind is in syntaxs
+      # syported by vim, use it for file type of preview window. Else use
+      # type text by default.
+      if reply.result.contents->has_key('kind')
+\          && !reply.result.contents.kind->empty()
+\          && getcompletion('', 'filetype')->index(reply.result.content.kind) > -1
+        exe 'setlocal ft=' .. reply.result.contents.kind
+      else
+        set setlocal ft=text
       endif
       deletebufline(bufnr(), 1,  getbufinfo(bufnr())[0].linecount)
       append(0, hoverText)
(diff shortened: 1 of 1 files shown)


autoload/handlers.vim
```diff
@@ -331,19 +331,26 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
...
+    silent! pedit HoverReply
+    wincmd P
+    setlocal buftype=nofile
+    setlocal bufhidden=delete
+    # If lsp server provide kind of hover reply and this kind is in syntaxs
+    # syported by vim, use it for file type of preview window. Else use
+    # type text by default.
+    if reply.result.contents->has_key('kind')
+\        && !reply.result.contents.kind->empty()
+\        && getcompletion('', 'filetype')->index(reply.result.content.kind) > -1
```
Shane-XB-Qian, line 343: should be `contents` vs `content`

autoload/handlers.vim
```diff
@@ -331,19 +331,26 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
...
-      cursor(1, 1)
-      wincmd p
+    silent! pedit HoverReply
+    wincmd P
+    setlocal buftype=nofile
+    setlocal bufhidden=delete
+    # If lsp server provide kind of hover reply and this kind is in syntaxs
+    # syported by vim, use it for file type of preview window. Else use
+    # type text by default.
+    if reply.result.contents->has_key('kind')
```
Shane-XB-Qian, line 341: perhaps was a `list` not a `dict`

autoload/handlers.vim
```diff
@@ -331,19 +331,26 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
...
+\        && !reply.result.contents.kind->empty()
+\        && getcompletion('', 'filetype')->index(reply.result.content.kind) > -1
+      exe 'setlocal ft=' .. reply.result.contents.kind
+    else
+      set setlocal ft=text
+    endif
+    deletebufline(bufnr(), 1,  getbufinfo(bufnr())[0].linecount)
```
Shane-XB-Qian, line 348: looks can be simple to `deletebufline(bufnr(), 1, '$')`

autoload/handlers.vim
```diff
@@ -331,19 +331,26 @@ def s:processHoverReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>):
...
+    wincmd P
+    setlocal buftype=nofile
+    setlocal bufhidden=delete
+    # If lsp server provide kind of hover reply and this kind is in syntaxs
+    # syported by vim, use it for file type of preview window. Else use
+    # type text by default.
+    if reply.result.contents->has_key('kind')
+\        && !reply.result.contents.kind->empty()
+\        && getcompletion('', 'filetype')->index(reply.result.content.kind) > -1
+      exe 'setlocal ft=' .. reply.result.contents.kind
```
Shane-XB-Qian, line 344: what i meant in that comment about ft syntax was e.g hovertext for bash perhaps was  `markdown` with `man` type content, maybe had syntax for such `man` content (though/and in `markdown` kind/ft), but that maybe not highlighted initially before switched into preview window, though not very sure why or maybe not much people care about that...

Shane-XB-Qian (commented):
just a rough review, unluckly though simple but such things too much possibility... :-)

indelog:
I  did not pay attention, but the test for result content kind is already done at the beginning of the `s:processHoverReply` so no need to do it before display preview (i'm not very experimented, so i do stupid things).

autoload/handlers.vim
indelog: This is a suggestion. May be we can add an option for an lsp server to force the type of syntax to use with hover reply for special case with something like :

Example of server config : 
```vim
call lsp#addServer({
\       'filetype': 'bash',
\       'path': path/to/lsp',
\       'hoverReplyKind': 'man' " <- force to set syntax type for hover reply to 'man' for the bash lsp
\ })
```

In `autoload/handler.vim` for `s:processHoverReply` : 
```vim
# If we set the configuration of the server to use specific file type to display reply, use this type.
if lspserver->has_key('hoverReplyKind') && ! lspserver.hoverReplyKind->empty() && getcompletion('', 'filetype')->index(lspserver.hoverReplyKind) > -1
  hoverKind = lspserver.hoverReplyKind
else
  ...
endif
```

Shane-XB-Qian: i think IF it is a 'markdown' kind:
1, if it is a '```foo' such kind markdown content, then to do as following, or just a 'markdown';
2, directly get ft from its markdown mark e.g '```man' then means 'man', or just a 'markdown';
3, get rid (or conceal) of first line for this mark (and the last line '```')
it can write a separate func for this, we can the same thing for the popup hover as well if so.
// just like what i tried in pr #4, but just now we can set the ft both for lsphover preview and lsphover popup per this func.

Shane-XB-Qian: or whatever, above you did for preview lsphover perhaps was good enough for now.
// i've done such logic for popup lsphover at my local, and actually not much lsp server offerred markdown kind hover doc.

PR #29, opened 2022-01-15
Option to disable switching window on show references

kohnish:


yegappan (commented):
Can you also update doc/lsp.txt (lsp-options) with a description of this new option?

kohnish:
With pleasure!

kohnish:
It seems after resolving the conflict, showreferences now opens in near full window size. Is this expected?

yegappan:
This is not expected and I have fixed this in 8c12f1d804a71c9f148e1aa17bf51ef86ab7d67a.

doc/lsp.txt
```diff
@@ -202,6 +202,8 @@ following optional items:
     autoHighlightDiags	Automatically place signs on the lines with a
 			diagnostic message from the LSP server. By default
 			this is set to true.
+    keepFocusInReferences	Keep focus on quickfix window after ShowReferences
```
yegappan, line 205: The location list window is used to display the references. Can you change this to "Keep the focus in the location list window after |LspShowReferences| instead of moving the cursor back to the previous window."?

autoload/lspoptions.vim
```diff
@@ -17,6 +17,8 @@ export var lspOptions: dict<any> = {
   autoHighlightDiags: true,
   # Show the symbol documentation in the preview window instead of in a popup
   hoverInPreview: false,
+  # Focus on quickfix window on ShowReferences
```
yegappan, line 20: Can you change this comment to "Focus on the location list window after LspShowReferences"?

PR #30, opened 2022-01-16
Optional vim-lsc like behaviours

kohnish:
- Option to show diagnostics on status line
 - Option to close on select on reference window
 - Option to not suppress the diag hover appearing on mouse hover
 - Make outline configurable
 - Option to suppress adding a new line on completion selection

### DIFF
--- a/autoload/lsp.vim
+++ b/autoload/lsp.vim
@@ -35,6 +35,7 @@ if has('patch-8.2.4019')
   diag.DiagsGetErrorCount = diag_import.DiagsGetErrorCount
   diag.ShowAllDiags = diag_import.ShowAllDiags
   diag.ShowCurrentDiag = diag_import.ShowCurrentDiag
+  diag.ShowCurrentDiagInStatusLine = diag_import.ShowCurrentDiagInStatusLine
   diag.LspDiagsJump = diag_import.LspDiagsJump
   diag.DiagRemoveFile = diag_import.DiagRemoveFile
   symbol.ShowSymbolMenu = symbolsearch_import.ShowSymbolMenu
@@ -54,6 +55,7 @@ else
 	DiagsGetErrorCount,
 	ShowAllDiags,
 	ShowCurrentDiag,
+	ShowCurrentDiagInStatusLine,
 	LspDiagsJump} from './diag.vim'
   import ShowSymbolMenu from './symbolsearch.vim'
   import {OpenOutlineWindow, SkipOutlineRefresh} from './outline.vim'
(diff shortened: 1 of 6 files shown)


autoload/handlers.vim
```diff
@@ -446,6 +446,13 @@ def s:processReferencesReply(lspserver: dict<any>, req: dict<any>, reply: dict<a
 			text: text})
   endfor
 
+  if opt.lspOptions.referencesCloseOnSelect
+    augroup LspReferences
+      autocmd!
+      autocmd BufLeave * lclose
```
yegappan, line 452: As this autocmd applies to all the buffers, this may lead to unexpected side-effects.
Can you change this autocmd to use the <buffer> tag so that it only applies to the
current buffer?

autoload/lsp.vim
```diff
@@ -548,6 +555,23 @@ def lsp#showCurrentDiag()
...
+  var ftype = &filetype
+  if ftype == '' || @% == ''
+    return
+  endif
+
+  var lspserver: dict<any> = s:lspGetServer(ftype)
+  if lspserver->empty()
+    return
+  endif
+  if !lspserver.running
```
yegappan, line 568: Can you merge this if check with the previous if check?

autoload/lspoptions.vim
```diff
@@ -19,6 +19,20 @@ export var lspOptions: dict<any> = {
...
+  # Open outline window on right side
+  outlineOnRight: false,
+  # Suppress diagnostic hover from appearing when the mouse is over the line instead of the signature
+  suppressDiagHoverOnLine: true,
+  # Make quickfix window small like vim-lsc
+  smallReferencesWindow: false,
+  # Close the references window after selection
+  referencesCloseOnSelect: false,
+  # Show a diagnostic message on a status line
+  showDiagOnStatusLine: false,
```
yegappan, line 35: Can you update the help with a description for these new options?

autoload/outline.vim
```diff
@@ -234,7 +239,12 @@ export def OpenOutlineWindow()
 
   var prevWinID: number = win_getid()
 
-  :topleft :20vnew LSP-Outline
+  if opt.lspOptions.outlineOnRight
+    :setlocal splitright
```
yegappan, line 243: Instead of setting the 'splitright' option which may have unexpected side-effects,
can you use the "rightbelow' command modifier for the vnew?

plugin/lsp.vim
```diff
@@ -31,6 +41,7 @@ command! -nargs=0 -bar LspPeekImpl call lsp#gotoImplementation(v:true)
 command! -nargs=0 -bar LspShowSignature call lsp#showSignature()
 command! -nargs=0 -bar LspDiagShow call lsp#showDiagnostics()
 command! -nargs=0 -bar LspDiagCurrent call lsp#showCurrentDiag()
+command! -nargs=0 -bar LspDiagCurrentInStatusLine call lsp#showCurrentDiagInStatusLine()
```
yegappan, line 44: Can you update the help text with a description of this new command?

autoload/handlers.vim
kohnish: It'll affect all location list from other plugins as well, most likely. :/
I'll revert this one. I think it should go to vimrc of users, because it's not how locationlist is supposed to behave.

plugin/lsp.vim
kohnish: I'll remove this one because LspDiagCurrent is enough for explicit calling of the command. This is only for automatic status list showing which won't interrupt a user by removing linebreaks.

Issue #31, opened 2022-01-19
Gtabedit command in fugitive breaks lsp


kohnish:
1. Check out some git project.
2. Open test.c (assuming lsp, clangd, futigive installed and configured).
3. Gtabedit some_branch:test.c (some_branch must have the test.c file)
4. Do some lsp commands like finding references on the branch file. (file in a .git dir)
5. Go back to the test.c file and you can't do any lsp calls unless you restart the whole vim process.

kohnish:
I managed to screenshot the error on step 4.  

```
Error detected while processing function <SNR>25_output_cb[3]..handler#ProcessMessages[59]..handlers#ProcessReply[26]..<SNR>26_processReferencesReply[1]..symbol#ShowReferences:    
```
```
E684: list index out of range: 0
```

yegappan:
Thanks for reporting this issue. I have committed c2e8d7e3fda225acadc6e980fb1037cb1539f0ec to address this.

kohnish:
Thanks for fixing!

PR #32, opened 2022-01-19
Gtabedit command in fugitive breaks lsp #31

kohnish:
Not entirely sure what's happening here. But this fixes the issue.


kohnish:
c2e8d7e is the real fix. No need to be defensive in this part here.

PR #53, opened 2022-03-11
shane/mdf: to avoid if location (result) of defDecl was null

### LINKED ISSUE #51: location of DefDecl maybe null
```patch
diff --git a/autoload/lsp/handlers.vim b/autoload/lsp/handlers.vim
index 80f3ed8..47cab24 100644
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -131,20 +131,18 @@ enddef
 # process the 'textDocument/definition' / 'textDocument/declaration' /
 # 'textDocument/typeDefinition' and 'textDocument/implementation' replies from
 # the LSP server
 # Result: Location | Location[] | LocationLink[] | null
 def ProcessDefDeclReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>): void
   var location: dict<any>
-  if reply.result->type() == v:t_list
-    if !reply.result->empty()
-      location = reply.result[0]
-    else
-      location = {}
-    endif
-  else
+  if reply.result->type() == v:t_list && !reply.result->empty() && reply.result[0]->type() == v:t_dict
+    location = reply.result[0]
+  elseif reply.result->type() == v:t_dict
     location = reply.result
+  else
+    location = {}
   endif
 
   symbol.GotoSymbol(lspserver, location, req.method)
 enddef
 
 # process the 'textDocument/switchSourceHeader' reply from the LSP server
```

Shane-XB-Qian:
pr for issue #51 
closes #51

### DIFF
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -133,14 +133,14 @@ enddef
 # Result: Location | Location[] | LocationLink[] | null
 def ProcessDefDeclReply(lspserver: dict<any>, req: dict<any>, reply: dict<any>): void
   var location: dict<any>
-  if reply.result->type() == v:t_list
-    if !reply.result->empty()
-      location = reply.result[0]
-    else
-      location = {}
-    endif
-  else
+  if reply.result->type() == v:t_list && !reply.result->empty() && reply.result[0]->type() == v:t_dict
+    location = reply.result[0]
+  elseif reply.result->type() == v:t_dict
+    # not sure if there possible 'dict' type of 'result' but just in case
     location = reply.result
+  else
+    # cannot assign 'null' to 'location' and/so all else cases assign '{}'
+    location = {}
   endif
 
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
thanks for merge.
// i saw you rewrite it later, which is 'return' as long as it is 'empty', that's ok, but just no prompting info msg e.g something like 'no def from lsp' if so.

yegappan:
Yes. I wanted to simplify the code. This change should not result in any message if an empty reply is received from the LSP server. Let me know if you see any messages.

Shane-XB-Qian:
ok, then.
but such confirming msg personally i preferred get, for get msg about that symbol actually no def can be found.
other than that, some msg about lsp had no some capibility e.g `workspace/configuration` or `telemetry/event` (etc) specially the latter one print annoyed `unsupported balabala...` every editting, that perhaps can be a bit refined (for example just prompt one time)

PR #55, opened 2022-03-13
shane/mdf: request.id of response to lsp server perhaps was not a number

### LINKED ISSUE #54: type-mismatch: request.id maybe string not num
```vim
def SendResponse(lspserver: dict<any>, request: dict<any>, result: dict<any>, error: dict<any>)
  var resp: dict<any> = lspserver.createResponse(request.id)
```

Shane-XB-Qian:
closes #54

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -282,7 +282,12 @@ enddef
 
 # send a response message to the server
 def SendResponse(lspserver: dict<any>, request: dict<any>, result: dict<any>, error: dict<any>)
-  var resp: dict<any> = lspserver.createResponse(request.id)
+  if (type(request.id) == v:t_string && (trim(request.id) =~ '[^[:digit:]]\+' || trim(request.id) == ''))
+    || (type(request.id) != v:t_string && type(request.id) != v:t_number)
+    util.ErrMsg("Error: request.id of response to LSP server is not a correct number")
+    return
+  endif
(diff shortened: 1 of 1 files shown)


autoload/lsp/lspserver.vim
```diff
@@ -282,7 +282,11 @@ enddef
 
 # send a response message to the server
 def SendResponse(lspserver: dict<any>, request: dict<any>, result: dict<any>, error: dict<any>)
-  var resp: dict<any> = lspserver.createResponse(request.id)
+  if trim(string(request.id), "'") =~ '[^[:digit:]]\+' || trim(string(request.id), "'") == ''
```
yegappan, line 285: Do we need a check a for type(request.id) == v:t_string here before the call to trim()?

Shane-XB-Qian:
if it's not a number, after string() and trim(), I supposed there should some other chars (not number chars), so probably ok.
// the vim9 cannot compare number with other types directly, so had to cast first.

--
shane.xb.qian

autoload/lsp/lspserver.vim
Shane-XB-Qian: thx review, i am changing the way you suggested.

Shane-XB-Qian:
pls review again, i changed the way to you suggested.

castedo:
To wrap up here, a proper fix for requests from an LSP server with a string id has just been merged:
https://github.com/yegappan/lsp/pull/768

PR #58, opened 2022-03-30
Fire event when LSP client is attached to a buffer

gpanders:
This provides a hook for users to configure buffers that use LSP. For
example,

    autocmd User LSPAttached {
        nnoremap <buffer> <C-]> <Cmd>LspGotoDefinition<CR>
    }

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -406,6 +406,8 @@ export def AddFile(bnr: number): void
     endif
   augroup END
 
+  doautocmd <nomodeline> User LspAttached
+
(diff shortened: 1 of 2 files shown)


autoload/lsp/lsp.vim
```diff
@@ -406,6 +406,8 @@ export def AddFile(bnr: number): void
     endif
   augroup END
 
+  doautocmd <nomodeline> User LSPAttached
+
```
yegappan, line 410: Can you update the help file with a description of this autocmd event?

gpanders: Done, let me know if you have any feedback on the wording in the docs.

PR #64, opened 2022-09-14
Send an `initialized` message after sending "initialize"

tenderlove:
`rust-analyzer` will exit if we don't send an `initialized` message after sending the `initialize` message.  It doesn't seem like we really need to do anything to initialize, so this doesn't wait for the initialize request to return, it just sends the `initialized` message.

[Here is the message in the
spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialized)

I'm mainly using this with `clangd` and `rust-analyzer` right now.  I verified that `clangd` still works after making this change, but I'm not sure how to write a test (sorry, I'm very new to writing Vim extensions 😅)

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -118,8 +118,16 @@ def StartServer(lspserver: dict<any>, isSync: bool = false): number
   lspserver.running = true
 
   lspserver.initServer(isSync)
+  lspserver.initialized(isSync)
 
   return 0
+enddef
+
+def Initialized(lspserver: dict<any>, isSync: bool = false)
+  var req = {}
+  req.jsonrpc = '2.0'
+  req.method = 'initialized'
+  lspserver.sendMessage(req)
 enddef
 
 # Request: 'initialize'
@@ -1045,6 +1053,7 @@ export def NewLspServer(path: string, args: list<string>): dict<any>
   lspserver->extend({
     startServer: function(StartServer, [lspserver]),
     initServer: function(InitServer, [lspserver]),
+    initialized: function(Initialized, [lspserver]),
     stopServer: function(StopServer, [lspserver]),
     shutdownServer: function(ShutdownServer, [lspserver]),
     exitServer: function(ExitServer, [lspserver]),


dezza:
I can't say for certain, but I think this is work-in-progress.

I stumbled upon the same issue as you with `golang` or the `gopls` server.

However, after checking out your commits I think it might be redundant (there is already a `isSync`-boolean on StartServer/InitServer function signatures, but I think it's yet to be introduced in the `LspServer`-dict commonly put in `vimrc` (thats my theory, without knowing the intention) .. But it seems like the only way, right now.. to trigger `isSync` is to manually set the default argument to `true` since I could not find it used elsewhere.

https://github.com/yegappan/lsp/blob/875c7d859f2e0d631b7a3c5f232e108869194f70/autoload/lsp/lspserver.vim#L80-L181

Check out my fork/commit below, which makes a `sync`-property on the `LspServer`-dict
https://github.com/dezza/lsp/commit/c7a0128726ea2d58a2bcf269c49e4d1a2fd47fc5

This works; **Can you check if it does for you too?**

I believe these lines are where it happens (haven't checked or tested)

https://github.com/yegappan/lsp/blob/875c7d859f2e0d631b7a3c5f232e108869194f70/autoload/lsp/lspserver.vim#L1067

https://github.com/yegappan/lsp/blob/875c7d859f2e0d631b7a3c5f232e108869194f70/autoload/lsp/lspserver.vim#L125-L131

I also found this old issue talking about the flow with `gopls` https://github.com/golang/go/issues/35817#issuecomment-558087464 I imagine its a similar situation with rust-analyzer.

dezza:
rust-analyzer works for me with `sync: true`

```vim
g:lspServers = [
        \   {
        \   'filetype': ['rust'],
        \   'path': '/home/dza/.cargo/bin/rust-analyzer',
        \   'args': [],
        \   'sync': v:true,
        \   'omnicompl': v:true,
        \   },
        \   ]
lsp#lsp#AddServer(g:lspServers)
```
rust-anal

dezza:
@yegappan can you shed some light on this?

Does the `isSync`-property have anything to do with `{"capabilities":{"textDocumentSync":{"openClose":true` should `isSync` be true when some capabilities is returned by the server?

yegappan:
> 
> Does the `isSync`-property have anything to do with `{"capabilities":{"textDocumentSync":{"openClose":true` 
> should `isSync` be true when some capabilities is returned by the server?
>

The 'isSync' flag is used when restarting a LSP server (by the LspServerRestart) command.  This command
adds all the applicable buffers again to the LSP server.  This can be done only after the LSP server sends
the "intitialized" response.  So the "isSync" flag is used to wait for the "initialized" response from the
server.

Does the rust analyzer respond to the "initialize" request message from the client?  If it does, then the
plugin should send the "initialized" notification message.

yegappan:
@dezza Can you create a PR with your changes to make the isSync as an option in the LSP option dict?

dezza:
> @dezza Can you create a PR with your changes to make the isSync as an option in the LSP option dict?

Yes I considered this but was careful because it seemed like it "had purpose"/WIP.

The problem here is that both gopls and rust-analyzer won't work without sync, I've no idea since what you said makes sense.. 

I'm gonna try attempting enabling the trace variable you have and see if there's a clue. I vaguely remember this working out of the box 1-2 months ago when I last tested, but maybe not..

yegappan:
This issue should be addressed by the new syncInit flag that is added for each LSP server (e8c6f79666abddc9ba6823af47b5fed10c8a76f1 and 2f8ef07a087c62027467cef4bbe5d427f72150dc).  Are any of the changes in this PR still needed?

dezza:
@yegappan I mean, it looks like `sendInitializedNotif` is already handled as mentioned above, so I guess not. @tenderlove whats your take?

I have noticed crashes/errors with rust-analyzer, but I assume its more than likely just corner-cases where the client might send "too short" messages.. You can reproduce this if you type say `println!` and then press backspace slowly and repeatedly. I think its merely a task of polishing what is sent and when..

PR #67, opened 2022-09-23
lspserver: sync property

dezza:
* Removes isSync as parameter from StartServer(..), InitServer(..)
* Adds isSync parameter to NewLspServer(...)
* Adds `sync` to lspserver-properties (Defaults to `sync: false` in LspServer dictionary).

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -497,7 +497,12 @@ export def AddServer(serverList: list<dict<any>>)
       return
     endif
 
-    var lspserver: dict<any> = lserver.NewLspServer(server.path, args)
+    if !server->has_key('sync')
+      server['sync'] = v:false
+    endif
+
+    var lspserver: dict<any> = lserver.NewLspServer(server.path, args,
+          \ server.sync)
 
     if server.filetype->type() == v:t_string
       LspAddServer(server.filetype, lspserver)
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -77,9 +77,7 @@ enddef
 
 # Start a LSP server
 #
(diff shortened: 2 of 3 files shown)


dezza:
I'm happy if this is merged, because then I can test the markdown syntax on gopls and rust-analyzer, thats why I went back to this issue when I was working on the markdown PR.

autoload/lsp/lsp.vim
```diff
@@ -497,7 +497,12 @@ export def AddServer(serverList: list<dict<any>>)
       return
     endif
 
-    var lspserver: dict<any> = lserver.NewLspServer(server.path, args)
+    if !server->has_key('sync')
+      server['sync'] = v:false
+    endif
+
+    var lspserver: dict<any> = lserver.NewLspServer(server.path, args, 
+          \ server.sync ? server.sync : v:false)
```
yegappan, line 505: As you are initializing server.sync to a default value of v:false above, do you need to check whether it is true here?

yegappan (commented):
Can you update the doc/lsp.txt file (under the lsp-options tag) with a description of the sync option?

autoload/lsp/lsp.vim
dezza: correct, thats a mistake.. should have removed that when I made it default.

dezza:
> Can you update the doc/lsp.txt file (under the lsp-options tag) with a description of the sync option?

yes I'll fix both of ur requests

dezza:
Looks ok now?

dezza:
Let me know if it looks OK.

dezza:
Thank you! Working on an lsp-option "markdownCompact", "markdownHighlight" which will then enable the markdown features!

Issue #12, opened 2021-03-09
[Mark] didChangeConfiguration


Shane-XB-Qian:
looks haveNot impl method: ‘workspace/didChangeConfiguration’ yet
supposed you knew, but for a mark, so opening this ticket...

yegappan:
Yes. The 'workspace/didChangeConfiguration' notification is not yet implemented. But from the LSP protocol
specification it is not clear to me when the client is supposed to send this notification to the LSP server and
how the LSP server will use this notification.

Shane-XB-Qian:
actually, i am not sure, but i saw some lsp server would say what it supported cfg is, i am not sure if it was uniformed.

PR #70, opened 2022-09-29
Provide LSP server initializationOptions support

atgote:
This is to add support for LSP server initializationOptions. For me it was important to allow transmission of the licenseKey to the intelephense PHP language server, however it is a standard field and other servers configurations may require these too.

Some of configurations are (or may be) expected to be determined by the editor ('LSP client') however this should require individual per-server 'wrappers' or something similar.

* Add 'initializationOptions' as LSP server configuration option.
* Check if 'initializationOptions' are not empty and if so transmit these to the LSP server
* Updated README.md
* Updated plugin documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,15 @@ To register a LSP server, add the following lines to your .vimrc file (use only 
 		\	 filetype: ['fortran'],
 		\	 path: '/usr/local/bin/fortls',
 		\	 args: ['--nthreads=1', '--use_signature_help', '--hover_signature']
+		\      },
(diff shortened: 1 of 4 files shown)


Shane-XB-Qian:
cool! thx, this is this lsp client big diff with others.
// though some lsp server can read cfg from e.g ~/.config/foooo also.

Shane-XB-Qian:
but how/what about #12 `didChangeConfiguration` (or even `workspace/configuration`) ?

Issue #66, opened 2022-09-21
Markdown popup (WIP)


dezza:
![vim9lsp-markdown](https://user-images.githubusercontent.com/402927/191416460-b4e78467-a886-49fc-8cad-aaef723a131e.png)

```vim
g:markdown_fenced_languages = ['help', 'python', 'bash=sh', 'vim', 'rust'] # add as needed
g:markdown_syntax_conceal = 1

set completeopt=menuone,popup,noinsert,popuphidden
def MarkdownPopup()
  w:vim9lsp_md_applied = get(w:, 'vim9lsp_md_applied', 0)
  var win = popup_findinfo()
  if w:vim9lsp_md_applied == 0
    var buf = winbufnr(win)
    if !!buf && buf != 1
      setwinvar(win, "&conceallevel", 3)
      setbufvar(buf, "&number", "0")
      setbufvar(buf, "&filetype", "markdown")
      w:vim9lsp_md_applied = 1
    endif
  endif
  popup_show(win)
enddef
autocmd CompleteChanged * MarkdownPopup()
```

I've been working on getting markdown syntax into the popup like the other lsp clients. I ended up using a similar strategy to how
[`complete-popuphidden`](https://vimhelp.org/insert.txt.html#complete-popuphidden) does it, to integrate with vim's own [`complete()`](https://vimhelp.org/builtin.txt.html#complete%28%29) function and `complete_info(dict)` https://github.com/yegappan/lsp/blob/7cc46b241b6e542fe0a8f2fafee65da3ddb1899b/autoload/lsp/handlers.vim#L292 (other clients uses `bufadd` and then `popup_create`).

There is still some things I'd like to improve. [vim-lsp](https://i.postimg.cc/3Nmd6Dy0/vim9lsp-markdown.png), for example unbinds the default markdown syntax regions (in the complete-popup that is) to achieve an even higher level of cleanliness. Mainly what it does is that it joins \`\`\` lang markdown-tokens (or code-fences) so they conceal on the same line (not leaving an empty line where \`\`\` was concealed). Also it defines its own fenced_languages, so your `g:markdown_fenced_languages` can stay agnostic and unassuming to your own `.md`-filetype preferences. I like this, I have already done it and saved lines that I can add to this function later, goes something like this
```vim
syntax sync clear
unlet g:markdown_fenced_languages
runtime! syntax/markdown.vim
let g:markdown_fenced_languages = ['help']
syntax clear markdownCodeBlock | syntax clear markdownCode
[...]

PR #75, opened 2022-10-22
markdown complete-popup

dezza:


autoload/lsp/lsp.vim
```diff
@@ -242,6 +242,7 @@ enddef
 
 # A new buffer is opened. If LSP is supported for this buffer, then add it
 export def AddFile(bnr: number): void
+  unsilent echomsg "Lsp: AddFile"
```
yegappan, line 245: Can you remove this debug message?

autoload/lsp/util.vim
```diff
@@ -21,7 +21,7 @@ if has('unix')
 else
   lsp_log_dir = $TEMP .. '\\'
 endif
-var lsp_server_trace: bool = false
+var lsp_server_trace: bool = true
```
yegappan, line 24: Can you revert this change?

plugin/lsp.vim
```diff
@@ -9,6 +9,8 @@ vim9script
 
 import autoload '../autoload/lsp/options.vim'
 import autoload '../autoload/lsp/lsp.vim'
+import autoload '../autoload/lsp/util.vim'
+import autoload '../autoload/lsp/documentation.vim' as doc
```
yegappan, line 13: Why are these files imported here?

dezza:
Sorry for the mistakes 👍 its getting late over here.

dezza:
helplsp

dezza:
I'll make another update. I realize I'd rather call it "CompletePopup" and make it run in any case with markdown being conditional in the option.

dezza:
@yegappan what do you think now?

Btw, do you think we should add some defaults in:
https://github.com/yegappan/lsp/blob/598ba2e8c79bca0000fe2bd16739243c6914632f/autoload/lsp/lsp.vim#L287-L291

Or move these settings out of `opt.lspOptions.autoComplete` and make it apply in any case? I see no harm.. But I can't see a reason why we should do it differently for omnicomplete.

https://github.com/yegappan/lsp/blob/598ba2e8c79bca0000fe2bd16739243c6914632f/autoload/lsp/lsp.vim#L275-L279

yegappan:
I am trying to integrate your changes and the PR with the changes to support complete/resolve.  I see there is a overlap in the way 
CompleteDone autocmd is used by both the PRs.  I am going to merge the complete/resolve PR first.  Can you then update your PR and merge those changes?

dezza:
Yes I will check it.

I just right now discovered a few issues I missed. Let me have a look later if I can combine them about to sleep.

rcasta74:
I tried to integrate this code (598ba2e) with my last changes, but unfortunately it look a bit ugly to me (I'm using gruvbox colorscheme vith `background=dark`):
image
Not sure if I have made it wrongly.
Maybe the code should be off by default letting the user to enable it if wanted.

dezza:
@rcasta74 please tell me which java plugins/setup you have and I'll take a look.

My branch has been thru some iterations the last 2 days, also discovered several bugs which I just fixed. But the markdown looks fine for me (testing with rust-analyzer, empty fn main println statement)

dezza:
@yegappan please see comments in recent push

https://github.com/yegappan/lsp/blob/419179d28d34dd147bae42463df621d827dd0121/autoload/lsp/lsp.vim#L244-L280

Otherwise, merged and tested. Not with `eclipse-jdt-ls` though. I can try it out if I get confirmation thats where the visual quirks are.

dezza:
Should be ready for merge now with docs. I'm waiting for your input on the 2 lines then I'll edit them out again so you can merge.

autoload/lsp/lsp.vim
```diff
@@ -252,14 +253,21 @@ def BufferInit(bnr: number): void
   # add a listener to track changes to this buffer
   listener_add(Bufchange_listener, bnr)
 
+  # FIXME: Maybe we should introduce defaults regardless of auto/omnicompl
```
yegappan, line 256: Can you remove this change from this PR?  If needed, we can make this change in a separate PR.

dezza: sure, understandable - get the history right.

I'll open a PR afterwards.

Shane-XB-Qian:
> Maybe the code should be off by default letting the user to enable it if wanted.

yes, i agree, this pr refactor/change too much, specially most of them actually was user preference.
for example, i personally do not like set `conceal` in this info doc of lsp compl.

perhaps just let keep it which `au user` (already one there) to set it by themselves.
// and markdown ft already many sytanx plugin too, suggest let user to choose better.

dezza:
@Shane-XB-Qian all fixed now and you got your `User` `LspCompletePopup` to override everything incl example :)

Also managed to evade a `E847: Too many syntax includes` with an added `syntax clear` in front. The syntax includes are also checked against `$VIMRUNTIME` neither of which are handled in vim-lsp as of now.. But I'll report them.

The popup can fly by now buttery-smooth now without stacking `syntax include` (especially with fast redraw on gvim)

rcasta74:
> please tell me which java setup (lsp-server, plugins) you have and I'll take a look.

@dezza  ok, just figured out that the ugly visualization was due to conceal feature, as pointed out by Shane-XB-Qian
I'm trying to work on it.

On regard to this PR, I'm just wondering why you haven't used a ftplugin; I think it could be cleaner.
We are already forcing `filetype='markdown'` when the server replay `kind='markdown'`; I think we could instead force an ad-hoc filetype in this case (let say for example 'markdown_lsp') so to not override the default 'markdown' ftplugin.
In this way it should also be easier to let the user to customize the popup visualization for each server; we can add a server property named 'markdown': if not configured, the standard behaviour will apply, otherwise if is configured as (for example) `markdown='jdt'` then the popup will be forced to have `filetype='markdown_jdt'` and the ftplugin of that name will be applied (that could be provided into lsp plugin itself or created by scratch by the user).
What do you think about this idea?

yegappan:
> On regard to this PR, I'm just wondering why you haven't used a ftplugin; I
> think it could be cleaner.  We are already forcing filetype='markdown' when
> the server replay kind='markdown'; I think we could instead force an ad-hoc
> filetype in this case (let say for example 'markdown_lsp') so to not override
> the default 'markdown' ftplugin. What do you think about this idea?
>

I like this idea.  I was thinking along the same lines.  We can use a separate
'filetype' for the insert completion popup documentation buffer and use a
separate syntax and filetype plugin files.  These files can be packaged with
the LSP plugin.   If a user wants to override these, he can use the "after"
directory.

dezza:
@rcasta74 what is the outlook for such a ftplugin? Strip out relevant parts of runtime and trim it down like that to only what we use?

I can't find where we force `filetype=markdown` are you referring to my commit?

Overall I think it sounds like a good idea.

I'll present some more issues when I've slept.. Currently battling `bash-language-server`, but I've written a cool shellscript to write to named pipes to ease the development/testing workflow. Something like this:

```sh
exec 3>pipelsp # write without blocking, by using exec to pipe to a custom fd3 on the parent shell process (takes over this tty)

# -r prevents expansion of escape sequences
read -d '' -r lsp_initialized <<-'EOF'
Content-Length: 58\r\n
\r\n
{"method": "initialized", "jsonrpc": "2.0", "params": {}}\r
EOF

read -d '' -r lsp_open <<-'EOF'
Content-Length: 194\r\n
\r\n
{"method": "textDocument/didOpen", "jsonrpc": "2.0", "params": {"textDocument": {"uri": "file:///home/dza/test2/test2.sh", "version": 1, "languageId": "sh", "text": "echo\\necho \"hestepik\""}}}\r
EOF

read -d '' -r lsp_didChange2 <<-'EOF'
Content-Length: 190\r\n
\r\n
{"method": "textDocument/didChange", "jsonrpc": "2.0", "params": {"textDocument": {"uri": "file:///home/dza/test2/test2.sh", "version": 2}, "contentChanges": [{"text": "echo hestepik\\n"}]}}\r
EOF

echo "lsp_initialize"
while read -r line
do
  printf '%b' "${line}" >&3
  printf '%b\n' "${line}"
done <<< "${lsp_initialize}"
sleep 7 # Increase here until it gets capabilities back

echo "lsp_initialized"
while read -r line
do
  printf '%b' "${line}" >&3
  printf '%b\n' "${line}"
done <<< "${lsp_initialized}"
sleep 2

echo "lsp_didOpen"
while read -r line
do
  printf '%b' "${line}" >&3
  printf '%b\n' "${line}"
done <<< "${lsp_open}"
#
sleep 2
echo "lsp_didChange2"
while read -r line
do
  printf '%b' "${line}" >&3
  printf '%b\n' "${line}"
done <<< "${lsp_didChange2}"

sleep 10
echo "closing pipe|happens if script ends anyways"
#exec 3<&-
```

On the other side lspserver-side:
```sh
bash-language-server start 0<pipelsp
```

Feel free to use this so you can test faster :)

rcasta74:
> I can't find where we force `filetype=markdown`

It is part of merged #77:
https://github.com/yegappan/lsp/blob/9e558794fac6dcece03a493925a584906dacae2e/autoload/lsp/handlers.vim#L225-L227
https://github.com/yegappan/lsp/blob/9e558794fac6dcece03a493925a584906dacae2e/autoload/lsp/handlers.vim#L253-L256

It is also done with Hover functionality (with preview window):
https://github.com/yegappan/lsp/blob/9e558794fac6dcece03a493925a584906dacae2e/autoload/lsp/handlers.vim#L283-L285
https://github.com/yegappan/lsp/blob/9e558794fac6dcece03a493925a584906dacae2e/autoload/lsp/handlers.vim#L315-L320

so if we use an ad-hoc filetype for both, I guess they will be rendered in the same way.
(Maybe filetype could be forced also for Hover in popup window)

rcasta74:
> If a user wants to override these, he can use the "after"
> directory.

This way the same plugin will be used for all lsp servers: it may work well with a lsp server but not with another one.
With my proposal (an optional 'markdown' property when registering the server list) will permit to use different plugin for each servers.

Shane-XB-Qian:
> I think we could instead force an ad-hoc filetype in this case (let say for example 'markdown_lsp')

i was saying and doing similar, see #4 
at my local there `lsphover` and `lspsign` ft actually, and this `lspinfo` if you'd like the name. :-)

// by the way, your #77 pr perhaps need to check/impl signature also, seems it did not work after so, not sure due to `async` broken it or it was not working (from `jdt`) originally.

rcasta74:
> // by the way, your #77 pr perhaps need to check/impl signature also, seems it did not work after so, not sure due to `async` broken it or it was not working (from `jdt`) originally.

Not sure to have understood how signature help should work, but if I trigger it at this position
image
the request looks good to me
`{'id': 9, 'jsonrpc': '2.0', 'method': 'textDocument/signatureHelp', 'params': {'textDocument': {'uri': 'file:///home/roberto/VIM-IDE/appo/HelloWorld.java'}, 'position': {'character': 25, 'line': 3}}}`
but the server reply with empty signatures
`{'id': 9, 'jsonrpc': '2.0', 'result': {'signatures': []}}`
So it may be an issue with jdt server itself.

Shane-XB-Qian:
most case it kicked at first `(` or end of that key word e.g `println` in your case, but anyway seems yes, jdt did not offer that/signature or some way (like its info doc was different like most others) we're not familiar with yet. :-)

--
shane.xb.qian

Shane-XB-Qian:
> So it may be an issue with jdt server itself.

yes, i tried again, looks yes, jdt does not support signature help like general others.
// sorry for the noise.

as for this info doc of lsp compl here, 
1, short workaround is like https://github.com/yegappan/lsp/issues/66#issuecomment-1288450404 said.
2, long term suggestion is like above discussed, perhaps a `au user` is good enough on its ft.

rcasta74:
@yegappan I have just noticed that you had to fix the indentation after my PR: could you update the modeline in the scripts so that your style will be always applied?
At least `noexpandtab` shoud be added; I don't know if other settings are needed

yegappan:
>
> @yegappan I have just noticed that you had to fix the indentation after my
> PR: could you update the modeline in the scripts so that your style will be
> always applied?  At least noexpandtab shoud be added; I don't know if other
> settings are needed
>

The plugin source files use a mix of tabs and spaces (softtabstop).  The modeline
is already present in these files with the proper indentation related options.
I had to align the resolveSupport field in your PR.

rcasta74:
> The plugin source files use a mix of tabs and spaces (softtabstop). The modeline
> is already present in these files with the proper indentation related options.

 I have `set expandtab` in my configuration; because of it I broked the proper indentation.
Maybe also `tabstop` option should be forced to your value (I'm using `tabstop=4`).
If you want I can open a PR with these additions.

rcasta74:
> With my proposal (an optional 'markdown' property when registering the server list) will permit to use different plugin for each servers.

OK, looking deeper at LSP specification and just found that if the format is markdown the content should follow a precise specification ([markupContent](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContent)), so my proposal doesn't make sense :-)

yegappan:
I am planning to merge this.  Let me know if you see any problems in merging this.

yegappan:
I have made changes to way the asynchronous callbacks are used.  You will see conflicts when you pull in the latest changes.

rcasta74:
I'm still working on the parser (rcasta74/lsp@2aea363); so far not all [GFM Specification](https://github.github.com/gfm/) are covered, but it should be enough for most of the documentations.
This is how it looks for jdt documentation:
image
Please give it a try with other language servers and let me know your thought.
It can be also tested with any markdown file: just open it and use a command like:
```
command TestPopup call win_execute(getline(1, '$')->popup_create({'pos': 'center', 'maxwidth': 80, 'highlight': 'PMenu', 'border': [0, 1, 0, 1], 'borderchars': [' '], 'moved': 'any'}), $'setlocal linebreak ft=lspgfm')
```
Currently I'm trying to add the language highlight for the fenced code blocks, but I'm finding it quite tricky. My idea was to use commands like:
```
syntax include @Java syntax/java.vim
syntax region codeJava start='\%1l^' end='\%1l$' contains=@Java
```
for each line where there is fenced code: it works if I try them directly in Ex-mode, but I cannot get them to work within the script.

yegappan:
This looks much better.  I tried this with the Rust language server.  This server doesn't support the lazy resolution of
the documentation information for a completion item.  The documentation is supplied along with the completion
matches.  I made the following changes to use the 'markdown' file type for the documentation popup window.
```
index 4f62aaf..6081c8c 100644
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -293,6 +293,11 @@ def AddBufLocalAutocmds(lspserver: dict<any>, bnr: number):
 void
     endif
   endif
 
+  acmds->add({bufnr: bnr,
+                 event: 'CompleteChanged',
+                 group: 'LSPBufferAutocmds',
+                 cmd: 'LspSetFileType()'})
+
   # Execute LSP server initiated text edits after completion
   acmds->add({bufnr: bnr,
              event: 'CompleteDone',
@@ -656,6 +661,27 @@ def LspResolve()
   endif
 enddef
 
+# If the completion popup documentation window displays 'markdown' content,
+# then set the 'filetype' to 'lspgfm'.
+def LspSetFileType()
+  var item = v:event.completed_item
+  if !item->has_key('user_data') || empty(item.user_data)
+    return
+  endif
+
+  var cItem = item.user_data
+  if !cItem->has_key('documentation') || cItem->type() != v:t_dict
+                               || cItem.documentation.kind != 'markdown'
+    return
+  endif
+
+  var id = popup_findinfo()
+  if id > 0
+    var bnum = id->winbufnr()
+    setbufvar(bnum, '&ft', 'lspgfm')
+  endif
+enddef
+
 # omni complete handler
 def g:LspOmniFunc(findstart: number, base: string): any
   var lspserver: dict<any> = CurbufGetServerChecked()
```

rcasta74:
Opened PR #104 containing your above changes.
Compared to last week, I have added the syntax for fenced code with string info and a fallback to markdown syntax if the buffer is not in a popup window or the parser fails.

yegappan:
Now that 3fe5fdf30a78d6456207e31d55eeaac2730ec648 has been merged to support the markdown syntax in the completion info popup window, do we need any changes from this PR?

Shane-XB-Qian:
> do we need any changes from this PR?

to be honest, i felt this changed too much at one time..
1, how to make sure those added au event just worked for this lsp client?
// please note there many others complete type there, 
// but e.g au event `CompleteChanged` really just worked for only this lsp compl?
2, ... something else, but just forgot what to say for now.. :-( aha..
// i spent much time to check the outage after this and that complete.vim code move.. :-(

Shane-XB-Qian:
those this lsp specific ft `lspgfm` and `au completechanged` etc perhaps need some fence or protection which not to impact others `compl` and/or its content/syntax/popup-option/etc...

hrsh7th:
Feel free to take the markdown syntax codes.

https://github.com/hrsh7th/vim-vital-vs/blob/master/autoload/vital/__vital__/VS/Vim/Syntax/Markdown.vim

PR #79, opened 2022-10-24
fix: check trig char if exist first otherwise broken.

Shane-XB-Qian:


autoload/lsp/handlers.vim
```diff
@@ -25,8 +25,9 @@ def ProcessInitializeReply(lspserver: dict<any>, req: dict<any>, reply: dict<any
   lspserver.caps = caps
 
   if opt.lspOptions.autoComplete && caps->has_key('completionProvider')
-    var triggers = caps.completionProvider.triggerCharacters
-    lspserver.completionTriggerChars = triggers
+    if caps.completionProvider->has_key('triggerCharacters')
+	lspserver.completionTriggerChars = caps.completionProvider.triggerCharacters
```
yegappan, line 29: Can you change this to?

lspserver.completionTriggerChars = caps.completionProvider->get('triggerCharacters', [])

Shane-XB-Qian:
completionTriggerChars default value or init value was `[]` already.
or if you insisted that `get()` style I can change, though a little dup.

--
shane.xb.qian

yegappan:
By using the get() function, you can remove the "if" statement.

Shane-XB-Qian:
yes. but `if` was to emphasize that `has_key` existed first. here original problem was due to it did not check vs not the init value :-)
anyway, just saying my thoughts this just a tiny fix. :lol

--
shane.xb.qian

PR #81, opened 2022-10-25
Add option to show diagnostics in a popup

D4yvid:
Added a option to show diagnostics in a floating popup

image

The option name is `showDiagInPopup` and it's default is `false`

### DIFF
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -33,7 +33,9 @@ export var lspOptions: dict<any> = {
   # Show a diagnostic message on a status line
   showDiagOnStatusLine: false,
   # Don't print message when a configured language server is missing.
-  ignoreMissingServer: false
+  ignoreMissingServer: false,
+  # Make diagnostics show in a popup instead of echoing
+  showDiagInPopup: true
 }
 
 # set LSP options from user provided options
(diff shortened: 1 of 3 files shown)


autoload/lsp/options.vim
```diff
@@ -33,7 +33,9 @@ export var lspOptions: dict<any> = {
   # Show a diagnostic message on a status line
   showDiagOnStatusLine: false,
   # Don't print message when a configured language server is missing.
-  ignoreMissingServer: false
+  ignoreMissingServer: false,
+  # Make diagnostics show in a popup instead of echoing
+  showDiagInPopup: true
```
yegappan, line 38: Can you add also update the lsp.txt help file with a description of this new option?

D4yvid: Yeah, i'll do that

D4yvid: Done, i've added

doc/lsp.txt
```diff
@@ -264,6 +264,9 @@ showDiagOnStatusLine	Show a diagnostic message on a status line.
 			By default this is set to false.
 ignoreMissingServer	Do not print a missing language server executable.
 			By default this is set to false.
+showDiagInPopup		Use a popup to display server diagnostics instead of
+			echoing.
+			By default this is set to false.
```
yegappan, line 269: By default, is this option set to true or false?

D4yvid: Wait, i've forgot to change it back to false, i was testing it

D4yvid: Sorry, i've changed it back to false

PR #83, opened 2022-10-26
fix: not jump if cur one is same buf also

### LINKED ISSUE #9: jump per winid may mess-up
i saw this plugin used to use window jump per winid..
it probably good for 'save-id -> do-something -> jump-back-To-saved-id' this style.
But:
- `var wid: number = fname->bufwinid()`
    * perhaps there were many same fname opened there, including current one also is..........
    * but such would get only the first found. mostly jump is Not wanted when/if current one also is.
- `wid->win_gotoid()`
    * Specially such jump would over 'Tabs', that may looks mess, user perhaps used to focus on cur tab.
    * BTW: latest win_gotoid() func had 'bug': it can escape from popup/terminal window, that may happen unexpected issue.

// I suggest to add a judgement for above case:
- if cur one is same 'fname', then donot jump.
- ~~jump should not over Tabs, if winid not in cur tab, then give-up.~~ try cur tab first.

Shane-XB-Qian:
closes #9 

`fname->bufwinid()` perhaps randomly pick up one, but cur window perhaps was the same buf
// so firstly pick up cur buf if was same buf, then same buf in cur tab if there was one,
// or if none of them, then randomly pick one, or not found.

### DIFF
--- a/autoload/lsp/symbol.vim
+++ b/autoload/lsp/symbol.vim
@@ -103,13 +103,16 @@ def JumpToWorkspaceSymbol(popupID: number, result: number): void
     else
       var wid = fname->bufwinid()
       if wid != -1
+        # if had one in cur tab
+        # and cur win is same buf
         if bufwinid(bufnr()) == wid
-          # do nothing if cur one is
+          # do nothing
         else
-          # jump to one in cur tab
+          # or pick up one in cur tab
           wid->win_gotoid()
         endif
       else
+        # or pick up one in one tab
         winList[0]->win_gotoid()
       endif
     endif
@@ -301,7 +304,10 @@ export def GotoSymbol(lspserver: dict<any>, location: dict<any>, peekSymbol: boo
     # jump to the file and line containing the symbol
     var wid = fname->bufwinid()
     if wid != -1
-      # do not jump if cur one is same buf
+      # `wid` maybe just is one of windows in cur tab which had same buf
+      # jump to `wid` only if cur one is not same buf or maybe a bit mess
+      # which should always try to re-use cur one firstly (if it is same)
+      # otherwise cursor perhaps dumbly jumpped to `wid` though same buf
       if bufwinid(bufnr()) != wid
         wid->win_gotoid()
       endif
--- a/autoload/lsp/symbol.vim
+++ b/autoload/lsp/symbol.vim
@@ -101,7 +101,17 @@ def JumpToWorkspaceSymbol(popupID: number, result: number): void
 	exe $'confirm edit {symTbl[result - 1].file}'
       endif
     else
-      winList[0]->win_gotoid()
+      var wid = fname->bufwinid()
+      if wid != -1
+        if bufwinid(bufnr()) == wid
+          # do nothing if cur one is
+        else
+          # jump to one in cur tab
+          wid->win_gotoid()
+        endif
+      else
+        winList[0]->win_gotoid()
+      endif
     endif
     # Set the previous cursor location mark. Instead of using setpos(), m' is
     # used so that the current location is added to the jump list.
@@ -291,7 +301,10 @@ export def GotoSymbol(lspserver: dict<any>, location: dict<any>, peekSymbol: boo
     # jump to the file and line containing the symbol
     var wid = fname->bufwinid()
     if wid != -1
-      wid->win_gotoid()
+      # do not jump if cur one is same buf
+      if bufwinid(bufnr()) != wid
+        wid->win_gotoid()
+      endif
     else
       var bnr: number = fname->bufnr()
       if bnr != -1


autoload/lsp/symbol.vim
```diff
@@ -291,7 +301,10 @@ export def GotoSymbol(lspserver: dict<any>, location: dict<any>, peekSymbol: boo
     # jump to the file and line containing the symbol
     var wid = fname->bufwinid()
     if wid != -1
-      wid->win_gotoid()
+      # do not jump if cur one is same buf
+      if bufwinid(bufnr()) != wid
```
yegappan, line 305: The bufwinid() function returns the window-ID of the first window associated with a buffer.  The 'wid' variable is set to the window ID returned by fname->bufwinid().  So I am not sure what this check is preventing?  If the current window is the correct window, then jumping to it using win_gotoid() should not change anything.  Does jumping to the current window using win_gotoid() changing anything?

autoload/lsp/symbol.vim
```diff
@@ -101,7 +101,17 @@ def JumpToWorkspaceSymbol(popupID: number, result: number): void
 	exe $'confirm edit {symTbl[result - 1].file}'
       endif
     else
-      winList[0]->win_gotoid()
+      var wid = fname->bufwinid()
+      if wid != -1
+        if bufwinid(bufnr()) == wid
```
yegappan, line 106: To check if the current buffer already has the desired file, the following check should be used:
```
      if fname->bufnr() == bufnr()
```

Shane-XB-Qian:
it meant if cur one is the one already then do not jump to another though that one is same buf.

- for example:
buf_a / win_1 | buf_a / win_2

current focus or cursor at win_2,
now you are going to find a symbol which buf_a had,

this pr is preventing you are not jumped to win_1, since cur window (win_2) is buf_a too!


- now if it is:
tab_foo: buf_a / win_1 | buf_b / win_2
tab_bar: buf_a / win_3 | buf_b / win_4

current focus or cursor at win_2,
now you are going to find a symbol which buf_a had,

then jumped to tab_foo / win_1 firstly which preventing you jump to tab_bar directly.


- if cur tab really had no buf_a, then now random pick up one or it was not one found.

--
shane.xb.qian

Shane-XB-Qian:
added more comment, wish now it is more clarified.

autoload/lsp/symbol.vim
```diff
@@ -101,7 +101,20 @@ def JumpToWorkspaceSymbol(popupID: number, result: number): void
...
-      winList[0]->win_gotoid()
+      var wid = fname->bufwinid()
+      if wid != -1
+        # if had one in cur tab
+        # and cur win is same buf
+        if bufwinid(bufnr()) == wid
+          # do nothing
+        else
+          # or pick up one in cur tab
+          wid->win_gotoid()
```
yegappan, line 112: Is there a need to check bufwinid(bufnr()) == wid?  If we always do wid->win_gotoid() instead, what will not work?

autoload/lsp/symbol.vim
```diff
@@ -291,7 +304,13 @@ export def GotoSymbol(lspserver: dict<any>, location: dict<any>, peekSymbol: boo
     # jump to the file and line containing the symbol
     var wid = fname->bufwinid()
     if wid != -1
-      wid->win_gotoid()
+      # `wid` maybe just is one of windows in cur tab which had same buf
+      # jump to `wid` only if cur one is not same buf or maybe a bit mess
+      # which should always try to re-use cur one firstly (if it is same)
+      # otherwise cursor perhaps dumbly jumpped to `wid` though same buf
+      if bufwinid(bufnr()) != wid
```
yegappan, line 311: This condition is true if the current window ID is not the same as the window ID of the buffer containing fname.  This doesn't match the description in the comment.

autoload/lsp/symbol.vim
Shane-XB-Qian: `bufwinid(bufnr()) == wid` is to explain : should not jump if cur win had the same buf then do not jump,
which to be clear just let it be jumped when cur window it is Not same buf.

Q:
> If we always do wid->win_gotoid() instead, what will not work?

A: user perhaps had opened same buf in some windows in cur tab (or not), if just win->win_gotoid() your cursor maybe dumbly fly to another window, that's annoyed and weird, since cur one it is already!

autoload/lsp/symbol.vim
Shane-XB-Qian: there should be another 
```vim
        # if had one in cur tab
        # and cur win is same buf
        if bufwinid(bufnr()) == wid
          # do nothing
        else
          # or pick up one in cur tab
          wid->win_gotoid()
        endif
```

i simplified it just `bufwinid(bufnr()) != wid`

autoload/lsp/symbol.vim
```diff
@@ -101,7 +101,20 @@ def JumpToWorkspaceSymbol(popupID: number, result: number): void
 	exe $'confirm edit {symTbl[result - 1].file}'
       endif
     else
-      winList[0]->win_gotoid()
+      var wid = fname->bufwinid()
```
yegappan, line 104: This can be replaced with the following:
```
     if bufnr() != fname->bufnr()
        var winID = fname->bufwinid()
        if winID == -1
          # not present in the current tab page
          winID = winList[0]
        endif
        winID->win_gotoid()
      endif
    endif
```

PR #84, opened 2022-10-26
Pr ft server status

Shane-XB-Qian:
add LspServerStatus(ft) func to detect ft lsp server was running or not (or not found)

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -97,19 +97,6 @@ export def EnableServerTrace()
   util.ServerTrace(true)
 enddef
 
-# Get LSP server status for filetype 'ftype'.
-# Return true if running, otherwise false if not found or not running.
-# But lsp init for cur 'buf' maybe not done yet even this func return true,
-# check 'g:LspServerReady()' for that instead.
-def g:LspServerStatus(ftype: string): bool
-  for [ft, lspserver] in ftypeServerMap->items()
-    if ftype ==# ft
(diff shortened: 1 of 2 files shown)


plugin/lsp.vim
```diff
@@ -26,6 +26,10 @@ def g:LspServerReady(): bool
   return lsp.ServerReady()
 enddef
 
+def g:LspServerStatus(ftype: string): bool
```
yegappan, line 29: What is the use case for this new function?  Can you update the help text with a description
of this function and where this can be used?

Shane-XB-Qian:
just like LspServerReady() func for ready but this one for (checking if it is) running (or not).
// this was created for #5

--
shane.xb.qian

autoload/lsp/lsp.vim
```diff
@@ -111,6 +111,17 @@ export def ShowServers()
   endfor
 enddef
 
+# Get LSP server status for filetype 'ftype'
+# Return true if running, or false if not found or not running
+export def ServerStatus(ftype: string): bool
```
yegappan, line 116: As this function returns the server running status, can you change the name of this function to ServerRunning?

plugin/lsp.vim
yegappan, line 29: As this function returns the server running status, can you change the name of this function to LspServerRunning?

Shane-XB-Qian:
renamed as your suggested.

PR #89, opened 2022-10-27
Add sent messages to TraceLog

rcasta74:
When enabling TraceLog to debug a lsp server, it is usefull to have also the messages sent to it.

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -14,7 +14,7 @@ import './callhierarchy.vim' as callhier
 
 # LSP server standard output handler
 def Output_cb(lspserver: dict<any>, chan: channel, msg: any): void
-  util.TraceLog(false, msg->string())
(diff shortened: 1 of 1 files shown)


autoload/lsp/lspserver.vim
```diff
@@ -261,6 +261,7 @@ def SendMessage(lspserver: dict<any>, content: dict<any>): void
     return
   endif
   ch->ch_sendexpr(content)
+  util.TraceLog(false, $'Sent: {content->string()}')
```
yegappan, line 264: This will result in too many messages in the trace file (as the plugin constantly sends buffer updates to the server).  We need to exclude the "textDocument/didChange" message from this trace.

rcasta74: What about to check for req.id? If I'm not wrong, it is added only to messages for which a response is expected.

PR #90, opened 2022-10-30
Correct HandleCodeAction behavior for CodeAction

atgote:
Both the `CodeAction` and `Command` interfaces contain the `command` member
but these should be handled differently for them. Corrected the
`HandleCodeAction` method to correctly handle these.

Tested with _intelephense_ PHP language server and _clangd_.


yegappan:
How did you test the 'Command' interface with clangd?  Can you provide sample steps for that?  I can add a test for it.

Shane-XB-Qian:
#95

PR #93, opened 2022-10-31
Add CompleteDone handler to execute text edits

atgote:
Add `CompleteDone` autocommand handler to verify if activated completion require additional text edits ('server-initiated edits') and to perform these.
The additional text edits are used for example when an import is required for the completed entity.

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -11,6 +11,7 @@ import './options.vim' as opt
 import './lspserver.vim' as lserver
 import './util.vim'
 import './buffer.vim' as buf
+import './textedit.vim'
 import './diag.vim'
 import './symbol.vim'
 import './outline.vim'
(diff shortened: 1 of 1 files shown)


yegappan:
Which language server did you use to test this and what are the steps to trigger the server initiated edits?

atgote:
> Which language server did you use to test this and what are the steps to trigger the server initiated edits?

I've used the *intelephense* PHP LS. The simple test environment may be as follows:

```
Project
|
+--A            # namespace A
|  |
|  +--C2.php    # class A\C2
|
+--C1.php       # class C1
```

When you try to create an instance of C2 inside the C1 class *intelephense* provides a completion with a hint *use A\C2* and if you select that completion option a new statement `use A\C2;` appears in the imports section.

There are no other necessary steps (at least for *intelephense*) - the necessary additional text edits are provided with all proposed completion options according to LSP specs - nhttps://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

Sadly I've had no success when trying to use this feature with *clangd*.

PR #101, opened 2022-11-10
Pr correct test script

Shane-XB-Qian:
refine test case script which made it more robust or correct, or fail sometime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 tags
 *.swp
+
+test/Xtest.c
+test/Xtest.cpp
+test/results.txt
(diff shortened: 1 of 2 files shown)


yegappan (commented):
Can you create a PR only with the changes to support clangd-14?  I will merge those changes and then we can debug the CI test failure with the remaining changes.

yegappan:
The CI currently uses Vim version 9.0.  The support for deleting a file after a function exits by using the 'D' flag in writefile() was added in Vim 9.0.0379.  Can you remove this change from your PR?  I want to make sure the plugin works with Vim 9.0 and not rely on the latest Vim version.

Shane-XB-Qian:
ok, but if so i am adding test/Xtest.c and Xtest.cpp into gitignore as well, since in case it failed, it maybe left on the local.
// thx, anyway, now ci passed finnally... :rofl: l

PR #105, opened 2022-11-13
Check for dict type before checking for keys

techntools:


yegappan:
Which language server did you use to reproduce this issue?  Can you describe the steps to reproduce this problem?

PR #106, opened 2022-11-14
adding 'OptionsGet' func

Shane-XB-Qian:
let user be easier to access lsp option, specially in func.

### DIFF
--- a/plugin/lsp.vim
+++ b/plugin/lsp.vim
@@ -14,8 +14,8 @@ def g:LspOptionsSet(opts: dict<any>)
   options.OptionsSet(opts)
 enddef
 
-def g:LspOptionsGet(key: string): any
-  return options.OptionsGet(key)
+def g:LspOptionsGet(): dict<any>
+  return options.OptionsGet()
 enddef
 
 def g:LspAddServer(serverList: list<dict<any>>)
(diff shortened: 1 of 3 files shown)


plugin/lsp.vim
```diff
@@ -14,6 +14,10 @@ def g:LspOptionsSet(opts: dict<any>)
   options.OptionsSet(opts)
 enddef
 
+def g:LspOptionsGet(key: string): any
```
yegappan, line 17: What is use case for this function?  Can you update doc/lsp.txt with a description of this function and how to use it?

Shane-XB-Qian: the usage mostly is for coding, since it cannot directly access lsp#options#lspOptions in user func.

Shane-XB-Qian:
any other concerns to merge this?

yegappan:
The LspOptionsSet() function accepts a Dict of options and sets them.  Similarly the LspOptionsGet() function should return a Dict of options.   So instead of returning the value of a particular option, can you change this function to return the entire lspOptions Dict? The user can extract the desired key/value from this Dict.

Shane-XB-Qian:
update: now changed to return entire dict as you wished.

PR #110, opened 2022-11-14
fix: somehow 'reply' may not have 'result' (or not yet)

Shane-XB-Qian:
maybe not ready or somehow there is no 'result'

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -612,6 +612,7 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
   endif
 
   if !reply->has_key('result')
+    util.ErrMsg($'Error(LSP): request {method} failed (no result or not yet somehow)')
     return
   endif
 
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
> > @@ -611,6 +611,10 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
>      return
>    endif
>  
> +  if !reply->has_key('result')
> +    return
> 
> Can you call util.Errmsg() to log an error message here?  Did you see this issue with the Java language server? Under what conditions, did you see this issue?

perhaps that's not a err, but just lsp server was not ready or somehow,
or whatever, let me try to add a prompt hint log.

// that's from bash lsp server when just startup vim to edit it while a the first compl.

-- 
shane.xb.qian

yegappan:
According to the [ResponseMessage](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#responseMessage) description in the language server protocol specification (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/), in a successful reply message, the result field is required and should not be present in case of an error.

Shane-XB-Qian:
i donot know, or not sure, but seems that's just happened at first init time, the latter compl request had 'result' in reply.

PR #114, opened 2022-11-17
feat: make workspace/workspaceFolders really work and correct sendResponse func

Shane-XB-Qian:
fix issue found at bash lsp server https://github.com/bash-lsp/bash-language-server/pull/535

but `workspace/workspaceFolders`  method was not really worked, 
and `sendResponse()` code logic of judge `errer` and `result` type was wrong.

this pr is trying to address and fix that.

Signed-off-by: shane.xb.qian <shane.qian@foxmail.com>

### DIFF
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -124,11 +124,11 @@ def ProcessWorkspaceFoldersReq(lspserver: dict<any>, request: dict<any>)
     lspserver.sendResponse(request, null, {})
     return
   endif
-  if empty(lspserver.workspaceFolders)
+  if lspserver.workspaceFolders->empty()
     lspserver.sendResponse(request, [], {})
   else
     lspserver.sendResponse(request,
-	  \ map(copy(lspserver.workspaceFolders), '{name: v:val->fnamemodify(":t"), uri: util.LspFileToUri(v:val)}'),
+	  \ lspserver.workspaceFolders->copy()->map('{name: v:val->fnamemodify(":t"), uri: util.LspFileToUri(v:val)}'),
 	  \ {})
   endif
 enddef
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -120,7 +120,17 @@ enddef
 # Request: "workspace/workspaceFolders"
 # Param: none
 def ProcessWorkspaceFoldersReq(lspserver: dict<any>, request: dict<any>)
-  lspserver.sendResponse(request, {}, {})
+  if !lspserver->has_key('workspaceFolders')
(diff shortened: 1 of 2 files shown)


autoload/lsp/handlers.vim
```diff
@@ -120,7 +120,17 @@ enddef
 # Request: "workspace/workspaceFolders"
 # Param: none
 def ProcessWorkspaceFoldersReq(lspserver: dict<any>, request: dict<any>)
-  lspserver.sendResponse(request, {}, {})
+  if !lspserver->has_key('workspaceFolders')
+    lspserver.sendResponse(request, null, {})
+    return
+  endif
+  if empty(lspserver.workspaceFolders)
```
yegappan, line 127: Can you use the method calling convention here?

lspserver.workspaceFolders->empty()

autoload/lsp/lspserver.vim
```diff
@@ -528,7 +528,7 @@ def SendResponse(lspserver: dict<any>, request: dict<any>, result: dict<any>, er
   endif
   var resp: dict<any> = lspserver.createResponse(
 	    request.id->type() == v:t_string ? request.id->str2nr() : request.id)
-  if result->type() != v:t_none
+  if empty(error)
```
yegappan, line 531: Can you use `error->empty()` here?

autoload/lsp/handlers.vim
```diff
@@ -120,7 +120,17 @@ enddef
...
+    lspserver.sendResponse(request, [], {})
+  else
+    lspserver.sendResponse(request,
+	  \ map(copy(lspserver.workspaceFolders), '{name: v:val->fnamemodify(":t"), uri: util.LspFileToUri(v:val)}'),
```
yegappan, line 131: Can you use `lspserver.workspaceFolders->copy()->map('....')` here?

Shane-XB-Qian:
I think method calling style in vim code had no 100% advantage, since it is Not always used the existed first placeholder parameter as the first caller, that maybe easy to make mistake.
and existing code using legacy func calling style was there many places already and looks clean also.

but anyway if you insisted, I would modify it at tomorrow morning ASAP.

--
shane.xb.qian

autoload/lsp/handlers.vim
Shane-XB-Qian: done

autoload/lsp/lspserver.vim
Shane-XB-Qian: done

autoload/lsp/handlers.vim
Shane-XB-Qian: done

Shane-XB-Qian:
update: changed to method calling style.

PR #115, opened 2022-11-17
Pr completion resolve item null

Shane-XB-Qian:
the reply.result of complete resolve perhaps is null.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -182,7 +182,7 @@ enddef
 
 # process the 'completionItem/resolve' reply from the LSP server
 # Result: CompletionItem
-export def CompletionResolveReply(lspserver: dict<any>, cItem: dict<any>)
+export def CompletionResolveReply(lspserver: dict<any>, cItem: any)
   if cItem->empty()
     return
   endif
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
if i was not reading that lsp specification wrong, looks it said reply.result of complete resolve can be list, dict, and null.
now we only handle dict, that perhaps was a problem,
for now, i noticed perl lsp server displayed as `Lazy doc` but never really updated the info doc later.
for now, not sure if that was the cause.
// anyway, you may can refine it a bit later....

yegappan:
According to the LSP specification https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem_resolve

the result of a complete resolve request is CompletionItem (a Dict).  This should
not be a List or null.  Need to see why the perl language server is returning 
some other type for the resolve request.

yegappan:
I tried reproducing this issue with the Perl language server and couldn't reproduce the issue
when using the completion menu and the completion documentation is properly displayed.
Are you using the latest version of the plugin?

I am using the https://github.com/bscan/PerlNavigator Perl language server.  Which perl
language server are you using?

Shane-XB-Qian:
it is `PLS` you can install it via cpan.
// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

--
shane.xb.qian

PR #116, opened 2022-11-28
fix: make lsp hl not be overrode by cursorline

Shane-XB-Qian:
Signed-off-by: shane.xb.qian <shane.qian@foxmail.com>

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -41,9 +41,9 @@ def LspInitOnce()
 		{name: 'LspDiagHint', text: 'H>', texthl: 'Question',
 						linehl: 'MatchParen'}])
 
-  prop_type_add('LspTextRef', {highlight: 'Search'})
-  prop_type_add('LspReadRef', {highlight: 'DiffChange'})
-  prop_type_add('LspWriteRef', {highlight: 'DiffDelete'})
+  prop_type_add('LspTextRef', {highlight: 'ErrorMsg', override: true})
+  prop_type_add('LspReadRef', {highlight: 'DiffChange', override: true})
+  prop_type_add('LspWriteRef', {highlight: 'DiffDelete', override: true})
 
   inlayhints.InitOnce()
 
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -41,7 +41,7 @@ def LspInitOnce()
 		{name: 'LspDiagHint', text: 'H>', texthl: 'Question',
 						linehl: 'MatchParen'}])
 
-  prop_type_add('LspTextRef', {highlight: 'ErrorMsg', override: true})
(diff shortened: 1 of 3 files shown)


autoload/lsp/lsp.vim
```diff
@@ -41,9 +41,9 @@ def LspInitOnce()
 		{name: 'LspDiagHint', text: 'H>', texthl: 'Question',
 						linehl: 'MatchParen'}])
 
-  prop_type_add('LspTextRef', {highlight: 'Search'})
-  prop_type_add('LspReadRef', {highlight: 'DiffChange'})
-  prop_type_add('LspWriteRef', {highlight: 'DiffDelete'})
+  prop_type_add('LspTextRef', {highlight: 'ErrorMsg', override: true})
```
yegappan, line 44: Why is the highlight group changed from 'Search' to 'ErrorMsg'?  The ErrorMsg highlight group uses a red background.  In this case as there is no error, the Search highlight group (with a yellow background) was used.

Shane-XB-Qian:
because the yellow background is very common highlight, you perhaps had manually searched something, or some keyword syntax highlight etc those default is 'yellow'. red background should be more outstanding, besides it should would disappear when winleave/bufleave, so I think it should be better in this case.

--
shane.xb.qian

Shane-XB-Qian:
or if you preferred or insisted the 'yellow', i can change it back to 'search' still, let me know.

yegappan:
Can you update the PR to include only the 'override' change for the prop_type() calls?

autoload/lsp/outline.vim
```diff
@@ -259,7 +259,7 @@ export def OpenOutlineWindow()
     highlight default link LSPTitle Title
   endif
 
-  prop_type_add('LspOutlineHighlight', {bufnr: bufnr(), highlight: 'Search'})
+  prop_type_add('LspOutlineHighlight', {bufnr: bufnr(), highlight: 'diffchange', override: true})
```
yegappan, line 262: Can you remove the 'highlight' group change here also?

test/unit_tests.vim
```diff
@@ -523,7 +523,7 @@ enddef
 # Test for :LspGotoDefinition, :LspGotoDeclaration and :LspGotoImpl
 def Test_LspGotoSymbol()
   silent! edit Xtest.cpp
-  sleep 200m
+  sleep 600m
```
yegappan, line 526: Is this change needed?

autoload/lsp/outline.vim
Shane-XB-Qian: this here probably common choice, e.g tagbar is using this (diffchange) also.
but anyway, i follow if you preferred.

test/unit_tests.vim
Shane-XB-Qian: yes, it happened (failed) sometimes, so extend it a bit.

PR #118, opened 2022-11-30
Sort the diagnostics by line number when populating the location-list

andlrc:
The function `items({dict})` returns a list in arbitrary order, and it would be nice that the errors are sorted by how the occur in the source file.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -164,7 +164,7 @@ def DiagsUpdateLocList(lspserver: dict<any>, bnr: number): bool
   var qflist: list<dict<any>> = []
   var text: string
 
-  for [lnum, diag] in lspserver.diagsMap[bnr]->items()
+  for [lnum, diag] in lspserver.diagsMap[bnr]->items()->sort((a, b) => a[0]->str2nr() - b[0]->str2nr())
     text = diag.message->substitute("\n\\+", "\n", 'g')
     qflist->add({filename: fname,
 		    lnum: diag.range.start.line + 1,


Shane-XB-Qian:
the order of loc list seems was already sorted by the line num ?

andlrc:
> the order of loc list seems was already sorted by the line num ?

I think it depends on how the `diagsMap` have been populated, consider the following two examples:

```vim
:vim9cmd echo items({1: 1, 2: 2, 3: 3, 4: 4, 5: 5, 6: 6, 7: 7, 8: 8, 9: 9, 10: 10})
[['1', 1], ['2', 2], ['3', 3], ['4', 4], ['5', 5], ['6', 6], ['7', 7], ['8', 8], ['9', 9], ['10', 10]]
```

seems good, but moving the `10: 10` to the front messes things up:

```vim
:vim9cmd echo items({10: 10, 1: 1, 2: 2, 3: 3, 4: 4, 5: 5, 6: 6, 7: 7, 8: 8, 9: 9})
[['1', 1], ['2', 2], ['3', 3], ['4', 4], ['10', 10], ['6', 6], ['7', 7], ['8', 8], ['9', 9], ['5', 5]]
```

In the case with `diagsMap` it depends on what is send by the language server in "textDocument/publishDiagnostics". As `diagsMap` is simply populated by iterating all the send diagnostics.

Shane-XB-Qian:
it was not about items() returns not-fixed-order list, but the loc list was ordered.
anyway which lsp server you were testing?

--
shane.xb.qian

andlrc:
> it was not about items() returns not-fixed-order list, but the loc list was ordered.

I don't think these are sorted, consider the following example:

```bash
$ seq 10 > a.txt
$ grep -Hn . a.txt  | sort -r > errors.err
$ cat errors.err
a.txt:9:9
a.txt:8:8
a.txt:7:7
a.txt:6:6
a.txt:5:5
a.txt:4:4
a.txt:3:3
a.txt:2:2
a.txt:1:1
a.txt:10:10
$ vim -q
```

Will open a.txt on line 9, `:cn` will bring you to line 8.

I the same holds true with the loclist. Using `a.txt` from above:

```vim
:call setloclist(0, [{ 'filename': 'a.txt', 'lnum': 9}, {'filename': 'a.txt', 'lnum':7}])
```

PR #119, opened 2022-11-30
Add g:loaded_lsp variable for easy detection

craigmac:
Useful to detect if the plugin was loaded before doing things in, for example, `.vim/after/plugin/lsp.vim` by doing:

```
if !get(g:, 'loaded_lsp') | finish | endif
```

### DIFF
--- a/plugin/lsp.vim
+++ b/plugin/lsp.vim
@@ -7,8 +7,14 @@ vim9script
 
 # Language Server Protocol (LSP) plugin for vim
 
+if exists('g:loaded_lsp')
+  finish
+endif
+g:loaded_lsp = 1
+
 import autoload '../autoload/lsp/options.vim'
 import autoload '../autoload/lsp/lsp.vim'
+
 
 # Set LSP plugin options from 'opts'.
 def g:LspOptionsSet(opts: dict<any>)


hardselius:
So it looks like it's a bad idea to bail out in the way it's done in #122. Either use `vimscript9 noclear` or just set the global variable without bailing out.

```
:h vim9-reload

						*vim9-reload* *E1149* *E1150*
When loading a legacy Vim script a second time nothing is removed, the
commands will replace existing variables and functions, create new ones, and
leave removed things hanging around.

When loading a Vim9 script a second time all existing script-local functions
and variables are deleted, thus you start with a clean slate.  This is useful
if you are developing a plugin and want to try a new version.  If you renamed
something you don't have to worry about the old name still hanging around.

If you do want to keep items, use: >
	vim9script noclear

You want to use this in scripts that use a `finish` command to bail out at
some point when loaded again.  E.g. when a buffer local option is set to a
function, the function does not need to be defined more than once: >
	vim9script noclear
	setlocal completefunc=SomeFunc
	if exists('*SomeFunc')
	  finish
	endif
	def SomeFunc()
	....
```

hardselius:
Issues with bailing out early. See https://github.com/yegappan/lsp/issues/122#issuecomment-1348762412.

Issue #87, opened 2022-10-27
Autocompletion improvements


mttigg:
I have noticed odd behaviours while using autocompletion. I'm running vim `9.0.809` and I am using `tsserver` for an lsp.

1. If you type one new character on a new line the results in the completion menu will not be filtered down by the characters already typed. For example if you have `foo` and `bar` in scope and you type the letter 'f' your first result will be `bar`. Any more key presses filters down the list.
<img width="489" alt="Screen Shot 2022-10-26 at 9 35 58 PM" src="https://user-images.githubusercontent.com/26121302/198185961-f82b5e50-9ccd-48c2-8142-28c381276192.png">
<img width="481" alt="Screen Shot 2022-10-26 at 9 36 41 PM" src="https://user-images.githubusercontent.com/26121302/198185879-9de7d06c-7d9e-42c5-8e0f-9e47ed23768d.png">

2. If you have already typed `foo` and press backspace the current list in the completion menu is removed and a new list is added to the completion menu. this list is also not filtered by what you have inserted so far.
<img width="516" alt="Screen Shot 2022-10-26 at 9 36 23 PM" src="https://user-images.githubusercontent.com/26121302/198185609-b11c70bb-da98-44ea-9c40-e905731da504.png">

I have looked at how autocompletion is implemented in this plugin and noticed that it only fetches the autocompletion results when the completion menu is closed and a valid key is pressed. And I have also noticed that the completion results from the `lsp` are not filtered down, I think that vims completion menu implementation filters this list down itself.

For the first scenario I think the following is happening. Because the result from the `lsp` for autocompletion is asynchronous we have to wait for the suggested results. The problem is the vim completion menu only filters down the results on key press, so we are only given the full list from the `lsp`.

For the second scenario I'm not completely sure what is going on. We already have the results from the `lsp`, we shouldn't need to fetch those results a second time but we do. This isn't the usual behaviour of the completion menu, usually on backspace the completion menu stays visible. I could see fetching on backspace covering some edge cases though.

andlrc:
Maybe this PR #120 is related?

mttigg:
@andlrc I think you are right. This is the same behaviour being observed with autocompletion being enabled or disabled.

techntools:
I am facing the same issue as well

![image](https://user-images.githubusercontent.com/71767293/224552505-f26d39e4-c306-42e5-84b9-2794da82f775.png)

andlrc:
[...]

PR #120, opened 2022-12-01
Support partially typing the omni completed keyword

andlrc:
Currently when using:

    call LspOptionsSet(#{ autoComplete: v:false })

Omni completion is set, but the partially typed word isn't respected when completing, consider the following:

    $ cat a.c
    int foo = 1;
    int frog = 1;
    int bar = 1;
    int main() {
      int a = f|
    }

Cursor is `|` typing `<C-x><C-o>` the list will contain all three integers: `foo`, `frog`, and `bar`. I would expect only `foo` and `frog`.

Currently it's really hard to write a test for this, as it's not possible to change the "autoComplete" option after the buffer have been initialized, as autocmds and vim options are being set in `BufferInit`.

I think one solution to this problem would be to refactor the current code, to internally react to autocmds when the options have been changed. Alternative the default for the tests should to disable the "autoComplete" option, or another test file should be created with "autoComplete" disabled.

In any case then this PR tries to address only showing the relevant omni completion suggestions.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -273,9 +273,12 @@ def g:LspOmniFunc(findstart: number, base: string): any
     # locate the start of the word
     var line = getline('.')
     var start = charcol('.') - 1
+    var keyword: string = ''
     while start > 0 && line[start - 1] =~ '\k'
+      keyword = line[start - 1] .. keyword
       start -= 1
     endwhile
+    lspserver.omniCompleteKeyword = keyword
     return start
   else
     # Wait for the list of matches from the LSP server
@@ -289,7 +292,7 @@ def g:LspOmniFunc(findstart: number, base: string): any
     endwhile
 
     var res: list<dict<any>> = lspserver.completeItems
-    return res->empty() ? v:none : res
+    return res->empty() ? v:none : res->filter((i, v) => v.word =~# '^' .. lspserver.omniCompleteKeyword)
   endif
 enddef
 


yegappan:
I am not able to reproduce this problem on a Ubuntu system with clangd 14.0 and the latest plugin version.  When I use omnicomplete the words starting with 'f', I see only 'foo' and 'frog' listed in the completion menu.  Some other option
setting is interfering with this.  Can you include the Vim options that you are enabling?

andlrc:
> I am not able to reproduce this problem on a Ubuntu system with clangd 14.0 and the latest plugin version.

Turns out that you are correct.

The problem was originally observed in a typescript file, running with [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server).

```
$ typescript-language-server --version
2.2.0
```

Can you help me figure out if this is actually an error of `typescript-language-server` and therefore something that needs to be addressed by them.

Here the error exists:

```
$ cat a.ts
const foo = 1;
const frog = 1;
const bar = 1;

const main = () => {
	const a = f|
}
```

andlrc:
This is still a problem for the mentioned typescript server. I'm still not sure if it's the language server or this plugin that is to blame.

andlrc:
Looking at the specifications for completions:

It seems like the client should be in charge of the filtering as long as `isIncomplete` isn't set to `true`:

> The language server protocol uses the following model around completions:
> 
> - to achieve consistency across languages and to honor different clients usually the client is responsible for filtering and sorting. This has also the advantage that client can experiment with different filter and sorting models. However servers can enforce different behavior by setting a `filterText` / `sortText`
> - for speed clients should be able to filter an already received completion list if the user continues typing. Servers can opt out of this using a [`CompletionList`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList) and mark it as `isIncomplete`.

The above can be interpreted in two ways:

- The returned list is filtered based on the partially typed keyword, this happens in `clangd`:

```c
struct A {
  int foo;
  int bar;
  int baz;
};

struct A my;

my.b
```

will suggest `bar` and `baz`.

- The returned list contains all possible items, and isn't filtered on anything, this is how `typescript-language-server` does it:

```
const my = {
  foo: 1;
  bar: 1;
  baz: 1;
};
my.b
```

will suggest `foo`, `bar`, `baz`.

Both these language servers sets `isIncomplete = false`.

I don't know much about vim's completion API, but for my naive testing it seems that vim does filtering on the items send to `:h complete(` as well.

andlrc:
@yegappan I think this should be merged, as it solve problems when using omnifunc, and using a language server that does no filtering at all. I'm not sure if it solves #87 though? 🤔

Issue #127, opened 2023-01-08
Alternative completion behavior


noorwachid:
Is there options to change the default vim like completion? And more like modern text editor or IDE.

Especially these two:
1. Always highlight the first match
2. Using tab to confirm completion

andlrc:
> 1. Always highlight the first match

After the `:h LspAttached` user autocmd, you can remove 'noselect' from 'completeopt':

```
augroup LspSetup
  au!
  au User LspAttached set completeopt-=noselect
augroup END

> 2. Using tab to confirm completion

You can add an insert mapping:

```
inoremap <expr> <Tab> pumvisible() ? "\<C-y>" : "\<Tab>"
```

**Comment by snailtree008** [comment]
> > 1. Always highlight the first match
> 
> After the `:h LspAttached` user autocmd, you can remove 'noselect' from 'completeopt':
> 
> ```
> augroup LspSetup
>   au!
>   au User LspAttached set completeopt-=noselect
> augroup END
> ```
> 
>  > 2. Using tab to confirm completion
> 
> You can add an insert mapping:
> 
> ```
> inoremap <expr> <Tab> pumvisible() ? "\<C-y>" : "\<Tab>"
> ```

hello,I am using your lsp as the launcher, and successfully invoked LuaLS/lua-language-server as my lua language server, but I found that the tab key and the enter key can supplement a single line of code during use. But this completion doesn't work for multi-line code:
```lua 
if a == 1 then 
end 
```

snailtree008:
> > > 1. Always highlight the first match
> > 
> > 
> > After the `:h LspAttached` user autocmd, you can remove 'noselect' from 'completeopt':
> > ```
> > augroup LspSetup
> >   au!
> >   au User LspAttached set completeopt-=noselect
> > augroup END
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > 2. Using tab to confirm completion
> > 
> > 
> > You can add an insert mapping:
> > ```
> > inoremap <expr> <Tab> pumvisible() ? "\<C-y>" : "\<Tab>"
> > ```
> 
> hello,I am using your lsp as the launcher, and successfully invoked LuaLS/lua-language-server as my lua language server, but I found that the tab key and the enter key can supplement a single line of code during use. But this completion doesn't work for multi-line code:
> 
> ```lua
> if a == 1 then 
> end 
> ```

lsp follows the supertab plugin conflict

[...]

PR #128, opened 2023-01-08
 Use tab and select first match in completion #127

noorwachid:
It can be enabled through LspOptionsSet (It really should be named LspSetOptions)

```
call LspOptionsSet({
	ignoreMissingServer: true,
	noNewlineInCompletion: true,
	noSelInCompletion: false,
	useTabInCompletion: true,
})
```

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -425,16 +425,23 @@ enddef
 export def BufferInit(lspserver: dict<any>, bnr: number, ftype: string)
   # set options for insert mode completion
   if opt.lspOptions.autoComplete
+    var noSel = opt.lspOptions.noSelInCompletion ? ',noselect' : ''
+
     if lspserver.completionLazyDoc
-      setbufvar(bnr, '&completeopt', 'menuone,popuphidden,noinsert,noselect')
+      setbufvar(bnr, '&completeopt', 'menuone,popuphidden,noinsert' .. noSel)
       setbufvar(bnr, '&completepopup', 'width:80,highlight:Pmenu,align:item,border:off')
     else
-      setbufvar(bnr, '&completeopt', 'menuone,popup,noinsert,noselect')
+      setbufvar(bnr, '&completeopt', 'menuone,popup,noinsert' .. noSel)
       setbufvar(bnr, '&completepopup', 'border:off')
     endif
     # <Enter> in insert mode stops completion and inserts a <Enter>
     if !opt.lspOptions.noNewlineInCompletion
       inoremap <expr> <buffer> <CR> pumvisible() ? "\<C-Y>\<CR>" : "\<CR>"
+    endif
+
+	# <Tab> in insert mode confirms selected item in completion list
+    if opt.lspOptions.useTabInCompletion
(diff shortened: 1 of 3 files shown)


Tachi107:
"noSelInCompletion" is a bit of an odd name. What about "selectFirstCompletion", which defaults to false? Or in any case, I'd drop the "no" prefix and flip the default value.

autoload/lsp/completion.vim
```diff
@@ -425,16 +425,23 @@ enddef
...
       setbufvar(bnr, '&completepopup', 'border:off')
     endif
     # <Enter> in insert mode stops completion and inserts a <Enter>
     if !opt.lspOptions.noNewlineInCompletion
       inoremap <expr> <buffer> <CR> pumvisible() ? "\<C-Y>\<CR>" : "\<CR>"
+    endif
+
+	# <Tab> in insert mode confirms selected item in completion list
+    if opt.lspOptions.useTabInCompletion
+      inoremap <expr> <buffer> <Tab> pumvisible() ? "\<C-Y>" : "\<Tab>"
```
andlrc, line 444: I think that these two completions should be unified to one option, possible:

```
lsp.completionTrigger = "cr" | "tab" | ""
```

Which would cause these two keybinds:

    if opt.lsp.completionTrigger == "cr"
      inoremap <expr> <buffer> <CR> pumvisible() ? "\<C-Y>\<CR>" : "\<CR>"
    elseif opt.lsp.completionTrigger == "tab"
      inoremap <expr> <buffer> <Tab> pumvisible() ? "\<C-Y>" : "\<Tab>"
    endif
    
Where `lspOptions.noNewlineInCompletion = true` would be deprecated in favor of  `opt.lsp.completionTrigger = ''` and the default would be `opt.lsp.completionTrigger = 'cr'`.

autoload/lsp/completion.vim
```diff
@@ -425,16 +425,23 @@ enddef
...
-      setbufvar(bnr, '&completeopt', 'menuone,popup,noinsert,noselect')
+      setbufvar(bnr, '&completeopt', 'menuone,popup,noinsert' .. noSel)
       setbufvar(bnr, '&completepopup', 'border:off')
     endif
     # <Enter> in insert mode stops completion and inserts a <Enter>
     if !opt.lspOptions.noNewlineInCompletion
       inoremap <expr> <buffer> <CR> pumvisible() ? "\<C-Y>\<CR>" : "\<CR>"
+    endif
+
+	# <Tab> in insert mode confirms selected item in completion list
```
Shane-XB-Qian, line 442: indent broken.

autoload/lsp/completion.vim
```diff
@@ -425,16 +425,23 @@ enddef
...
+    if opt.lspOptions.useTabInCompletion
```
Shane-XB-Qian, line 443: most likely user side had self defined or plugin defined `<tab>` mapping already, so this may override its setting.
or classic vimer style (vs so called modern IDE style) users used to `c-y`, 
so maybe just simply re-map it in vimrc anyway.

PR #131, opened 2023-01-11
fix: autoComplete hangs, add BufChange timer

dezza:
Sometimes for some servers autocomplete completely hangs as it is called repeatedly.

Adds a 300ms delay before calling again.

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -167,6 +167,10 @@ def g:LspShowSignature(): string
   return ''
 enddef
 
+def BufChangeTimerCallback(timer_id: number)
+  remove(b:, 'LspBufChangeTimer')
+enddef
+
 # buffer change notification listener
 def Bufchange_listener(bnr: number, start: number, end: number, added: number, changes: list<dict<number>>)
   var lspserver: dict<any> = buf.CurbufGetServer()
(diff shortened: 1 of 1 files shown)


dezza:
I realize now its not a good fix.

Would it be better if the timer delayed `lspserver.sendNotification('didChange')` or do you have an idea why this happens?

Shane-XB-Qian:
did not get your case/problem, but i guess you meant compl kept displaying, even 'c-y' (or 'cr') confirmed or cancelled that compl item, but it continued to show/get compl, that seems because cursormovedi, perhaps it should give a break.

dezza:
Yeah sorry, I mostly just made a quick fix and wanted to leave it before I went to bed.

Realized after it definetily should trigger in all cases.

I am unsure whether it could be syntax coughing.

So I'll keep an eye out. I have made a minimal vimrc I'll be testing against in the coming days if I see it.

Nice work on the repo btw, its starting to become quite solid and I don't have as many issues with individual servers such as bash, vim anymore.

Shane-XB-Qian:
1, syntax do may cause sth hang if complex;
2, autocomplete keep getting compl even confirm/cancel (c-y/e) that maybe someone not like;
3, bash/vim lsp server had flaw for now, e.g bashls not support codeaction yet, vimls not support vim9.

dezza:
Normally I use omni, but I'll use autocomplete from now on to see if it persists.

Also found lacygoill/vim9-syntax that helped.

Shane-XB-Qian:
> Normally I use omni

me too.

bashls mostly because it's fitting for vscode much but less testing for vim,
vimls mostly because vim9 is not completely finished dev yet.

dezza:
> > Normally I use omni
> 
> 
> 
> me too.
> 
> 
> 
> bashls mostly because it's fitting for vscode much but less testing for vim,
> 
> vimls mostly because vim9 is not completely finished dev yet.

Yes I figured 👍 Hopefully seems like bashls memory issues are closed and gone now. 8GB memory to edit shellscripts is a bit excessive even by todays standards :D

PR #138, opened 2023-03-05
fix: Use character position in TagFunc

ilango100:
While using `lsp#lsp#TagFunc`, `CTRL-]` jumps to proper line, but not the proper column. This PR makes it jump to proper column using the LSP character position.

### DIFF
--- a/autoload/lsp/symbol.vim
+++ b/autoload/lsp/symbol.vim
@@ -468,7 +468,7 @@ export def TagFunc(lspserver: dict<any>,
     var tagitem = {}
     tagitem.name = pat
     tagitem.filename = util.LspUriToFile(tagloc.uri)
-    tagitem.cmd = (tagloc.range.start.line + 1)->string()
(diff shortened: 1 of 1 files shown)


autoload/lsp/symbol.vim
```diff
@@ -468,7 +468,7 @@ export def TagFunc(lspserver: dict<any>,
     var tagitem = {}
     tagitem.name = pat
     tagitem.filename = util.LspUriToFile(tagloc.uri)
-    tagitem.cmd = (tagloc.range.start.line + 1)->string()
+    tagitem.cmd = printf("/\\%%%dl\\%%%dc", tagloc.range.start.line + 1, tagloc.range.start.character + 1)
```
yegappan, line 471: Can you change this to use string interpolation?

    tagtitem.cmd = $"/\\%{tagloc.range.start.line + 1}l\\%{tagloc.range.start.character + 1}c"

ilango100: It's done

PR #139, opened 2023-03-05
Ignore TypeScript language server custom notification

andlrc:
Ignore the "$/typescriptVersion" custom notification send by this language server:

https://github.com/typescript-language-server/typescript-language-server

### DIFF
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -89,7 +89,10 @@ export def ProcessNotif(lspserver: dict<any>, reply: dict<any>): void
       'telemetry/event': ProcessUnsupportedNotifOnce,
       # Java language server sends the 'language/status' notification which is
       # not in the LSP specification
-      'language/status': IgnoreNotif
+      'language/status': IgnoreNotif,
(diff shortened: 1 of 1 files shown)


andlrc:
I can see that some custom notifications are already ignored, but it might make sense to make an interface like the one for `:h lsp-custom-commands`:

```
g:LspRegisterNotificationHandler('$/typescriptVersion', (lspserver: dict<any>, reply: dict<any>) => {
 echom printf("Typescript version = %s", reply.version)
})
```

Or it might make sense to be able to provide a `customNotificationHandlers` property to the AddServer dictionary:

```
g:LspAddServer([
  {
    filetype: ['javascript', 'typescript'],
    path: 'typescript-language-server',
    args: ['--stdio'],
    customNotificationHandlers: {
      '$/typescriptVersion': (lspserver: dict<any>, reply: dict<any>) => {
        # ignore message
      }
    }
  }
])
```

What do you think?

yegappan:
I like the idea of supporting custom notification handlers.  Can you create a PR for this change?

PR #140, opened 2023-03-05
Add support for "customNotificationHandlers"

andlrc:
We need to discuss all the details, currently this MR adds a new property to the "AddServer" dictionary, which when provided will extend and override existing notification handlers.

1. Does it make sense to override the existing notification handlers, or should it only be possible to provide new?
2. Does it make sense to "leak" the internal API `(lspserver, reply)`? Or should only `reply.params` be send?

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,9 @@ To use the plugin features with a particular file type(s), you need to first reg
 
 To register a LSP server, add the following lines to your .vimrc file (use only the LSP servers that you need from the below list).  If you used [vim-plug](https://github.com/junegunn/vim-plug) to install the LSP plugin, the steps are described later in this section.
 ```
+   function! TypeScriptCustomNotificationHandler(lspserver, reply) abort
+     echom printf("TypeScript Version = %s", reply.params.version)
+   endfunction
    let lspServers = [
 		\     #{
 		\        filetype: ['c', 'cpp'],
@@ -59,6 +62,9 @@ To register a LSP server, add the following lines to your .vimrc file (use only 
 		\	 filetype: ['javascript', 'typescript'],
 		\	 path: '/usr/local/bin/typescript-language-server',
 		\	 args: ['--stdio']
+		\	 customNotificationHandlers: {
+		\	   '$/typescriptVersion': function('TypeScriptCustomNotificationHandler')
+		\	 }
 		\      },
 		\     #{
 		\	 filetype: 'sh',
@@ -115,6 +121,7 @@ filetype|One or more file types supported by the LSP server.  This can be a Stri
 path|complete path to the LSP server executable (without any arguments).
 args|a list of command-line arguments passed to the LSP server. Each argument is a separate List item.
 initializationOptions|User provided initialization options. May be of any type. For example the *intelephense* PHP language server accept several options here with the License Key among others. 
+customNotificationHandlers|A dictionary of notifications and functions that can be specified to add support for custom language server notifications.
 
 The LSP servers are added using the LspAddServer() function. This function accepts a list of LSP servers with the above information.
 
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -73,10 +73,6 @@ def ProcessUnsupportedNotifOnce(lspserver: dict<any>, reply: dict<any>)
   endif
 enddef
 
-# silently ignore an unsupported notification message
-def IgnoreNotif(lspserver: dict<any>, reply: dict<any>)
-enddef
-
 # process notification messages from the LSP server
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -59,6 +59,7 @@ To register a LSP server, add the following lines to your .vimrc file (use only
 		\	 filetype: ['javascript', 'typescript'],
 		\	 path: '/usr/local/bin/typescript-language-server',
 		\	 args: ['--stdio']
+		\	 customNotificationHandlers: { '$/typescriptVersion': { lspserver, reply -> 0 }
```
andlrc, line 62: Is this a valid vim8 lambda?

README.md
```diff
@@ -115,6 +116,7 @@ filetype|One or more file types supported by the LSP server.  This can be a Stri
 path|complete path to the LSP server executable (without any arguments).
 args|a list of command-line arguments passed to the LSP server. Each argument is a separate List item.
 initializationOptions|User provided initialization options. May be of any type. For example the *intelephense* PHP language server accept several options here with the License Key among others. 
+customNotificationHandlers|A dictionary of notifications and functions that will be called instead of the default notifcation handler. This is also useful if you want to support a custom notification from a given language server
```
andlrc, line 119: This  message might need a bit of work?

doc/lsp.txt
```diff
@@ -220,6 +220,14 @@ To add a language server, the following information is needed:
...
 			this dictionary and if present will be transmitted to
 			the lsp server.
+	customNotificationHandlers
+			(Optional) some lsp servers (e.g.
+			typescript-language-server) will send additional
+			notifications which you might want to silence or handle,
+			or you would like to override the default notification
+			handler, in either case this can be done by sending a
+			dictionary with a notification name and a function,
+			which will then be called instead.
```
andlrc, line 230: This message might need a bit of work?

autoload/lsp/handlers.vim
```diff
@@ -92,7 +92,9 @@ export def ProcessNotif(lspserver: dict<any>, reply: dict<any>): void
       'language/status': IgnoreNotif
     }
 
-  if lsp_notif_handlers->has_key(reply.method)
+  if lspserver.customNotificationHandlers->has_key(reply.method)
+    lspserver.customNotificationHandlers[reply.method](lspserver, reply)
+  elseif lsp_notif_handlers->has_key(reply.method)
```
andlrc, line 97: 1. If a custom notification handler is provided it will ignore the built-ins Does that make sense?
2. Currently the internal API `(lspserver, reply)` is leaked. Does that make sense?

README.md
```diff
@@ -59,6 +59,7 @@ To register a LSP server, add the following lines to your .vimrc file (use only
 		\	 filetype: ['javascript', 'typescript'],
 		\	 path: '/usr/local/bin/typescript-language-server',
 		\	 args: ['--stdio']
+		\	 customNotificationHandlers: { '$/typescriptVersion': { lspserver, reply -> echom printf("TypeScript Version = %s", reply.params.version) }
```
andlrc, line 62: I don't think this is a valid vim8 lambda

autoload/lsp/handlers.vim
yegappan: 1. If the built-in handler properly handles a notification (e.g. textDocument/publishDiagnostics), then the custom notification handler cannot override the built-in handler.  But if the built-in handler doesn't handle a notification or ignores it (e.g. language/status, $/typescriptVersion, etc.), then the custom handler can override it.
2. If the custom handler needs to send a message to the lsp server while processing the notification, then it will need "lspserver".  So passing "lspserver" and "reply" as parameter makes sense.

andlrc:
I would like to add a test for this change. But I'm not sure how I would go about finding a language server that comes with custom notifications and which is also easy to install and run in the test runner.

Any suggestions?

Tachi107:
If I'm not mistaken jdtls seems to send those. It is not packaged by Debian, but you can [download the archive](https://download.eclipse.org/jdtls/milestones/1.20.0/) in the runner, extract it and symlink `bin/jdtls` to `/usr/local/`

andlrc:
Here I though that I found a lsp server which could be installed with `apt`, but turns out it's so old `v.1.3.3` that the `$/progress` notification isn't triggered.

yegappan:
Can you update the github action to install some other language server that supports the `$/progress` notification?

andlrc:
> Can you update the github action to install some other language server that supports the `$/progress` notification?

I'm not which one it should be? It's my understanding that `clangd` will use that notification when indexing, I havn't been able to get it to do that though. https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/ClangdLSPServer.cpp#L1739

PR #142, opened 2023-03-09
Add support for "capabilities.*.linkSupport"

### LINKED ISSUE #141: Consider adding support for client capability: `textDocument.*.linkSupport`.
In version 3.14.0 of the LSP specification the `linkSupport` was added as a boolean to four `textDocument` properties: `definition`, `declaration`, `typeDefinition` and `implementation`. The current specification is linked here:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink

It doesn't seem that `LocationLink` is widely supported currently. But it's my understanding that the difference between `Location` and `LocationLink` is that the later will contain a range which will include any relevant comments, etc, which could be used for the peek commands.

I haven't found a single language server that returns a different range when `linkSupport` is send, but I have found one which only supports `LocationLink`, which means that it doesn't work with this lsp client. (See issue: https://github.com/angular/vscode-ng-language-service/issues/1874)

The angular language server mentioned above also sends a lot of custom notifications, but these could be ignored by the user if #140 gets merged.

I suggest that we add support for `LocationLink` the following way:

- always send the `linkSupport` property when sending client capabilities.
- refactor the relevant code parts to have a fall-though approach when deducting `uri` and `range`:

```vim9script
var uri: string
if location->has_key('targetUri')
  # LocationLink
  uri = location.targetUri
else
  # Location
  uri = location.uri
endif
```

At a later stage we could consider refactoring the `PeekSymbolLocation` function to use the `LocationLink.targetRange` if it differs from `LocationLink.targetSelectionRange`, and show all the code, instead of creating a popup with height = 10 and centering the symbol.

andlrc:
This change should be backward compatible, but allow us to use all the properties from "LocationLink" in other commits, as long the language server replies with "LocationLink".

Closes #141

### DIFF
--- a/autoload/lsp/util.vim
+++ b/autoload/lsp/util.vim
@@ -50,6 +50,17 @@ export def ClearTraceLogs()
   writefile([], $'{lsp_log_dir}lsp-server.err')
 enddef
 
+# Convert a 'Location' to a dict that looks a lot like a 'LocationLink' but with
+# the caveat that 'targetRange' is null.
+export def LspLocationToLocationLink(location: dict<any>): dict<any>
+  return {
+    targetUri: location.uri,
+    targetSelectionRange: location.range,
+    targetRange: null,
+    originSelectionRange: null
+  }
+enddef
+
 # Convert a LSP file URI (file://<absolute_path>) to a Vim file name
 export def LspUriToFile(uri: string): string
   # Replace all the %xx numbers (e.g. %20 for space) in the URI to character
@@ -154,7 +165,7 @@ enddef
 # number and character number. The user specified window command modifiers
 # (e.g. topleft) are in 'cmdmods'.
 export def JumpToLspLocation(location: dict<any>, cmdmods: string)
-  var fname = LspUriToFile(location.uri)
+  var fname = LspUriToFile(location.targetUri)
 
   # jump to the file and line containing the symbol
   if cmdmods == ''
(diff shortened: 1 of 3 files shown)


autoload/lsp/util.vim
```diff
@@ -154,7 +154,20 @@ enddef
...
+
+  if location->has_key('targetUri')
+    # LocationLink
+    uri = location.targetUri
+    range = location.targetSelectionRange
+  else
+    # Location
+    uri = location.uri
+    range = location.range
+  endif
```
yegappan, line 168: As this code is repeated a few times, can you move this to a function in the util file and use that?  This function will take a LSP location and return [uri, range]?

andlrc: In that case I think we should return three values: `targetUri, targetRange` and `targetSelectionRange`. Possible making `targetRange` empty when mapping over `Location`.

See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink

yegappan:
This change introduced a regression.  For example, the ":LspPeekReferences" command is broken (because the Location in the result is not translated to LocationLink).

andlrc:
I think it's the `util.JumpToLspLocation` function that now expects a LocationLink instead of a Location?

yegappan:
Yes.  I pushed 3e39e8202981bd095d31717388e9058d27d929bc now to fix this issue.

andlrc:
Cool thanks, moving the parsing as close as possible to the usage makes sense. :+1:

PR #144, opened 2023-03-10
Hide disabled code actions

andlrc:
As of 3.16.0 of the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction) a code action can contain a `disabled` object, sniff on this object and remove disabled code actions if the `hideDisabledCodeActions` option is set.

### DIFF
--- a/autoload/lsp/codeaction.vim
+++ b/autoload/lsp/codeaction.vim
@@ -43,7 +43,13 @@ export def HandleCodeAction(lspserver: dict<any>, selAction: dict<any>)
   endif
 enddef
 
-export def ApplyCodeAction(lspserver: dict<any>, actions: list<dict<any>>): void
+export def ApplyCodeAction(lspserver: dict<any>, actionlist: list<dict<any>>): void
+  var actions = actionlist
+
+  if opt.lspOptions.hideDisabledCodeActions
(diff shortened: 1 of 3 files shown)


andlrc:
I think it's up the each language server to decide if they send all thinkable code actions or only the once that makes sense to apply in a given context. It's my understanding that `clangd` only sends the once that can be applied, while the typescript-language-server sends all available actions.

autoload/lsp/codeaction.vim
```diff
@@ -43,7 +43,13 @@ export def HandleCodeAction(lspserver: dict<any>, selAction: dict<any>)
   endif
 enddef
 
-export def ApplyCodeAction(lspserver: dict<any>, actions: list<dict<any>>): void
+export def ApplyCodeAction(lspserver: dict<any>, actionlist: list<dict<any>>): void
+  var actions = actionlist
+
+  if opt.lspOptions.hideDisabledCodeActions
+    actions = actions->filter((ix, act) => !act->has_key('disabled'))
+  endif
```
yegappan, line 51: In addition to this, the "disabledSupport" field in the client "CodeActionClientCapabilities" should be set to true.

PR #147, opened 2023-03-11
Make it possible to provide a new name directly to "LspRename"

andlrc:
Following on #146 I think it makes sense to be able to provide a new name directly to the rename command:

    :LspRename NewName


test/unit_tests.vim
```diff
@@ -445,6 +445,25 @@ def Test_LspRename()
...
+	++countvar;
+    }
+
+    void F2(int count)
+    {
+	count = 5;
+    }
+  END
+  assert_equal(expected2, getline(1, '$'))
+  sleep 100m
```
andlrc, line 466: I think the tests are flaky, as a new `textDocument.publishDiagnostics` will be received after a rename. Causing the test `Test_LspDiagsUpdated_Autocmd` to pick up one to many autocmd triggers. The 100ms delay seems to delay the execution enough for the ls to be able to send `textDocument.publishDiagnostics` before continuing onto the tests in `Test_LspDiagsUpdated_Autocmd`

Issue #143, opened 2023-03-09
The `textEdit.newText` is used without respect for `textEdit.range` when completing text.


andlrc:
Relevant Specification: [`CompletionItem`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem).

The `textEdit.newText` is used without respect for `textEdit.range`. This causes problems for typescript with `autoComplete: false` as the `textEdit.range` will be before the computed `start_col`, in cases where nothing filtering have been typed:

```
const a = {
  foo: 1,
  bar: 1,
  baz: 1,
};
a.
```

Will suggest `.foo`, `.bar`, and `.baz`, making the result something like: `a..foo`, as the `textEdit.range` points to the end of `a` while `start_col` points to the end of `.`.

A possible fix for this is to compute the `start_col` before iterating `CompletionItem`s as the `newText` would then be able to be sliced: `d.word = textEdit.newText[ start_col - textEdit.range.start.character : ]`. But I'm not sure this is correct to do either?

Another approach would be to not pick-up `d.word` from there and only use `insertText` and `label`.

PR #151, opened 2023-03-13
Don't use textEdit.newText, as textEdit.range isn't respected

andlrc:
This is one way to address #143


yegappan:
Can you try the following patch to see whether this addresses the problem?

```
index 03c29ab..da2be89 100644
@@ -85,6 +85,9 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     items = cItems.items
   endif
 
+  # Get the keyword before the current column
+  var prevKeyword = matchstr(getline('.'), $'\k\+\%{col(".")}c')
+
   var completeItems: list<dict<any>> = []
   for item in items
     var d: dict<any> = {}
@@ -99,6 +102,12 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
       # snippet completion.  Needs a snippet plugin to expand the snippet.
       # Remove all the snippet placeholders
+    else
+      # plain text completion.  If the completion item text doesn't start with
+      # the current keyword prefix, skip it.
+      if stridx(d.word, prevKeyword) == -1
+       continue
+      endif
     endif
     d.abbr = item.label
     d.dup = 1
```

andlrc:
> Can you try the following patch to see whether this addresses the problem?

It doesn't work:

>     +  var prevKeyword = matchstr(getline('.'), $'\k\+\%{col(".")}c')

When the completion is triggered on `obj.` where the cursor is just after the ".",  `prevKeyword` will be empty.

yegappan:
In this case, all the members of "obj" need to be displayed in the completion menu and there is no need to filter the matches returned by the language server.  Are you seeing that other members are displayed in the completion menu?  Do you have some some code to illustrate the problem?

andlrc:
> In this case, all the members of `obj` should be displayed in the completion menu and there is no need to filter the matches returned by the language server.

The problem is that `textEdit.newText` contains `.foo`, `.bar`, `.baz` resulting in `obj..foo`, `obj..bar`, `obj..baz`.

The completion starts after the first period ("."), but `textEdit.range.start` points to the column one to the left, and it therefore makes sense for `textEdit.newText` to include the period, but we simply append `textEdit.newText` to the document, and don't respect `textEdit.range`.

Shane-XB-Qian:
try PR #159

andlrc:
> try PR https://github.com/yegappan/lsp/pull/159

Doesn't work. As The typescript-lanaguage-server doesn't send '.' as a trigger character! And even if it worked it's not how textEdit should be used. See the relevant specifications: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

PR #152, opened 2023-03-13
Scan for the closest identifier under the cursor when resolving cursor location

andlrc:
This works somewhat like "CTRL-]" and "CTRL-R_CTRL-W".

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -789,22 +789,39 @@ def TextdocDidChange(lspserver: dict<any>, bnr: number, start: number,
 enddef
 
 # Return the current cursor position as a LSP position.
+# find_ident will search for a identifier in front of the cursor, just like
+# CTRL-] and c_CTRL-R_CTRL-W does.
+#
 # LSP line and column numbers start from zero, whereas Vim line and column
 # numbers start from one. The LSP column number is the character index in the
 # line and not the byte index in the line.
-def GetLspPosition(): dict<number>
+def GetLspPosition(find_ident: bool): dict<number>
   var lnum: number = line('.') - 1
   var col: number = charcol('.') - 1
+  var line = getline('.')
+
+  if find_ident
+    # 1. skip to start of identifier
+    while line[col] != '' && line[col] !~ '\k'
+      col = col + 1
+    endwhile
+
+    # 2. back up to start of identifier
+    while col > 0 && line[col - 1] =~ '\k'
+      col = col - 1
+    endwhile
+  endif
+
   return {line: lnum, character: col}
 enddef
 
 # Return the current file name and current cursor position as a LSP
 # TextDocumentPositionParams structure
-def GetLspTextDocPosition(): dict<dict<any>>
+def GetLspTextDocPosition(find_ident: bool): dict<dict<any>>
   # interface TextDocumentIdentifier
   # interface Position
   return {textDocument: {uri: util.LspFileToUri(@%)},
-	  position: GetLspPosition()}
+	  position: GetLspPosition(find_ident)}
 enddef
 
 # Get a list of completion items.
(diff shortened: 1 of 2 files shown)


autoload/lsp/lspserver.vim
```diff
@@ -792,19 +792,37 @@ enddef
...
 # line and not the byte index in the line.
-def GetLspPosition(): dict<number>
+def GetLspPosition(find_ident: bool): dict<number>
   var lnum: number = line('.') - 1
   var col: number = charcol('.') - 1
+  var line = getline('.')
+
+  if find_ident
+    # 1. skip to start of identifier
+    while line[col] != '' && line[col] !~ '\k'
```
yegappan, line 802: I think a check is needed to make sure "col" is less then line->len().

autoload/lsp/lspserver.vim
```diff
@@ -792,19 +792,37 @@ enddef
...
+      col = col + 1
+    endwhile
+
+    # 2. back up to start of identifier
+    while col > 1 && line[col - 1] =~ '\k'
```
yegappan, line 807: Should this check be changed to "col > 0" (to account for the first column)?

autoload/lsp/lspserver.vim
```diff
@@ -792,19 +792,37 @@ enddef
...
+      col = col - 1
+    endwhile
+
+    if line[col] == '' || col == 1 && line[col] !~ '\k'
```
yegappan, line 811: Note that charcol() returns the character number.  But the indexing of line[] is done here by using the byte index.  I am not sure whether this will cause a problem if multibyte characters are used in a keyword.

autoload/lsp/lspserver.vim
```diff
@@ -792,19 +792,37 @@ enddef
 # LSP line and column numbers start from zero, whereas Vim line and column
 # numbers start from one. The LSP column number is the character index in the
 # line and not the byte index in the line.
-def GetLspPosition(): dict<number>
```
yegappan, line 795: Can you add some comment here describing the purpose of the "find_ident" argument?

autoload/lsp/lspserver.vim
andlrc: Two things:

- Seems that in vim9 script `str[n]` works on characters, see `:h expr-[]`.

- Turns out that the character count should be based on utf-16 strings 🤕 https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocuments Making `😊` take up 2 characters instead of 1, or 4, which means this test will fail:

```
index f0397c8..fcff366 100644
@@ -795,6 +795,7 @@ enddef
 def GetLspPosition(): dict<number>
   var lnum: number = line('.') - 1
   var col: number = charcol('.') - 1
+  col += 4
   return {line: lnum, character: col}
 enddef
 
diff --git a/test/unit_tests.vim b/test/unit_tests.vim
index cbdf008..995f6a7 100644
--- a/test/unit_tests.vim
+++ b/test/unit_tests.vim
@@ -979,6 +979,40 @@ def Test_LspCustomNotificationHandlers()
   :%bw!
 enddef
 
+def Test_UnicodeColumnCalc()
+  :silent! edit Xtest.c
+  sleep 200m
+  var lines: list<string> =<< trim END
+    int count;
+    int fn(int a)
+    {
+      int 😊😊😊😊;
+      😊😊😊😊 = a;
+
+      int b;
+      b = a;
+      return    count + 1;
+    }
+  END
+  setline(1, lines)
+  :redraw!
+
+  # Test Unicode column calculation
+  cursor(5, 1) # 😊😊😊😊 = a;
+  search('a')
+  assert_equal([],
+              execute('LspGotoDefinition')->split("\n"))
+  assert_equal([2, 12], [line('.'), col('.')])
+
+  cursor(8, 1) # b = a;
+  search('a')
+  assert_equal([],
+              execute('LspGotoDefinition')->split("\n"))
+  assert_equal([2, 12], [line('.'), col('.')])
+
+  bw!
+enddef
+
```

autoload/lsp/lspserver.vim
andlrc: yes

autoload/lsp/lspserver.vim
andlrc: take a look and tell me if it makes sense .

autoload/lsp/lspserver.vim
andlrc: I think that `line[col] != ''` takes care of it, as `'abc'[123456] == ''`

yegappan: Yes. You are correct.  There is no need to check for the boundary condition here.

PR #153, opened 2023-03-14
Fix the unit test badge src

andlrc:
Shouldn't it be like this?


yegappan:
Yes. It is a cut-and-paste error from my other repository :-).  I don't know how I missed this so far.

PR #154, opened 2023-03-14
fix: correct vim9 'autoload' syntax

Shane-XB-Qian:


Shane-XB-Qian:
looks would make path in `test` a bit complex, so closing...

PR #157, opened 2023-03-15
fix: should allow autocompl items icase work and be more smooth

Shane-XB-Qian:


Shane-XB-Qian:
no, this should be NOT regarding to those, but just got the prepared compl items which not only case sensitive items.

-- 
shane.xb.qian

autoload/lsp/completion.vim
```diff
@@ -106,8 +106,8 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
       d.word = MakeValidWord(d.word)
     else
       # plain text completion.  If the completion item text doesn't start with
-      # the current keyword prefix, skip it.
-      if prefix != '' && stridx(d.word, prefix) != 0
+      # the current (case ignored) keyword prefix, skip it.
+      if prefix != '' && d.word !~? $'^{prefix}'
```
yegappan, line 110: Doing a regular expression comparison is more expensive than doing a simple string comparison.  We can convert "prefix" to lower case using tolower() outside this loop and then convert "d.word" to lower case using tolower() and then use stridx().

Shane-XB-Qian: // done
ok, regex was more readable, but yea, perhaps was a perf shortness if large set.

Issue #130, opened 2023-01-10
Double dollar sign insertion


noorwachid:
https://user-images.githubusercontent.com/42460975/211540929-04ec095d-cb95-474a-b843-4d82bfd5d72b.mov

andlrc:
What language is this? What language server is used? And how is it configured?

andlrc:
Seems like it's related to #143

yegappan:
I think this is PHP.  Can you attach the complete code sample, so that it will be easier to reproduce the problem?  Can you also mention the language server name?

yegappan:
Few changes are recently made to the completion code.  Can you try to reproduce this problem with the latest version of the plugin?

noorwachid:
Status: Still occuring
Vim: 9
Plugin: Latest Version
Language: PHP
LSP: intelephense

``` php
<?php 

$long_variable_name = 'noorwachid';

$$long_variable_name 
```

https://user-images.githubusercontent.com/42460975/225661965-2894160c-7d30-4f1a-a5bc-3264e7bf0316.mov

andlrc:
@noorwachid I think the following diff might solve your problem, but I'm not sure how it will behave in other cases:

```
diff --git a/autoload/lsp/completion.vim b/autoload/lsp/completion.vim
index 8ab8d6b..7c25f8d 100644
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -90,12 +90,29 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
   var starttext = getline('.')[ : col - 1]
   var prefix = tolower(matchstr(starttext, '\k*$'))
 
+  var start_col: number = 0
+
+  # LSP server didn't return a starting position for completion, search
+  # backwards from the current cursor position for a non-keyword character.
+  if start_col == 0
+    var line: string = getline('.')
+    var start = col('.') - 1
+    while start > 0 && line[start - 1] =~ '\k'
+      start -= 1
+    endwhile
+    start_col = start + 1
+  endif
+
   var completeItems: list<dict<any>> = []
   for item in items
     var d: dict<any> = {}
     # TODO: Add proper support for item.textEdit.newText and item.textEdit.range
     # Keep in mind that item.textEdit.range can start be way before the typed keyword.
-    if item->has_key('insertText')
+    if item->has_key('textEdit')
[...]

PR #159, opened 2023-03-17
fix: some lsp server may include trigger char into compl item

Shane-XB-Qian:


Shane-XB-Qian:
textedit.rang designed for multiple lines something, inserttext also, wish you can implement such logic into this lsp one day.
but so far I just wish it worked for most cases, specifically no outage err.

--
shane.xb.qian

andlrc:
> textedit.rang designed for multiple lines something, inserttext also

`textEdit.range` is not, not for completion. See the relevant specification: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

See 

> 	  * *Note:* The range of the edit must be a single line range and it must
>	  * contain the position at which completion has been requested.

Shane-XB-Qian:
> If it's kept and the range isn't respected then this error appears:

> https://asciinema.org/a/054j5b7Vg0XlSQQRKfrwuhDld


have you ever tried this pr?
where you saw this err?
we did not such `rang` here, though maybe need concern this one day.
--
shane.xb.qian

andlrc:
> have you ever tried this pr?

Yes

> where you saw this err?

The same errors as I linked above.

I recorded a video for you: https://asciinema.org/a/jW7zKNZRcnN6uLMSmRcNKaE9K

> we did not such `rang` here, though maybe need concern this one day.

Yes and that is a problem, because you NEED to respect the range if you want to use the value from `textEdit.newText`. Please take a look at the linked specifications.

andlrc:
I also recorded this for you, which is using the php language server mentioned in #130 

https://asciinema.org/a/6lZVci3U1W2afZgwqkpxHXmwL

Shane-XB-Qian:
have you recorded on correct commit base and branch?
why I saw it is omni compl?
or even it is, I cannot reproduce.

--
shane.xb.qian

Shane-XB-Qian:
not sure if still worth, but to phpactor, @yegappan your searched suggestion looks was better one, so closing..

PR #160, opened 2023-03-17
fix: omnifunc broken

Shane-XB-Qian:
PR #120 made omni complete be broken,
this makes icase work again, though if one day to support fuzzy filter, then this req refine again.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -301,7 +301,9 @@ def g:LspOmniFunc(findstart: number, base: string): any
     endwhile
 
     var res: list<dict<any>> = lspserver.completeItems
-    return res->empty() ? v:none : res->filter((i, v) => v.word =~# '^' .. lspserver.omniCompleteKeyword)
+    var prefix: string = lspserver.omniCompleteKeyword->tolower()
+    # To filter (case ignored) keyword prefixed compl items only.
+    return res->empty() ? v:none : res->filter((i, v) => v.word->tolower()->stridx(prefix) == 0)
   endif
 enddef
 


andlrc:
What language server replies with case folded matches?

yegappan:
Which language server shows this problem? Can you add a sample code that illustrates the problem with omni-completion?

Shane-XB-Qian:
you can play with gopls (golang), and wrap
your/this lsp Omnifunc with icase=1,
when type lower case prefix keyword, you will see none compl items there.

--
shane.xb.qian

Shane-XB-Qian:
@yegappan i think better to rollback this and pr #120 did, since compl items had been filter by/in `CompletionReply()` already, and some lsp server e.g gopls maybe `item->get('insertTextFormat', 1) == 2` but `item.kind` is not 15, such things made the filter be complex and difficult to judge, besides it had done the filter check already, no needed to do it twice.

Shane-XB-Qian:
i think this better be merged.

yegappan:
Do you have a sample .go file and the steps to reproduce the problem?  I can try to reproduce this in my local setup.

Shane-XB-Qian:
```bash
mkdir foo
cd foo
vi x.go
```

x.go:
```go
package main

import "log"

func main() {
	log.fa|
}
```

`|` is the cursor place, check when you typing `.fa` and omni at `|`.

Shane-XB-Qian:
looks i found the problem after bit to bit debug.. -_-#
1, firstly those filter added was unnecessary and incorrect, i have desc above.
2. the really problem looks just the start col of omni compl was wrong (but please check again, i maybe wrong)

andlrc:
So turns out that `gopls` can be configured to ignore case, fuzzy match or be case sensitive:

```
  {
    filetype: ['go'],
    path: 'gopls',
    initializationOptions: {
      matcher: 'CaseInsensitive'
    }
  },
```

The default is "Fuzzy":

https://github.com/golang/tools/blob/master/gopls/doc/settings.md#matcher-enum

Which means that doing a case sensitive match on the keyword in the omni completion will not work when using `gopls`.

I cannot see that they report any capabilities in this regard though. And I don't even though that the specifications supports reporting it. The `isIncomplete` is set to `true` though, meaning that we should request new completions for every word typed.

Really the `isIncomplete` should be respected, as to let the server compute the completions. I also think it might make sense to respect the `isIncomplete` when filtering out omni completion results, as the specifications states:

> for speed clients should be able to filter an already received completion list if the user continues typing. Servers can opt out of this using a [`CompletionList`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList) and mark it as `isIncomplete`.

The correct thing to do would be something like this https://github.com/yegappan/lsp/pull/163

Shane-XB-Qian:
```typescript
const a = {
  foo: 1,
  bar: 1,
  baz: 1,
};
a.bx|
```

it's not regarding to gopls or golang, but just it had 'fuzzy' feature;

the only problem is you raised in previous ticket the ts case, e.g above code, the cursor at `|` place, then omni, it would pop up compl items, but normally it should be None items! (no one matched `bx`)

so i changed my mind, still filter at omni side, after so, looks every looks fine so far.

andlrc:
> it's not regarding to gopls or golang, but just it had 'fuzzy' feature;

If fuzzy search should be support for all omni completions, then it should be a configuration option, and the vim fuzzy API should be used, something like this:

```
return res->empty() ? v:none : res->matchfuzzy(prefix, { key: 'word' })
```

But that doesn't change that we should in general respect the `isIncomplete` flag.

andlrc:
> If fuzzy search should be support for all omni completions, then it should be a configuration option, and the vim fuzzy API should be used, something like this:
>
>     return res->empty() ? v:none : res->matchfuzzy(prefix, { key: 'word' })

And even so, some language servers already filters the list eg `clangd`, which means that fuzzy matching would only work for servers which relies on the client to do the filtering, eg. `typescript-language-server`.

Shane-XB-Qian:
@yegappan could you reopen this pr, i'd like to rebase and got this be merged.

yegappan:
The "Reopen and comment" button is disabled.  I am not sure how to reopen this PR.  Can you open a new PR?

Shane-XB-Qian:
that due to pr #163 marks `closes #160` (to close this pr) in its ticket............. -_-#
// ok, i will try to open a new one.

PR #161, opened 2023-03-17
Add naive support for textEdit

### LINKED ISSUE #130: Double dollar sign insertion

andlrc:
This is a very naive way to add support for `CompletionItem.textEdit`, it addresses the problem mentioned in #130  but doesn't respect the specifications, as the range isn't actually respected but simply used to slice the beginning of the returned `textEdit.insertText`. 

Closes #130

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -86,20 +86,38 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
   endif
 
   # Get the keyword prefix before the current cursor column.
-  var col = charcol('.')
-  var starttext = getline('.')[ : col - 1]
-  var prefix = starttext->tolower()->matchstr('\k*$')
+  var chcol = charcol('.')
+  var starttext = getline('.')[ : chcol - 1]
+  var [prefix, start_idx, end_idx] = starttext->tolower()->matchstrpos('\k*$')
+  var start_col = start_idx + 1
 
   var completeItems: list<dict<any>> = []
   for item in items
     var d: dict<any> = {}
+
     # TODO: Add proper support for item.textEdit.newText and item.textEdit.range
     # Keep in mind that item.textEdit.range can start be way before the typed keyword.
-    if item->has_key('insertText')
+    if item->has_key('textEdit')
+      var start_charcol: number
+      if prefix != ''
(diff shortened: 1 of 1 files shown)


yegappan:
I made some modifications to your changes to optimize the character position computation.  Can you try this out?

```
index 07c6480..1bfc1a9 100644
@@ -86,16 +86,24 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
   endif
 
   # Get the keyword prefix before the current cursor column.
-  var col = charcol('.')
-  var starttext = getline('.')[ : col - 1]
-  var prefix = starttext->tolower()->matchstr('\k*$')
+  var chcol = charcol('.')
+  var starttext = getline('.')[ : chcol - 1]
+  var [prefix, start_col, end_col] = starttext->tolower()->matchstrpos('\k*$')
+  var start_charcol = charidx(starttext, start_col) + 1
 
   var completeItems: list<dict<any>> = []
   for item in items
     var d: dict<any> = {}
     # TODO: Add proper support for item.textEdit.newText and item.textEdit.range
     # Keep in mind that item.textEdit.range can start be way before the typed keyword.
-    if item->has_key('insertText')
+    if item->has_key('textEdit')
+      if item.textEdit.range.start.character != start_charcol
+       var offset = start_charcol - item.textEdit.range.start.character
+       d.word = item.textEdit.newText[ offset : ]
+      else
+       d.word = item.textEdit.newText
+      endif
+    elseif item->has_key('insertText')
       d.word = item.insertText
     else
       d.word = item.label
@@ -158,19 +166,6 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
       return
     endif
 
-    var start_col: number = 0
-
-    # LSP server didn't return a starting position for completion, search
-    # backwards from the current cursor position for a non-keyword character.
-    if start_col == 0
-      var line: string = getline('.')
-      var start = col('.') - 1
-      while start > 0 && line[start - 1] =~ '\k'
-       start -= 1
-      endwhile
-      start_col = start + 1
-    endif
-
     completeItems->complete(start_col)
   else
     lspserver.completeItems = completeItems
`

andlrc:
Yes, it works for my testings. I have updated the branch

PR #162, opened 2023-03-18
fix: correct dlcol of diag and jump to it

Shane-XB-Qian:
1, dlcol of diag looks was wrong
2, and jump to the right col of diag was a matter of code action, looks would be more smooth at least.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -210,16 +210,10 @@ def ShowDiagInPopup(diag: dict<any>)
     # The line number is outside the last line in the file.
     dlnum = lastline
   endif
-  # Make sure column is not outside the character in line.
   if dlcol < 1
-    dlcol = 1
-  elseif dlcol > ltext->len()
-    dlcol = ltext->len()
+    # The column is outside the last character in line.
+    dlcol = ltext->len() + 1
   endif
-
-  # Move cursor to the right col of current diag
-  cursor(dlnum, dlcol)
-
   var d = screenpos(0, dlnum, dlcol)
   if d->empty()
     # If the diag position cannot be converted to Vim lnum/col, then use
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -310,6 +310,7 @@ export def LspDiagsJump(lspserver: dict<any>, which: string): void
     var lnum = diag.range.start.line + 1
     if (which == 'next' && lnum > curlnum)
 	  || (which == 'prev' && lnum < curlnum)
+	  || (which == 'this' && lnum == curlnum)
       setcursorcharpos(diag.range.start.line + 1, diag.range.start.character + 1)
       return
     endif
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -210,10 +210,16 @@ def ShowDiagInPopup(diag: dict<any>)
     # The line number is outside the last line in the file.
     dlnum = lastline
   endif
+  # Make sure column is not outside the character in line.
   if dlcol < 1
-    # The column is outside the last character in line.
-    dlcol = ltext->len() + 1
+    dlcol = 1
+  elseif dlcol > ltext->len()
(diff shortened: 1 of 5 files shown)


Shane-XB-Qian:
1, original code put dlcol the end of that line (obviously over the  length)
2, or i am not sure if dlcol was possible a negative value? then slice from end to start of that line?
3,  i changed to just in the line, i supposed this is correct. please check again, i maybe wrong.

autoload/lsp/diag.vim
```diff
@@ -210,10 +210,16 @@ def ShowDiagInPopup(diag: dict<any>)
...
   if dlcol < 1
-    # The column is outside the last character in line.
-    dlcol = ltext->len() + 1
+    dlcol = 1
+  elseif dlcol > ltext->len()
+    dlcol = ltext->len()
   endif
+
+  # Move cursor to the right col of current diag
+  cursor(dlnum, dlcol)
```
andlrc, line 221: I don't think this function should move the cursor, it would make sense to change `LspDiagsJump` instead to move the cursor to the correct column, something like https://github.com/yegappan/lsp/pull/165

autoload/lsp/diag.vim
```diff
@@ -210,10 +210,16 @@ def ShowDiagInPopup(diag: dict<any>)
     # The line number is outside the last line in the file.
     dlnum = lastline
   endif
+  # Make sure column is not outside the character in line.
   if dlcol < 1
-    # The column is outside the last character in line.
-    dlcol = ltext->len() + 1
+    dlcol = 1
+  elseif dlcol > ltext->len()
+    dlcol = ltext->len()
```
andlrc, line 217: `dlcol` is character count, while `len(` returns byte count, you might be able to use `strcharlen` instead.

autoload/lsp/diag.vim
andlrc: Or possible the cursor movement should be moved to the function `ShowCurrentDiag`.

Shane-XB-Qian:
this is to make `LspDiagCurrent` jump to correct col,
but do not want to change too much which make `LspDiagPrev`, `LspDiagNext`, `LspDiagFirst` behavior change.

autoload/lsp/diag.vim
andlrc: This isn't resolved.

andlrc:
> this is to make `LspDiagCurrent` jump to correct col,

If so then I think the logic should possible be moved to the `ShowCurrentDiag` function.

Shane-XB-Qian:
@yegappan could you merge this as well? i wish `LspDiagCurrent` can jump to correct col as well.

autoload/lsp/diag.vim
yegappan: Yes.  The cursor should not be moved.  The popup window with the diag message should be displayed in the correct column.

yegappan:
Do you have some sample code where the diag popup window is displayed in the wrong location?

autoload/lsp/diag.vim
Shane-XB-Qian: what do you mean? i cannot get your point.
this is to get the diag message to be displayed in the correct column, and more than that, to make the code action took action for the correct one! some lsp server may have code action advise even no diag, unless cursor on correct place.

Shane-XB-Qian:
it did not mean sample code had it wrong location, i am actually curious when it would be `< 1`, 
but just obviously should not always locate at head or end, vs/but correct in the right place.

Shane-XB-Qian:
or simply question: how can i h/l/w/e/b the cursor to the right diag place for current line?

Shane-XB-Qian:
supposed you are fixing the est err at `Test_LspSelection`, those cursor location something was irrelevant to this one.

Shane-XB-Qian:
i rollback the changes on `LspDiagCurrent`, but adding a new `LspDiagThis` to jump to diag msg in current line.
now, ok?

PR #163, opened 2023-03-19
Respect "isIncomplete" when filtering completion items

andlrc:
The specifications states:

> for speed clients should be able to filter an already received
> completion list if the user continues typing. Servers can opt out of
> this using a CompletionList and mark it as isIncomplete

Please see attached video: https://asciinema.org/a/4pGB1vszSpaXQp0SDKfGIhozI

Closes #160 , #164

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -125,9 +125,10 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
       # snippet completion.  Needs a snippet plugin to expand the snippet.
       # Remove all the snippet placeholders
       d.word = MakeValidWord(d.word)
-    else
+    elseif !lspserver.completeItemsIsIncomplete
       # plain text completion.  If the completion item text doesn't start with
       # the current (case ignored) keyword prefix, skip it.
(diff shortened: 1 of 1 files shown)


andlrc:
@Shane-XB-Qian Can you see if commit solves your problem with omni completion and the `gopls`?

Shane-XB-Qian:
i have replied you, please check https://github.com/yegappan/lsp/pull/160#issuecomment-1475269183
the original code had done the complete_check already, though i am not sure how reliable it is, but this `isincomplete` flag as per general sense probably more NOT reliable.

BTW: please try to close another pr by `closes xxx`, that should be work for `issue`.

andlrc:
> but this isincomplete flag as per general sense probably more NOT reliable.

Using `isIncomplete` is the way the specifications allow the language server to inform the client that no filtering should be done. See the specifications that I linked above.

Shane-XB-Qian:
many lsp server did not fully compliance with the specification.

andlrc:
> many lsp server did not fully compliance with the specification.

Which sends `isIncomplete` while being complete? I havn't found any?

PR #165, opened 2023-03-19
Jump to the proper column when jumping to diagnostics

andlrc:


andlrc:
Need tests

Shane-XB-Qian:
i have replied you, please check https://github.com/yegappan/lsp/pull/162#issuecomment-1475275529

PR #167, opened 2023-03-20
fix: omnifunc broken

Shane-XB-Qian:
pr #160 was missed closed, so reopen a new one for it.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -311,7 +311,9 @@ def g:LspOmniFunc(findstart: number, base: string): any
       return res->empty() ? v:none : res
     endif
 
-    return res->empty() ? v:none : res->filter((i, v) => v.word =~# '^' .. lspserver.omniCompleteKeyword)
+    var prefix: string = lspserver.omniCompleteKeyword->tolower()
+    # To filter (case ignored) keyword prefixed compl items only.
+    return res->empty() ? v:none : res->filter((i, v) => v.word->tolower()->stridx(prefix) == 0)
   endif
 enddef
 
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -135,6 +135,7 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     endif
 
     d.abbr = item.label
+    d.icase = 1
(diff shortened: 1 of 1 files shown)


autoload/lsp/completion.vim
```diff
@@ -311,7 +312,9 @@ def g:LspOmniFunc(findstart: number, base: string): any
       return res->empty() ? v:none : res
     endif
 
-    return res->empty() ? v:none : res->filter((i, v) => v.word =~# '^' .. lspserver.omniCompleteKeyword)
+    var prefix: string = lspserver.omniCompleteKeyword->tolower()
+    # To filter (case ignored) keyword prefixed compl items only.
+    return res->empty() ? v:none : res->filter((i, v) => v.word->tolower()->stridx(prefix) == 0)
```
andlrc, line 317: What are you trying to archive? Allow incase-sensitive completion? If so please  create an option for it. Possible one that can also take `fuzzy`.

Some language servers filters the completion items, see `clangd` for an example on this, other relays on the client doing the filtering, see `typescript-language-server` for such a language server.

This code will only have affect for the later case, keep that in mind.

If an option is added, then I suggest that the line 132 above:

```
if prefix != '' && d.word->tolower()->stridx(prefix) != 0
```

is changed to respect this option as well.

yegappan:
Yes. I was wondering why this PR was closed.  Can you reopen the PR? Also, do you have an example that shows the problem with not ignoring the case?  I am going to start adding tests for insert completion checking for the various cases.

Shane-XB-Qian:
anyway, those had been at my local already, wish you can merge, as for dup=0 or not, please help think... :+1:

andlrc:
> thx. one case is a simple js code:
> 
> ```js
> console.c|
> ```
> 
> cursor at `|`, if no icase=1, looks it would lost console.Console (`.C` is upper case), though i am not very sure if it was solid, but so far i can say that.

@Shane-XB-Qian What language server are you using? And how is it configured?

Shane-XB-Qian:
> What language server

same one like ts

andlrc:
> Yes. I was wondering why this PR was closed.

I think it's because Shane pushed the branch which, compared to main, had zero changes.

PR #175, opened 2023-03-22
fix: rpc and async_rpc had err msg trace already

Shane-XB-Qian:
// remove dup err msg in GotoSymbolLoc

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -885,18 +885,6 @@ def GotoSymbolLoc(lspserver: dict<any>, msg: string, peekSymbol: bool,
 		  cmdmods: string)
   var reply = lspserver.rpc(msg, GetLspTextDocPosition(true))
   if reply->empty() || reply.result->empty()
-    var emsg: string
-    if msg ==# 'textDocument/declaration'
-      emsg = 'Error: symbol declaration is not found'
-    elseif msg ==# 'textDocument/typeDefinition'
-      emsg = 'Error: symbol type definition is not found'
-    elseif msg ==# 'textDocument/implementation'
-      emsg = 'Error: symbol implementation is not found'
-    else
-      emsg = 'Error: symbol definition is not found'
-    endif
-
-    util.WarnMsg(emsg)
     return
   endif
 
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -630,6 +630,9 @@ def Rpc(lspserver: dict<any>, method: string, params: any): dict<any>
       emsg ..= $', data = {reply.error.data->string()}'
     endif
     util.ErrMsg($'Error(LSP): request {method} failed ({emsg})')
+  else
+    # No 'result'
+    HandleSymbolLocErr(lspserver, method)
   endif
 
   return {}
@@ -864,6 +867,21 @@ def ResolveCompletion(lspserver: dict<any>, item: dict<any>): void
 			completion.CompletionResolveReply)
 enddef
 
+# Handle symbol err msg when getting response
+def HandleSymbolLocErr(lspserver: dict<any>, msg: string)
+  var emsg: string
+  if msg ==# 'textDocument/declaration'
+    emsg = 'Error: symbol declaration is not found'
+  elseif msg ==# 'textDocument/typeDefinition'
+    emsg = 'Error: symbol type definition is not found'
+  elseif msg ==# 'textDocument/implementation'
+    emsg = 'Error: symbol implementation is not found'
+  else
+    emsg = 'Error: symbol definition is not found'
+  endif
+  util.WarnMsg(emsg)
+enddef
+
 # Jump to or peek a symbol location.
 #
 # Send 'msg' to a LSP server and process the reply.  'msg' is one of the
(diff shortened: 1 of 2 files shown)


Shane-XB-Qian:
test fail due to tsserver_test.vim missed a comma, see commit here.

autoload/lsp/lspserver.vim
```diff
@@ -885,18 +885,6 @@ def GotoSymbolLoc(lspserver: dict<any>, msg: string, peekSymbol: bool,
...
-      emsg = 'Error: symbol declaration is not found'
-    elseif msg ==# 'textDocument/typeDefinition'
-      emsg = 'Error: symbol type definition is not found'
-    elseif msg ==# 'textDocument/implementation'
-      emsg = 'Error: symbol implementation is not found'
-    else
-      emsg = 'Error: symbol definition is not found'
-    endif
-
-    util.WarnMsg(emsg)
```
yegappan, line 899: When trying to jump to an non-existing symbol declaration/definition, we do want to display a warning message.  If this is removed, then how will the user know that the symbol is not found?

Shane-XB-Qian: for example: `Error(LSP): request textDocument/implementation failed (no object found, code = 0)`

yegappan: I don't see this error message with clangd when trying to jump to a non-existing symbol definition.  Which language server are you using?

Shane-XB-Qian: gopls for golang. supposed this should be standard later or now?

Shane-XB-Qian: pls check, i push a new commit, now should be able to handle both clangd and gopls.

autoload/lsp/lspserver.vim
```diff
@@ -630,6 +630,9 @@ def Rpc(lspserver: dict<any>, method: string, params: any): dict<any>
       emsg ..= $', data = {reply.error.data->string()}'
     endif
     util.ErrMsg($'Error(LSP): request {method} failed ({emsg})')
+  else
+    # No 'result'
+    HandleSymbolLocErr(lspserver, method)
```
yegappan, line 635: The RPC request can be for a method other than the symbol declaration, definition, etc.  If such a request fails, then the symbol related error message should not be printed.

Shane-XB-Qian: so far HandleSymbolLocErr only handle symbolloc err, such as define/declare/impls etc.
others if lsp server followed the 'error' specification, then it would print it by 'error' part.

andlrc:
Another approach would be to tell the function that it shouldn't handle errors, and let the caller handle errors, something like:

```
index e750c02..a57703b 100644
@@ -598,7 +598,7 @@ enddef
 
 # Send a sync RPC request message to the LSP server and return the received
 # reply.  In case of an error, an empty Dict is returned.
-def Rpc(lspserver: dict<any>, method: string, params: any): dict<any>
+def Rpc(lspserver: dict<any>, method: string, params: any, handleError: bool = true): dict<any>
   var req = {}
   req.method = method
   req.params = {}
@@ -623,13 +623,15 @@ def Rpc(lspserver: dict<any>, method: string, params: any): dict<any>
   endif
 
   if reply->has_key('error')
-    # request failed
-    var emsg: string = reply.error.message
-    emsg ..= $', code = {reply.error.code}'
-    if reply.error->has_key('data')
-      emsg ..= $', data = {reply.error.data->string()}'
+    if handleError
+      # request failed
+      var emsg: string = reply.error.message
+      emsg ..= $', code = {reply.error.code}'
+      if reply.error->has_key('data')
+        emsg ..= $', data = {reply.error.data->string()}'
+      endif
+      util.ErrMsg($'Error(LSP): request {method} failed ({emsg})')
     endif
-    util.ErrMsg($'Error(LSP): request {method} failed ({emsg})')
   endif
 
   return {}
@@ -883,7 +885,7 @@ enddef
 # Result: Location | Location[] | LocationLink[] | null
 def GotoSymbolLoc(lspserver: dict<any>, msg: string, peekSymbol: bool,
 		  cmdmods: string)
-  var reply = lspserver.rpc(msg, GetLspTextDocPosition(true))
+  var reply = lspserver.rpc(msg, GetLspTextDocPosition(true), false)
   if reply->empty() || reply.result->empty()
     var emsg: string
     if msg ==# 'textDocument/declaration'
```

yegappan:
I like the approach of the Rpc() function accepting the "handleError" argument.  Can you update the PR with this change?

andlrc:
> I like the approach of the Rpc() function accepting the "handleError" argument. Can you update the PR with this change?

https://github.com/yegappan/lsp/pull/179

Shane-XB-Qian:
> I like the approach of the Rpc() function accepting the "handleError" argument

this is unnecessary, as per the specification, if it had `error` key in the resp, that meant it had error, just some lsp server did not follow the specification, then now we specifically adding the symbolloc err handler for those ones.

// only one thing looks Ugly which needing to handle it had no error AND result is empty at 2 Places, it becauses the original code designed if no Rpc result is empty then return `{}` which cannot get original `error` !

I Think we do not need to make code complex which adding such many or unnecessary parameter.

Shane-XB-Qian:
but whatever, this should be not a matter to me, as long as no dup err.

Issue #172, opened 2023-03-21
Completion errors in some cases


ycycyyc:
```
package main

import "fmt"

func main() {
    var err error
    if err = fmt.<cursor is here>  err != nil {
    }
}
```
Then type a letter 'e'
![image](https://user-images.githubusercontent.com/30541266/226620162-738defda-b93b-4069-929d-17a24e9708f9.png)

Type ctrl-n 
![image](https://user-images.githubusercontent.com/30541266/226620581-67dd3c82-7e84-4d87-bed1-014a542b9e3b.png)

What I expected is "Errorf", but got "errorf". There was an bug in completion.

https://github.com/yegappan/lsp/blob/main/autoload/lsp/completion.vim#L93
```
var starttext = getline('.')[ : chcol - 1]
```
I found the variable starttext is "    if err = fmt.e ”,  but correct value should be missing the final space.

After making the following changes, the variable starttext is "    if err = fmt.e" and the completion is correct finally.
```
diff --git a/autoload/lsp/completion.vim b/autoload/lsp/completion.vim
index fe9799d..65ad42b 100644
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -90,7 +90,7 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
 
   # Get the keyword prefix before the current cursor column.
   var chcol = charcol('.')
-  var starttext = getline('.')[ : chcol - 1]
+  var starttext = getline('.')[ : chcol - 2]
   var [prefix, start_idx, end_idx] = starttext->tolower()->matchstrpos('\k*$')
   var start_col = start_idx + 1
```
In general, when there is no content behind the cursor, the value  of starttext is correct, because the value of chcol is equal to the length of the current line chcol plus 1.

yegappan:
Can you create a PR with this change?

Shane-XB-Qian:
need to take care if chcol is 1, then chcol - 2 is a negative value.

PR #176, opened 2023-03-22
fix: starttext maybe incorrect

Shane-XB-Qian:
https://github.com/yegappan/lsp/issues/172#issuecomment-1478014976

// set a safeguard of chcol

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -90,7 +90,7 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
 
   # Get the keyword prefix before the current cursor column.
   var chcol = charcol('.')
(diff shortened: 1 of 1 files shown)


autoload/lsp/completion.vim
```diff
@@ -90,7 +90,7 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
 
   # Get the keyword prefix before the current cursor column.
   var chcol = charcol('.')
-  var starttext = getline('.')[ : chcol - 2]
+  var starttext = chcol == 1 ? '' : getline('.')[ : chcol - 2]
```
yegappan, line 93: I tried various completion methods and found that the column number is always greater than 1.
Do you have an example that shows the condition under which the column is 1?

PR #178, opened 2023-03-23
Add "completionMatcher" which can be set to "icase", "case" or "fuzzy"

andlrc:
This only works for language servers which always sends the whole set of completion items to the client, like `tsserver`.

Which means this option wouldn't work for `clangd`.

This also makes it possible to opt-out of the ignore cased completion, which have been implemented in #157 and #167 

See a video of how it works here:

https://asciinema.org/a/ko4hFgnzoMUqSVQlvxI1u8OlS

### DIFF
--- a/doc/lsp.txt
+++ b/doc/lsp.txt
@@ -282,6 +282,17 @@ diagLineHL		Highlight used for diagnostic line.
 echoSignature		In insert mode, echo the current symbol signature
 			instead of showing it in a popup. By default this is
 			set to false.
+completionMatcher	Enable fuzzy or incase sensitive completion for language
+			servers that replies with a full list of completion
+			items.  Some language servers does completion filtering
+			in the server, while other relies on the client to do
+			the filtering.
+
+			This option only works for language servers that expect
+			the client to filter the completion items.
+
+			Can be either `fuzzy`, `icase` or `case`, default is
+			`case`.
 hideDisabledCodeActions Hide all disabled code actions
 ignoreMissingServer	Do not print a missing language server executable.
 			By default this is set to false.
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -91,7 +91,11 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
   # Get the keyword prefix before the current cursor column.
   var chcol = charcol('.')
   var starttext = chcol == 1 ? '' : getline('.')[ : chcol - 2]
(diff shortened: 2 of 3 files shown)


doc/lsp.txt
```diff
@@ -282,6 +282,17 @@ diagLineHL		Highlight used for diagnostic line.
...
+			servers that replies with a full list of completion
+			items.  Some language servers does completion filtering
+			in the server, while other relies on the client to do
+			the filtering.
+
+			This option only works for language servers that expect
+			the client to filter the completin items.
+
+			Can be either `fuzzy`, `icase` or `case`, default is
+			`case`.
```
andlrc, line 295: Currently "case" is the default. Which isn't backward compatible as ignored case completion have been added a week ago, and "completed" in #167

Tachi107: Is `case` a good default? I'd find it a bit annoying with languages that use camelCase like Java

andlrc: I don't know.

I find it annoying that `Hello` and `hello` will both be suggested when I type: `He`.

Vim's built-in completion `:help ins-completion` is case sensitive, with the twist that can set `:h 'infercase`.

Tachi107: I'm quite new to Vim, so if Vim's general default is case sensitive then it makes to have that default here too

andlrc: Either is fine I think. My personal favorite is `case` as that's how I would configure it myself.

Shane-XB-Qian:
> I find it annoying that `Hello` and `hello` will both be suggested when I type: `He`.

that did not change something, when you chose 'fuzzy', it would still get both!
besides 'fuzzy' so far is not a standard option for lsp server, most ones not or not plan to support it.
and it maybe a bit slow in some cases.

so this `option` would be another awkward option/parameter user really do not know how to choose or make it right.

Shane-XB-Qian:
i just said `one day` in pr #160 , did not mean now.. -_-#
one day we may need to support `fuzzy`, but now this seems made user difficult to choose or make it fully right.

andlrc:
> > I find it annoying that `Hello` and `hello` will both be suggested when I type: `He`.
> 
> that did not change something, when you chose 'fuzzy', it would still get both!

Is there anything in this video you don't understand? https://asciinema.org/a/ko4hFgnzoMUqSVQlvxI1u8OlS

Shane-XB-Qian:
yegappan is the author of 'fuzzy' feature of official vim, i was doing some test for that, so i think i should be understand it a bit.
but anyway, i'm actually glad you impl this, but just for now would be a bit difficult to choose this option.

PR #183, opened 2023-03-24
fix: 'diaghere' is to keep jumpping in the line

Shane-XB-Qian:


autoload/lsp/diag.vim
```diff
@@ -362,7 +362,7 @@ export def LspDiagsJump(lspserver: dict<any>, which: string): void
     if (which == 'next' && (lnum > curlnum || lnum == curlnum && col > curcol))
 	  || (which == 'prev' && (lnum < curlnum || lnum == curlnum
 							&& col < curcol))
-	  || (which == 'here' && (lnum == curlnum && col >= curcol))
+	  || (which == 'here' && (lnum == curlnum && col > curcol))
```
yegappan, line 365: With this change, if there is a diagnostic at the beginning of a line and the cursor is at the first column, then the ":LspDiagHere" command will fail and there is no way to jump to this diagnostic.  If there are multiple diagnostics
on a single line, you can use the ":LspDiagNext" command after the ":LspDiagHere" command to jump to the
next diagnostic on the line.  With this change, this will command will behave like ":LspDiagAfter".

Shane-XB-Qian:
this was intended to work with/before diagCurrent, or now you combined with them, that's ok though a bit mix the usage/design of jump vs display.

but whatever if the diag msg really at the line head col 1 and only that col, just press 0 (this is designed or when to turn around the diagHere/diagHhis), the diag msg would display.

--
shane.xb.qian

Shane-XB-Qian:
ok, i felt it's not a matter, so let it be.
but the doc i commit in this pr you may like to rephrase it still. :-)

yegappan:
Can you update the PR with only the doc changes? I will merge the changes.

Shane-XB-Qian:
> Can you update the PR with only the doc changes?

#185

PR #184, opened 2023-03-25
fix: Apply proper mapping for <CR>

ilango100:
The "\<C-Y>\<CR>" seems to not properly insert ^Y

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -494,7 +494,7 @@ export def BufferInit(lspserver: dict<any>, bnr: number, ftype: string)
     endif
     # <Enter> in insert mode stops completion and inserts a <Enter>
     if !opt.lspOptions.noNewlineInCompletion
-      inoremap <expr> <buffer> <CR> pumvisible() ? "\<C-Y>\<CR>" : "\<CR>"
+      inoremap <expr> <buffer> <CR> pumvisible() ? "\<C-Y>" : "\<CR>"
     endif
   else
     if LspOmniComplEnabled(ftype)


autoload/lsp/completion.vim
```diff
@@ -494,7 +494,7 @@ export def BufferInit(lspserver: dict<any>, bnr: number, ftype: string)
     endif
     # <Enter> in insert mode stops completion and inserts a <Enter>
     if !opt.lspOptions.noNewlineInCompletion
-      inoremap <expr> <buffer> <CR> pumvisible() ? "\<C-Y>\<CR>" : "\<CR>"
+      inoremap <expr> <buffer> <CR> pumvisible() ? "\<C-Y>" : "\<CR>"
```
yegappan, line 497: You can set the "noNewlineInCompletion" option to TRUE to not add the newline in the map.  If you remove the newline from the map, then it will not be inserted when pressing C-Y.

ilango100: That is fine. I want <C-Y> to **be** mapped. But when we have this mapping, hitting <CR> doesn't properly insert the completion, atleast for snippet completions:
lsp_complete

Hitting <C-Y> on snippet completion properly inserts it, but <CR> doesn't insert it. But with the above fix, it is properly inserted.

yegappan:
This behavior is consistent with the built-in Vim behavior described under `:help popupmenu-keys`.  Pressing `Enter` in the first and third state of a completion popup menu, inserts the selected item and a line break.  In the second state, it only inserts the selected item.

ilango100:
Okay. I mistook the intended behaviour of this option to be to map `<CR>` to always insert the current match, but it is to always insert newline. Closing this PR as it works as intended.

PR #186, opened 2023-03-27
fix: error in diags display

ycycyyc:


ycycyyc:
image
When we use the lspDiagNext, LspDiagPrev and lspDiagFirst commands, but the diag is not in the current screen, the above error display will appear in this case

ycycyyc:
image

When we use the lspDiagNext, LspDiagPrev and lspDiagFirst commands, but the diag is not in the current screen, the above error display will appear in this case

PR #187, opened 2023-03-27
add diagnostic messages display with virtual text and pointer

greeschenko:
Hi, I like your project and use it in my main .vimrc 

I suggest adding optional diagnostic messages inline display

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -27,6 +27,11 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
   # Remove all the existing diagnostic signs
   sign_unplace('LSPDiag', {buffer: bnr})
 
+  if has('patch-9.0.1157') && opt.lspOptions.showDiagWithVirtualText
+      # Remove all virtual text
+      prop_remove({type: 'LspDiagVirtualText', bufnr: bnr, all: true})
+  endif
+
   if !lspserver.diagsMap->has_key(bnr) ||
       lspserver.diagsMap[bnr].sortedDiagnostics->empty()
     return
@@ -41,6 +46,15 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
     signs->add({id: 0, buffer: bnr, group: 'LSPDiag',
 				lnum: lnum,
 				name: DiagSevToSignName(diag.severity)})
+
+    if has('patch-9.0.1157') && opt.lspOptions.showDiagWithVirtualText
+        prop_add( lnum, 0, {
+            'bufnr': bnr,
+            'type': 'LspDiagVirtualText',
+            'text': '┌─ ' .. diag.message,
+            'text_align': 'above',
+            'text_padding_left': diag.range.start.character})
+    endif
   endfor
 
   signs->sign_placelist()
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -24,6 +24,7 @@ enddef
 
 # Refresh the signs placed in buffer 'bnr' on lines with a diagnostic message.
 def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
+  bnr->bufload()
   # Remove all the existing diagnostic signs
   sign_unplace('LSPDiag', {buffer: bnr})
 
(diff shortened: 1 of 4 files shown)


doc/lsp.txt
```diff
@@ -334,6 +334,9 @@ showInlayHints		|Boolean| option.  Show inlay hints from the language
 			server.  By default this is set to false.  The inlay
 			hint text is displayed as a virtual text.  Needs Vim
 			version 9.0.0178 or later.
+showDiagWithVirtualText		|Boolean| option.  Show diagnostic messages text from the language
+			server with virtual text.  By default this is set to false.
+            Needs Vim version 9.0.1157 or later.
 showSignature		|Boolean| option.  In insert mode, automatically show
```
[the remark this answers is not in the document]
greeschenko: yes of course i will add it

andlrc:
For me it seems like the diagnostic popup (The popup drawn by `lsp#diag#LspDiagsJump -> DisplayDiag` fails when applying this PR:

BROKEN: https://asciinema.org/a/f41fSMzEZVtR9Je2pBR5WvPih

autoload/lsp/diag.vim
```diff
@@ -41,6 +46,15 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
...
+
+    if has('patch-9.0.1157') && opt.lspOptions.showDiagWithVirtualText
+        prop_add( lnum, 0, {
+            'bufnr': bnr,
+            'type': 'LspDiagVirtualText',
+            'text': '┌─ ' .. diag.message,
+            'text_align': 'above',
+            'text_padding_left': diag.range.start.character})
+    endif
   endfor
```
yegappan, line 58: Can you modify the DiagsHighlightDisable() function to remove the virtual text?

autoload/lsp/diag.vim
```diff
@@ -41,6 +46,15 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
     signs->add({id: 0, buffer: bnr, group: 'LSPDiag',
 				lnum: lnum,
 				name: DiagSevToSignName(diag.severity)})
+
+    if has('patch-9.0.1157') && opt.lspOptions.showDiagWithVirtualText
+        prop_add( lnum, 0, {
+            'bufnr': bnr,
```
yegappan, line 52: Can you remove the single quotes from the key names?  It is not needed in Vim9 script.

yegappan (changes requested):

autoload/lsp/diag.vim
```diff
@@ -27,6 +27,11 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
   # Remove all the existing diagnostic signs
```
yegappan, line 27: I noticed that when loading multiple buffers, there are error messages about trying to add a text property to an unloaded buffer.  At the beginning of this function, you need to add

   bnr->bufload()

to fix these issues.

autoload/lsp/diag.vim
andlrc: > > I don't understand why though, as these shouldn't affect each other?
> 
> Did you try with the latest plugin? A redraw call was added before displaying the popup.

I tried it out on d161db7, see https://asciinema.org/a/0VAnghdBa73OatXW2lCC7vnDV

greeschenko: fix added in https://github.com/yegappan/lsp/pull/187/commits/6a7464d2dfa68fa5e4525e876712bcbed5f0300b

autoload/lsp/diag.vim
greeschenko: added in https://github.com/yegappan/lsp/pull/187/commits/6a7464d2dfa68fa5e4525e876712bcbed5f0300b

autoload/lsp/diag.vim
greeschenko: in https://github.com/yegappan/lsp/pull/187/commits/fdcb549003e49a48f5846911f25afaa4b2bef42d connect virtual text display to the autoHighlightDiags option as result we have on / off all Diags highlight and virtual text with :LspDiagHighlightEnable() :LspDiagHighlightDisable()

autoload/lsp/diag.vim
greeschenko: > > > I don't understand why though, as these shouldn't affect each other?
> > 
> > 
> > Did you try with the latest plugin? A redraw call was added before displaying the popup.
> 
> I tried it out on [d161db7](https://github.com/yegappan/lsp/commit/d161db7c5d8ed8e2a4d73fa31e67dc54221f4eaa), same problem, see https://asciinema.org/a/0VAnghdBa73OatXW2lCC7vnDV

I think there is an error in the virtual text function at the editor level.
 I noticed that the virtual text doesn't always work correctly with pop-up dialogs.

PR #190, opened 2023-03-29
add diagnostics signs options

greeschenko:


autoload/lsp/options.vim
```diff
@@ -53,6 +53,25 @@ export var lspOptions: dict<any> = {
...
+      info: {
+          text: 'I>',
+          texthl: 'Pmenu',
+      },
+      hint: {
+          text: 'H>',
+          texthl: 'Question',
+      },
+  },
 }
```
andlrc, line 75: It's worth to keep in mind that using nested objects as options, will force the user to set all properties as once, as otherwise some properties will be removed, eg:

    g:LspOptionsSet({ diagSignOpts: { error: { text: 'E~', texthl: 'ErrorMsg' } } }) 

will result in:

 ```
  diagSignOpts: {
      error: {
          text: 'E~',
          texthl: 'ErrorMsg',
      },
  },
  ```

It might be worth mentioning it in the help text, or change the code to have many options:

    diagSignErrorText: 'E>',
    diagSignErrorTexthl: 'ErrorMsg',
    ...

greeschenko: Thank you for comments I change options to separate

doc/lsp.txt
```diff
@@ -351,6 +351,23 @@ usePopupInCodeAction    |Boolean| option.  When using the |:LspCodeAction|
 			command to display the code action for the current
 			line, use a popup menu instead of echoing.
 			By default this is set to false.
+diagSignErrorText       |String| option. Change diag sign text for errors
```
yegappan, line 354: The list of options in the help is alphabetically sorted.  Can you move the below text to the right place above?

greeschenko: fix added in https://github.com/yegappan/lsp/pull/190/commits/680004fed5a5eca138662e3d7ba5b4a514a18ad3

doc/lsp.txt
```diff
@@ -306,6 +300,30 @@ completionMatcher	|String| option.  Enable fuzzy or case insensitive
 			    case  - case sensitive matching (default).
 			    fuzzy - fuzzy match completion items.
 			    icase - ignore case when matching items.
+diagLineHL		|String| option. The highlight group used for a line
+			with one or more diagnostics.  By default uses
+			"DiffAdd".  Use "NONE" to disable.
+diagVirtualTextHL	|String| option.  The highlight group used for a
```
yegappan, line 306: Can you move the text related to "diagVirtualTextHL" to below the text for "diagSignHintTexthl"?

PR #191, opened 2023-03-29
Feat/inline diagnostic highlight

andlrc:
This might give inline highlighting:

See https://asciinema.org/a/DIRI6f3MDUewIWSgZc9IoPT75

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -19,6 +19,25 @@ export def InitOnce()
 		{name: 'LspDiagHint', text: 'H>', texthl: 'Question',
 						linehl: lineHL}])
 
+  if opt.lspOptions.highlightDiagInline
+    if !hlexists('LspDiagInlineError')
+      hlset([{name: 'LspDiagInlineError', linksto: opt.lspOptions.diagInlineErrorHL}])
+    endif
+    if !hlexists('LspDiagInlineWarning')
+      hlset([{name: 'LspDiagInlineWarning', linksto: opt.lspOptions.diagInlineWarningHL}])
+    endif
+    if !hlexists('LspDiagInlineInfo')
+      hlset([{name: 'LspDiagInlineInfo', linksto: opt.lspOptions.diagInlineInfoHL}])
+    endif
+    if !hlexists('LspDiagInlineHint')
+      hlset([{name: 'LspDiagInlineHint', linksto: opt.lspOptions.diagInlineHintHL}])
+    endif
+    prop_type_add('LspDiagInlineError', { highlight: 'LspDiagInlineError', override: true })
+    prop_type_add('LspDiagInlineWarning', { highlight: 'LspDiagInlineWarning', override: true })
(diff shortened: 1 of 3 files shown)


autoload/lsp/options.vim
```diff
@@ -40,6 +40,16 @@ export var lspOptions: dict<any> = {
...
+  # Highlight diagnostics inline
+  highlightDiagInline: false,
+  # The highlight group used inline Error diagnostics
+  diagInlineErrorHL: 'SpellBad',
+  # The highlight group used inline Warning diagnostics
+  diagInlineWarningHL: 'SpellCap',
+  # The highlight group used inline Info diagnostics
+  diagInlineInfoHL: 'SpellRare',
+  # The highlight group used inline Hint diagnostics
+  diagInlineHintHL: 'SpellLocal',
```
andlrc, line 52: These defaults might be a bad choice?

andlrc:
It might also be worth spending a bit of time on adding tests.

Tachi107:
> It might also be worth spending a bit of time on adding tests, but that would require a different vim version in the pipeline thinking

With GitHub Actions you can easily run tests in containers, so for example you could run these specific tests in a `debian:testing` or `debian:bookworm` container that has Vim 9.0.1000.

See [this file](https://github.com/Tachi107/cloudflare-ddns/blob/ca703350effc86dc141c294126383ad9ec48c0ab/.github/workflows/muon.yaml#L27-L29) for an example

andlrc:
> > It might also be worth spending a bit of time on adding tests, but that would require a different vim version in the pipeline thinking
> 
> With GitHub Actions you can easily run tests in containers, so for example you could run these specific tests in a `debian:testing` or `debian:bookworm` container that has Vim 9.0.1000.
> 
> See [this file](https://github.com/Tachi107/cloudflare-ddns/blob/ca703350effc86dc141c294126383ad9ec48c0ab/.github/workflows/muon.yaml#L27-L29) for an example

Come to think of it I think the limited use of `prop_add` that is used in this MR was introduced in late vim 8.

autoload/lsp/options.vim
yegappan: I will merge this PR.  We can fine tune the color based on any feedback.

yegappan:
>
> It might also be worth spending a bit of time on adding tests, but that would require a different vim version in the pipeline
>

I am going to merge this PR.  But can you add some tests for this in a separate PR?

autoload/lsp/diag.vim
```diff
@@ -71,6 +111,15 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
 				lnum: lnum,
 				name: DiagSevToSignName(diag.severity)})
 
+    if opt.lspOptions.highlightDiagInline
+      prop_add(diag.range.start.line + 1,
```
yegappan, line 115: When adding a few lines with diagnostics messages and then removing them (by doing an undo) , I see an "E966: Invalid line number" error.  It looks like by the time the diagnostics messages are processed, the lines are already removed.  I think we need to debounce the diagnostic messages.

PR #193, opened 2023-03-30
Ignore invalid lnum and column for prop_add

andlrc:
Diagnostics arrive asynchronous and the document might have changed before they reach the client.

This could cause the line number or column to be invalid when the diagnostics arrive.

Ignoring these errors isn't a big deal as new diagnostics will be send by the server shortly. And when there isn't any document changes, the client will be updated ASAP.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -111,22 +111,27 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
 				lnum: lnum,
 				name: DiagSevToSignName(diag.severity)})
 
-    if opt.lspOptions.highlightDiagInline
-      prop_add(diag.range.start.line + 1,
-		util.GetLineByteFromPos(bnr, diag.range.start) + 1,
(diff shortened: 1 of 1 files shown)


andlrc:
I havn't been able to reproduce the problem myself by it's mentioned here: https://github.com/yegappan/lsp/pull/191#discussion_r1153347862

yegappan:
Can you try the below script?

```
vim9script

new xx.c
var lines =<< trim END
  void func(void)
  {

      int a, b, c;
  }
END
setline(1, lines)
&undolevels = &undolevels
lines =<< trim END
  {
     a = thisVar;
     b = thatVar;
     c = otherVar;
  }
END
for i in range(4)
  append(4 + i * 5, lines)
  &undolevels = &undolevels
endfor

for i in range(5)
    undo
endfor
```

andlrc:
> Can you try the below script?
>
> [...]

Thanks for providing a script that can reproduce the error. It seems like the try-catch swallows the error:

https://asciinema.org/a/IP5qGrJhbYTqQvju58kQz3utf

PR #196, opened 2023-03-31
end_col of diag inline prop maybe overflow

Shane-XB-Qian:


autoload/lsp/diag.vim
```diff
@@ -134,7 +134,7 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
         prop_add(diag.range.start.line + 1,
                   util.GetLineByteFromPos(bnr, diag.range.start) + 1,
                   {end_lnum: diag.range.end.line + 1,
-                    end_col: util.GetLineByteFromPos(bnr, diag.range.end) + 1,
+                    end_col: util.GetLineByteFromPos(bnr, diag.range.end),
```
andlrc, line 137: This change breaks the highlight:

Before:

Skærmbillede 2023-03-31 kl  10 55 49


After:

Skærmbillede 2023-03-31 kl  10 55 28

autoload/lsp/diag.vim
```diff
@@ -146,7 +146,7 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
                            text_align: 'above',
                            text_padding_left: diag.range.start.character})
       endif
-    catch /E966\|E964/ # Invalid lnum | Invalid col
+    catch /E966\|E964\|E475/ # Invalid lnum | Invalid col | Invalid arg
```
andlrc, line 149: Can you tell me in which case is "E475" raised?

Shane-XB-Qian: when option value is invalid, e.g end_col is 0

autoload/lsp/diag.vim
Shane-XB-Qian: yea, had side effect if so.
i changed the way, now set a safeguard instead.

Shane-XB-Qian:
i cannot reproduce the case, but the prop hl was extended over the line end.
maybe the last vim patch helped this?
but whatever, now i changed the way which set a safeguard instead.

autoload/lsp/diag.vim
andlrc: The reason I ask is because I think the only scenario that E475 can be raised is iff the language server itself yields invalid ranges, which I think should be fixed in the language servers code, or at least be very explicitly handled in the client, with a reference to an issue on the specific language servers issue tracker.

autoload/lsp/diag.vim
```diff
@@ -131,10 +131,17 @@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
 
     try
       if opt.lspOptions.highlightDiagInline
+	var end_col: number = util.GetLineByteFromPos(bnr, diag.range.end) + 1
+	var end_lnum: number = diag.range.end.line + 1
+	var end_text: string = end_lnum->getline()
+	var end_len: number = end_text->len()
+	if end_col > end_len + 1
+	  end_col = end_len + 1
+	endif
```
andlrc, line 140: I think that introducing this code would mean we could remove the catch `E964`, meaning that either we adjust the range (as you do here), well knowing the diagnostics are expired, as new will arrive shortly, or we simply ignore it with the catch.

I still lean towards using try-catch, as in either case the diagnostics that are being rendered are invalid/expired.

autoload/lsp/diag.vim
Shane-XB-Qian: i think no, since actually not sure when or which one would be like that, but if it was invalid then prop hl would be disappered (removed) unexpectedly untill next upd, so had to try-catch it, unless we knew all invalid cases.

autoload/lsp/diag.vim
Shane-XB-Qian: no, i think for now we needed both, that safeguard code is for one of valid but incorrect cases, but did not know or not sure all invalid cases, so try-catch still required for now.

andlrc: My point is that your new code in only needed for diagnostics that are older than the current document, and new diagnostics will arrive shortly. So I think it is a bad idea to complicate the code for something that can be handled by a `try-catch`.

Since the client speaks utf-32 positions things like:

    int /*😭😭😭😭😭😭😭😭😭😭*/a=b;

Will at worst case result in one of the `😭` being highlighted instead of `b`.

autoload/lsp/diag.vim
andlrc: > so had to try-catch it, unless we knew all invalid cases.

Can you provide code that can cause the "E475" error to happen? Because if not, then I think we shouldn't add more complexity than needed.

Shane-XB-Qian: um, seems i had offered an example? e.g end_col is 0

autoload/lsp/diag.vim
Shane-XB-Qian: you may do not understand the different between 'invalid' with 'incorrect'?
some cases it is incorrect but was valid which would Not try-catch-ed by try-catch.

autoload/lsp/diag.vim
andlrc: > um, seems i had offered an example? e.g end_col is 0

When will that happen? Can you provide code that would result in a language server sending `diag.range.end == -1`? Remember that we add `+1`

Shane-XB-Qian: you knew that's just an example, right? there are not only one option end_col, and one condition, one lsp server, right?

andlrc: > The reason I ask is because I think the only scenario that E475 can be raised is iff the language server itself yields invalid ranges, which I think should be fixed in the language servers code, or at least be very explicitly handled in the client, with a reference to an issue on the specific language servers issue tracker.

Shane-XB-Qian: that's same for E966 and E964, why you added and catch that?
i think we sure would remove these ugly code later, but kept it working for now is required also, right?

andlrc: > that's same for E966 and E964, why you added and catch that?

I believe that there is a comment about those here https://github.com/yegappan/lsp/blob/8af226327289ea84c07415c2665875bde6a44931/autoload/lsp/diag.vim#L147

Shane-XB-Qian:
@yegappan no plan to merge this?

yegappan:
>
> @yegappan no plan to merge this?
>

It is not clear to me why we need the additional check to validate the column number.  Note that computing the line length for every diagnostics will take time.  So if a check is not necessary, then we should not include that.  If you have some sample code or the sequence of steps to reproduce the problem, then we can include the check.

Shane-XB-Qian:
> Note that computing the line length for every diagnostics will take time.

was not that just same like that one you just merged for `DiagWithVirtualText`?
// those are all just a safeguard in case overflow or something or maybe unexpected vim bug...

// or you preferred to listen someone else's advise? :rofl:

Shane-XB-Qian:
let it be.. closing... :smile:

PR #198, opened 2023-04-01
Feat/add option to show references in qf

andlrc:
Add the option "useQuickfixForLocations" to use the quickfix list for locations comming from ":LspShowReferences", ":LspGotoDefinition", etc.

### DIFF
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -72,6 +72,8 @@ export var lspOptions: dict<any> = {
   diagSignInfoTexthl: 'Pmenu',
   diagSignHintText: 'H>',
   diagSignHintTexthl: 'Question',
+  # ShowReferences in a quickfix list instead of a location list`
+  useQuickfixForLocations: false
 }
 
 # set the LSP plugin options from the user provided option values
--- a/doc/lsp.txt
+++ b/doc/lsp.txt
@@ -326,6 +326,9 @@ echoSignature		|Boolean| option.  In insert mode, echo the current
 			By default this is set to false.
 hideDisabledCodeActions |Boolean| option. Hide all the disabled code actions.
 			By default this is set to false.
+hoverInPreview		|Boolean| option. Show |:LspHover| in a preview window
+			instead of a popup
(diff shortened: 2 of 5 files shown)


doc/lsp.txt
```diff
@@ -272,8 +272,29 @@ server and to set the language server options.  For example: >
...
+<
+The LspOptionsGet() function returns a |Dict| of all the LSP plugin options,
+To get a particular option value you can use the following: >
+
+	echo LspOptionsGet()['autoHighlightDiags']
+<
+
+All options are listed below:
+
+							*lsp-options-func*
```
andlrc, line 297: This split of options might not make sense? I think the list of options is growing quite large though.

autoload/lsp/options.vim
```diff
@@ -72,6 +72,8 @@ export var lspOptions: dict<any> = {
   diagSignInfoTexthl: 'Pmenu',
   diagSignHintText: 'H>',
   diagSignHintTexthl: 'Question',
+  # ShowReferences in a quickfix list instead of a location list`
+  referencesInQf: false
```
andlrc, line 76: This option might be named poorly, as it is also used when showing multiple implemention/definitions/etc in a list.

yegappan: What about using the name "useQuickfix"?

doc/lsp.txt
yegappan: I am thinking of removing the highlight group related options.  The user can override the pre-defined highlight groups by defining them before sourcing the plugin.

andlrc: Yes, that makes perfect sense.

andlrc: I'll remove the commit that reorders the docs then.

autoload/lsp/options.vim
andlrc: > What about using the name "useQuickfix"?

We have the option "autoPopulateDiags" which populates the location list, and I want to keep it like that, as then with this new option it's possible to use both.

What about "useQuickfixForLocations"

PR #204, opened 2023-04-02
Chore/colors in readme

andlrc:
Add syntax highlight in README.md and start the code examples at column one.

Before|After
-----|-----------
Before|After

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,66 +49,66 @@ To use the plugin features with a particular file type(s), you need to first reg
 
 To register a LSP server, add the following lines to your .vimrc file (use only the LSP servers that you need from the below list).  If you used [vim-plug](https://github.com/junegunn/vim-plug) to install the LSP plugin, the steps are described later in this section.
 ```viml
-   function! TypeScriptCustomNotificationHandler(lspserver, reply) abort
-     echom printf("TypeScript Version = %s", a:reply.params.version)
-   endfunction
-   let lspServers = [
-		\     #{
-		\        filetype: ['c', 'cpp'],
-		\        path: '/usr/local/bin/clangd',
-		\        args: ['--background-index']
-		\      },
-		\     #{
-		\	 filetype: ['javascript', 'typescript'],
-		\	 path: '/usr/local/bin/typescript-language-server',
-		\	 args: ['--stdio'],
-		\	 customNotificationHandlers: {
-		\	   '$/typescriptVersion': function('TypeScriptCustomNotificationHandler')
-		\	 }
-		\      },
-		\     #{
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,17 +6,17 @@ Language Server Protocol (LSP) plugin for Vim. You need Vim version 9.0 or above
...
 You can install this plugin directly from github using the following steps:
 
-```
-    $ mkdir -p $HOME/.vim/pack/downloads/opt
-    $ cd $HOME/.vim/pack/downloads/opt
-    $ git clone https://github.com/yegappan/lsp
+```bash
+$ mkdir -p $HOME/.vim/pack/downloads/opt
+$ cd $HOME/.vim/pack/downloads/opt
+$ git clone https://github.com/yegappan/lsp
```
Tachi107, line 12: Please be careful when using `bash` (or `sh` or `shell`) as the language for terminal prompts, as it'll mess syntax highlighting up when also showing the command output.

For example, consider these two blocks, where the first has `bash` as the language and the second only `plaintext`:

```bash
$ git --help | head
usage: git [-v | --version] [-h | --help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>] [--config-env=<name>=<envvar>]
           <command> [<args>]

These are common Git commands used in various situations:

start a working area (see also: git help tutorial)
```

```plaintext
$ git --help | head
usage: git [-v | --version] [-h | --help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>] [--config-env=<name>=<envvar>]
           <command> [<args>]

These are common Git commands used in various situations:

start a working area (see also: git help tutorial)
```

In the first example, a lot of non-alphanumeric characters are highlighted, together with words which have special meaning in bash shell scripts like `in` or `help`, while in reality what we really wanted to show was a plain text output of a command.

In short, I'd generally recommend to only use `bash` or `sh` for actual shell scripts, while using `plaintext` for simple terminal command blocks.

Not that it matters much in this case, but it's still useful to know I guess :)

andlrc: Yes I figured that it would be OK for this case. TBH it was the indent that annoyed me the most :smile:

PR #205, opened 2023-04-02
Add `:LspDiagCurrent!` to _only_ show diags if they are under the cursor

andlrc:
Make it possible to show diagnostics directly under the cursor, and not for the current line.

See the difference in this video:

https://asciinema.org/a/TVLSmVSoOjGxUzI5kRPTaMNp6

### DIFF
--- a/plugin/lsp.vim
+++ b/plugin/lsp.vim
@@ -78,7 +78,7 @@ augroup END
 
 # LSP commands
 command! -nargs=? -bar -range LspCodeAction lsp.CodeAction(<line1>, <line2>, <q-args>)
-command! -nargs=0 -bar LspDiagCurrent lsp.LspShowCurrentDiag()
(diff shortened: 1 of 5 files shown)


plugin/lsp.vim
```diff
@@ -78,7 +78,7 @@ augroup END
 
 # LSP commands
 command! -nargs=? -bar -range LspCodeAction lsp.CodeAction(<line1>, <line2>, <q-args>)
-command! -nargs=0 -bar LspDiagCurrent lsp.LspShowCurrentDiag()
+command! -nargs=0 -bang -bar LspDiagCurrent lsp.LspShowCurrentDiag(<bang>false)
```
andlrc, line 81: I think that it would be more intuitive if these wore changed, so that `:LspDiagCurrent` would only show diags directly under the cursor, while `:LspDiagCurrent!` should scan forward for one.  But it's also worth being backward compatible.

Tachi107: I think it'd be more intuitive too ;)

PR #209, opened 2023-04-03
Show error message instead of an empty popup when there are not content for ":LspHover"

andlrc:
Instead of showing an empty popup an error is shown when there is no hover:

I'm aware that some language servers already replies in a way that would result in no poup, but not the typescript-language-server.

This PR attempts to normalize the behavior across all language servers, by moving the hoverText calculation to its own function and draw the error if that function didn't return any hoverText.

Before | After
-------|-------
Before|After

See video for example:

https://asciinema.org/a/S7urffOWgXtq8bwMXuJOPwz2W

### DIFF
--- a/autoload/lsp/hover.vim
+++ b/autoload/lsp/hover.vim
@@ -22,6 +22,10 @@ def GetHoverText(hoverResult: any): list<any>
       return [hoverResult.contents.value->split("\n"), 'lspgfm']
     endif
 
+    util.TraceLog(
+      true,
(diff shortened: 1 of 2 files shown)


andlrc:
This makes something like this work for all language servers as well:

def Hover(): string
  g:LSPTest = true
  :LspHover
  g:LSPTest = false
  if popup_list()->len() == 0
    return 'K'
  endif
  return ''
enddef

nnoremap <buffer> <expr> K <Sid>Hover()

PR #216, opened 2023-04-05
Render visual selection over inline diag

andlrc:
Before | After
-------|------
Before | After


Shane-XB-Qian:
i do not agree, but that's not on me.
after so, that looked "no" syntax highlight at all.

andlrc:
>  after so, that looked "no" syntax highlight at all.

There should still be syntax highlighting, the only change is that the visual selection is above the diagnostic instead of being below:

syntax-prop-visual

Shane-XB-Qian:
where is your cursorline? have you thought user set cursorline?

Issue #215, opened 2023-04-05
error when select ultisnip element in completion list with tsserver


greeschenko:
How I understand tsserver tries to handle completion elements and find documentation, description or simply this element in the server completion list.

How can I prevent this check for added completion elements from Ultisnips?

With gopls server, I don't see this problem.

```

Error(LSP): request completionItem/resolve failed (Request completionItem/resolve failed with message: Error processing request. Cannot read properties of unde
fined (reading 'indexOf')^@TypeError: Cannot read properties of undefined (reading 'indexOf')^@    at normalizeSlashes (/usr/lib64/node_modules/typescript/lib/
tsserver.js:6014:26)^@    at Object.normalizePath (/usr/lib64/node_modules/typescript/lib/tsserver.js:6128:16)^@    at Object.toNormalizedPath (/usr/lib64/node
_modules/typescript/lib/tsserver.js:159741:23)^@    at Session.getFileAndProjectWorker (/usr/lib64/node_modules/typescript/lib/tsserver.js:168067:35)^@    at S
ession.getFileAndProject (/usr/lib64/node_modules/typescript/lib/tsserver.js:168051:29)^@    at Session.getCompletionEntryDetails (/usr/lib64/node_modules/type
script/lib/tsserver.js:168255:31)^@    at Session.handlers.ts.Map.ts.getEntries._a.<computed> (/usr/lib64/node_modules/typescript/lib/tsserver.js:167087:61)^@
   at /usr/lib64/node_modules/typescript/lib/tsserver.js:168880:88^@    at Session.executeWithRequestId (/usr/lib64/node_modules/typescript/lib/tsserver.js:168
871:28)^@    at Session.executeCommand (/usr/lib64/node_modules/typescript/lib/tsserver.js:168880:33)^@    at Session.onMessage (/usr/lib64/node_modules/typesc
ript/lib/tsserver.js:168906:35)^@    at Interface.<anonymous> (/usr/lib64/node_modules/typescript/lib/tsserver.js:171511:31)^@    at Interface.emit (node:event
s:513:28)^@    at [_onLine] [as _onLine] (node:internal/readline/interface:422:12)^@    at [_normalWrite] [as _normalWrite] (node:internal/readline/interface:6
15:22)^@    at Socket.ondata (node:internal/readline/interface:249:23)^@    at Socket.emit (node:events:513:28)^@    at addChunk (node:internal/streams/readabl
[...]

PR #217, opened 2023-04-05
fix: add filter to ultisnip list

greeschenko:
before this update 

if we use completion not in the start of line or edit code completion function get all list of snippets for this buffer 

now we filter snippets with matchfuzzy()

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -92,9 +92,22 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     lspserver.completeItemsIsIncomplete = cItems.isIncomplete
   endif
 
+  # Get the keyword prefix before the current cursor column.
+  var chcol = charcol('.')
+  var starttext = chcol == 1 ? '' : getline('.')[ : chcol - 2]
+  var [prefix, start_idx, end_idx] = starttext->matchstrpos('\k*$')
+  if opt.lspOptions.completionMatcher == 'icase'
+    prefix = prefix->tolower()
+  endif
+
+  var start_col = start_idx + 1
+
   if opt.lspOptions.ultisnipsSupport
       call UltiSnips#SnippetsInCurrentScope(1)
       for [key, info] in items(g:current_ulti_dict_info)
+          if matchfuzzy([key], prefix)->empty()
+            continue
+          endif
           var parts = split(info.location, ':')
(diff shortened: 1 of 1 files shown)


autoload/lsp/completion.vim
```diff
@@ -92,9 +92,22 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
...
+  if opt.lspOptions.completionMatcher == 'icase'
+    prefix = prefix->tolower()
+  endif
+
+  var start_col = start_idx + 1
+
   if opt.lspOptions.ultisnipsSupport
       call UltiSnips#SnippetsInCurrentScope(1)
       for [key, info] in items(g:current_ulti_dict_info)
+          if matchfuzzy([key], prefix)->empty()
```
andlrc, line 108: If you extract all the keys at once, then you can get `matchfuzzy` to sort the results based on how well they match, I'm not sure if that is needed though?

    for key in matchfuzzy(g:current_ulti_dict_info->keys(), prefix)
      var item = g:current_ulti_dict_info[key]

greeschenko: thank you for recommendation add it in https://github.com/yegappan/lsp/pull/217/commits/addf39f49791ac446b4a06aabf83656c5f34b1ad

autoload/lsp/completion.vim
```diff
@@ -92,20 +92,30 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
...
+  var starttext = chcol == 1 ? '' : getline('.')[ : chcol - 2]
+  var [prefix, start_idx, end_idx] = starttext->matchstrpos('\k*$')
+  if opt.lspOptions.completionMatcher == 'icase'
+    prefix = prefix->tolower()
+  endif
+
+  var start_col = start_idx + 1
+
   if opt.lspOptions.ultisnipsSupport
       call UltiSnips#SnippetsInCurrentScope(1)
```
yegappan, line 106: I would like to eventually move the Ultisnips specific code to a separate file.  We can may be use a snippets subdirectory and add snippet plugin specific files there.

greeschenko: Ok, in next pull request I try to relocate it to separate file. But now I want find a solution for this https://github.com/yegappan/lsp/issues/215 and I don`t now how long it will take

Issue #220, opened 2023-04-07
Padding of virtual text for diagnostics is wrong if tabs are used within in the text


berggeist:
There seems to be a bug within the padding of virtual text for diagnostic messages if tab characters are used in the source code.
In the following screenshot, the left corner of the diagnostic message should be aligned to column 9, right above the first character **f** within _fmt.Printf_. Instead, it is aligned incorrectly to column 2.

![vim-lsp-virtual-text-tabs-wrong](https://user-images.githubusercontent.com/24396556/230606842-56e6c514-4618-4497-b121-7dd8b202c4ad.png)

The correct behavior should be like in the following screenshot:
 
![vim-lsp-virtualtext-tab-correct](https://user-images.githubusercontent.com/24396556/230607119-491c1b51-c267-4425-85a4-01edbb0555b3.png)

The problem seems to be that for the padding the attribute **diag.range.start.character** is used and this attribute holds the value 1, i.e. for one tab character.
In the corrected screenshot the real display width is being used for calculating the padding with **strdisplaywidth()**, by applying the following patch:

``` vim
diff --git a/autoload/lsp/diag.vim b/autoload/lsp/diag.vim
index c0dffc4..805e7c8 100644
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim@@ -137,11 +137,16
@@ def DiagsRefreshSigns(lspserver: dict<any>, bnr: number)
       endif

       if opt.lspOptions.showDiagWithVirtualText
+        var padding = diag.range.start.character
+        if padding > 0
+          padding = strdisplaywidth(getline(diag.range.start.line + 1)[ : diag.range.start.character - 1])
+        endif
         prop_add(lnum, 0, {bufnr: bnr,
                            type: 'LspDiagVirtualText',
                            text: $'┌─ {diag.message}',
                            text_align: 'above',
-                           text_padding_left: diag.range.start.character})
+                           text_padding_left: padding})
       endif
```

Especially when programming in the go programming language this is an issue, because the go-lsp-formatter insists on using tab characters for indentation.

yegappan:
Thanks for reporting the issue with a patch.  Can you create a PR with this change?

berggeist:
You're welcome. I would like to thank you for this great vim plugin.
Never made a PR up to now, but I'll try my best. Just give me a little time :-)

berggeist:
Just verified that with the merge of PR #222  this lsp issue is solved.

PR #222, opened 2023-04-07
Fixes the left padding of LspDiagVirtualText messages with tab chars.

berggeist:


berggeist:
This fixes issue #220

yegappan:
I am able to reproduce the issue even with the changes from this PR included when a line has more than one tab character.
For example, in the below sample Go code, there are two tab characters at the beginning of the "if err = fmt." line. 
The diagnostic virtual text is pointing to the wrong column.

```
package main
import "fmt"

func main() {
      var err error
            if err = fmt.<cursor is here>  err != nil {                         
      }
  }
```

berggeist:
My provided PR should work correctly for lsp. I've just verified it with your example. 
But you're right, there is also a bug in vim that causes problems with indentation of lines following virtual text.
I've already filed an issue in the vim repository this afternoon.
You can work around it by changing the width of the window. Then also the (wrong) indentation of the line changes.

PR #224, opened 2023-04-07
Use cwd as rootUri and initial workspaceFolder for files within cwd

andlrc:
- Use the CWD as the "rootUri" and initial "workspaceFolder"
- Use the parent directory for files not a descendent of the cwd

I think this makes more sense as a default, and it's also almost backward compatible for before ca7f26773bf4348d956e16d3d9dba82ca87c180c

It's hard to add a test as we reuse the LSP for every test.

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -153,14 +153,14 @@ def InitServer(lspserver: dict<any>, bnr: number)
       }
 
   # Compute the rootpath (based on the directory of the buffer)
-  var bufDir = bnr->bufname()->fnamemodify(':p:h')
   var rootPath = ''
   var rootSearchFiles = lspserver.rootSearchFiles
   if !rootSearchFiles->empty()
+    var bufDir = bnr->bufname()->fnamemodify(':p:h')
     rootPath = util.FindNearestRootDir(bufDir, rootSearchFiles)
   endif
   if rootPath == ''
-    rootPath = bufDir
+    rootPath = getcwd()
   endif
 
   lspserver.workspaceFolders = [rootPath]
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -155,12 +155,22 @@ def InitServer(lspserver: dict<any>, bnr: number)
   # Compute the rootpath (based on the directory of the buffer)
   var rootPath = ''
   var rootSearchFiles = lspserver.rootSearchFiles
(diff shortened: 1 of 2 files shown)


autoload/lsp/lspserver.vim
```diff
@@ -72,7 +72,15 @@ def StartServer(lspserver: dict<any>, bnr: number): number
   lspserver.completionLazyDoc = false
   lspserver.completionTriggerChars = []
   lspserver.signaturePopup = -1
-  lspserver.workspaceFolders = [getcwd()]
+  var bufpath = bnr->bufname()->fnamemodify(':p')
+  var cwd = getcwd()
+
+  # bufpath is within cwd
+  if bufpath[0 : cwd->strcharlen() - 1] == cwd
```
andlrc, line 79: In theory then this could fail on Windows as paths are incase sensitive.

autoload/lsp/lspserver.vim
```diff
@@ -154,22 +153,36 @@ def InitServer(lspserver: dict<any>, bnr: number)
...
     rootPath = util.FindNearestRootDir(bufDir, rootSearchFiles)
   endif
   if rootPath == ''
-    rootPath = bufDir
+    var cwd = getcwd()
+
+    # bufDir is within cwd
+    var bufDirPrefix = bufDir[0 : cwd->strcharlen() - 1]
+    if &fileignorecase
+        ? bufDirPrefix->tolower() == cwd->tolower()
```
yegappan, line 168: Will using the "==?" comparison operator work here instead of the tolower() function?

doc/lsp.txt
```diff
@@ -270,9 +270,10 @@ To add a language server, the following information is needed:
 			multiple directories are found, then the directory
 			closest to the directory of the current buffer is used
 			as the workspace root.  If this parameter is not
-			specified or the files are not found, then the
-			directory of the current buffer is used as the
-			workspace root.
+			specified or the files are not found, then the current
+			wording directory is used as the workspace root for
```
yegappan, line 274: Typo: wording -> working.

autoload/lsp/lspserver.vim
andlrc: Yes of cause, I have changed it to `==?` and `==#` even though `==` is ignoring case in vim9.

doc/lsp.txt
andlrc: Thanks, fixed.

PR #225, opened 2023-04-08
Make it possible to have multiple servers registred for a given filetype

### LINKED ISSUE #221: root uri option
I want to use Deno LSP and TypeScript LSP and I don't want to run those LSPs at the same time. vim-lsp has this option, for example: https://github.com/prabirshrestha/vim-lsp/wiki/Servers-TypeScript

andlrc:
⚠️ This PR doesn't add support for running multiple LSP servers for a buffer, but rather adds support for registering multiple LSP servers for a given filetype.

The server with the longest workspace root will be chosen as the most relevant server for a given buffer.

Closes #221

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -26,7 +26,7 @@ import './inlayhints.vim'
 var lspServers: list<dict<any>> = []
 
 # filetype to LSP server map
-var ftypeServerMap: dict<dict<any>> = {}
+var ftypeServerMap: dict<list<dict<any>>> = {}
 
 var lspInitializedOnce = false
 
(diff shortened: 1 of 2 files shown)


autoload/lsp/lsp.vim
```diff
@@ -45,13 +45,41 @@ enddef
...
+    var bufDir = bnr->bufname()->fnamemodify(':p:h')
+
+    for lspserver in lspservers
+      var score: number = 0
+      if !lspserver.rootSearchFiles->empty()
+        # The score is calculated by how deep the workspace root dir is, the
+        # deeper the better.
+        var path = util.FindNearestRootDir(bufDir, lspserver.rootSearchFiles)
+        score = path->strcharlen()
+      endif
```
andlrc, line 65: Maybe support should instead be added to run multiple language servers for a given filetype, and then another option should be added called `runIf` which takes a function with the current `bnr`:

    runIf: (bnr) => {
      return lsp#util#FindNearestRootDir(bnr->bufname()->fnamemodify(':p:h'), [
        'package.json'
      ])
    }

PR #226, opened 2023-04-10
fix: init workspace list was wrong if set rootSearch

Shane-XB-Qian:
// should respect the value of rootSearch after that feature.


andlrc:
I think this makes sense, I have included the commit in #224 as to make merging easier.

Issue #228, opened 2023-04-10
Multiple servers support


maximyurevich:

andlrc:
How would things like `LspGotoDefinition`, `LspPeekDefinition`, `LspPeekReferences` etc. work if two servers reports different locations?

How would things like `LspCodeAction` work? Should the actions be merged into one list? If so what happens with potential "duplicates"?

What happens with `LspFold` if the two servers reports different folds, `LspHighlight` if the server reports different positions, etc? Should the user have to specify which running server that should be used?

I think that there are many questions that should be answered before it makes sense to add support for multiple servers? Is there somewhere else that we can draw inspiration from?

maximyurevich:
> How would things like `LspGotoDefinition`, `LspPeekDefinition`, `LspPeekReferences` etc. work if two servers reports different locations?
> 
> How would things like `LspCodeAction` work? Should the actions be merged into one list? If so what happens with potential "duplicates"?
> 
> What happens with `LspFold` if the two servers reports different folds, `LspHighlight` if the server reports different positions, etc? Should the user have to specify which running server that should be used?
> 
> I think that there are many questions that should be answered before it makes sense to add support for multiple servers. Is there somewhere else that we can draw inspiration from? I'm sure VSCode have this stuff figured out?

For inspiration https://github.com/prabirshrestha/vim-lsp
I want to add [efm-langserver](https://github.com/mattn/efm-langserver) as second LSP for linting and formatting

andlrc:
> For inspiration https://github.com/prabirshrestha/vim-lsp

I not sure I understand how in works in the linked plugin, can you show a few examples for how the above mentioned concerns are addressed?

yegappan:
I did a google search for the multiple language servers per file support and found the following relevant links:

https://github.com/emacs-lsp/lsp-mode/issues/424
https://github.com/Microsoft/vscode-extension-samples/tree/main/lsp-multi-server-sample
https://github.com/helix-editor/helix/pull/2507
https://neovim.discourse.group/t/multiple-parallel-langauge-servers/235
https://www.reddit.com/r/neovim/comments/jy08md/multiple_languageserver_choose_format/
https://neovim.discourse.group/t/automatically-choose-one-language-server-to-format-code-when-using-multiple-language-servers/800/7
https://github.com/natebosch/vim-lsc/issues/57

I think we should allow the user to configure which LSP server should provide a particular LSP functionality.
[...]

PR #231, opened 2023-04-11
Feat/multiple servers

andlrc:
Add support for running multiple servers at once.

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -132,11 +132,13 @@ export def ShowAllServers()
   lines->add('Buffer Information')
   lines->add('==================')
   for bnr in range(1, bufnr('$'))
-    if buf.BufHasLspServer(bnr)
+    var lspservers: list<dict<any>> = buf.BufLspServersGet(bnr)
+    if !lspservers->empty()
       lines->add($"Buffer: '{bufname(bnr)}'")
-      var lspserver = buf.BufLspServerGet(bnr)
-      lines->add($"Server Path: '{lspserver.path}'")
-      lines->add($"Status: {lspserver.running ? 'Running' : 'Not running'}")
+      for lspserver in lspservers
+        lines->add($"Server Path: '{lspserver.path}'")
+        lines->add($"Status: {lspserver.running ? 'Running' : 'Not running'}")
+      endfor
       lines->add('')
     endif
   endfor
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -678,12 +678,10 @@ def ServerTraceSet(traceVal: string)
     return
   endif
 
-  var lspserver: dict<any> = buf.CurbufGetServerChecked()
-  if lspserver->empty()
-    return
-  endif
-
-  lspserver.setTrace(traceVal)
+  var lspservers: list<dict<any>> = buf.CurbufGetServers()
+  for lspserver in lspservers
+    lspserver.setTrace(traceVal)
+  endfor
 enddef
 
 # Display the diagnostic messages from the LSP server for the current buffer


andlrc:
I suggest that we consider cherry-picking some of the first commits, as it should be easy to confirm that these works properly.

autoload/lsp/diag.vim
```diff
@@ -456,16 +473,52 @@ export def GetDiagByPos(lspserver: dict<any>, bnr: number,
...
+  for lspDiags in diagsMap[bnr]->values()
+
+    if lspDiags.sortedDiagnostics->empty()
+      continue
+    endif
+
+    allDiagnostics->extend(lspDiags.sortedDiagnostics)
+  endfor
+
+  return SortDiags(allDiagnostics)
```
andlrc, line 521: Currently the diagnostics are sorted a lot of times, first when they arrive, and later when each list of LSP diagnostics are being merged into one.

autoload/lsp/lspserver.vim
```diff
@@ -1523,8 +1522,6 @@ export def NewLspServer(name_arg: string, path_arg: string, args: list<string>,
     processNotif: function(handlers.ProcessNotif, [lspserver]),
     processRequest: function(handlers.ProcessRequest, [lspserver]),
     processMessages: function(handlers.ProcessMessages, [lspserver]),
-    getDiagByPos: function(diag.GetDiagByPos, [lspserver]),
-    getDiagsByLine: function(diag.GetDiagsByLine, [lspserver]),
```
andlrc, line 1542: These methods doesn't really make sense anymore as the diagnostics aren't really tied to a single LSP server.

autoload/lsp/lsp.vim
```diff
@@ -43,36 +43,16 @@ def LspInitOnce()
...
-        score = path->strcharlen()
-      endif
-      if score > bestScore
-        bestMatch = lspserver
-        bestScore = score
-      endif
-    endfor
-
-    return bestMatch
+    # TODO score based inclusion/exclusion
```
andlrc: Currently the code implemented in #228 is being broken by this PR.

What should we do about it?

autoload/lsp/buffer.vim
```diff
@@ -41,8 +41,35 @@ export def BufLspServerGet(bnr: number): dict<any>
...
+
+    if !lspserver.customProviding->has_key(domain)
+        || !lspserver.customProviding[domain](lspserver)
+      continue
+    endif
+
+    return lspserver
+  endfor
+
+  return {}
```
andlrc: The idea behind the code is to use the first server that actually supports a given action. But also letting the user overriding which language server which should be used by providing a: `customProving` (⚠️ needs a better name) dictionary with functions that returns boolean if a language server should support a action.

autoload/lsp/buffer.vim
```diff
@@ -41,8 +41,45 @@ export def BufLspServerGet(bnr: number): dict<any>
...
+  # LSP server is configured to be a provider for 'domain'
+  for lspserver in possibleLSPs
+    if lspserver.customProviding->has_key(domain)
+        && lspserver.customProviding[domain](lspserver)
+      return lspserver
+    endif
+  endfor
+
+  # Return the first LSP server that supports 'domain'
+  return possibleLSPs[0]
```
andlrc: The idea behind the code is to use the first server that actually supports a given action. But also letting the user overriding which language server which should be used by providing a: customProving (⚠️ needs a better name) dictionary with functions that returns boolean if a language server should support a action.

andlrc:
@maximyurevich please try this out and see if it fulfills your needs.

@yegappan This is still early draft, but it feels like it works. Maybe parts of it can be merged? The two opened threads needs some discussion at least.

autoload/lsp/lsp.vim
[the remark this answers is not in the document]
andlrc: I previously purposed adding a `runIf` function on the LSP config:

    { path: '....', args: ['...'],
      runIf: (bufnr) => bnr->bufname()->fnamemodify(':p:h') =~ '^/some/path'
    }

autoload/lsp/buffer.vim
```diff
@@ -41,8 +41,45 @@ export def BufLspServerGet(bnr: number): dict<any>
...
 enddef
```
andlrc: The idea behind the code is to use the first server that actually supports a given action. But also letting the user overriding which language server which should be used by providing a: customProving (⚠️ needs a better name) dictionary with functions that returns boolean if a language server should support a action.

autoload/lsp/buffer.vim
```diff
@@ -41,8 +41,36 @@ export def BufLspServerGet(bnr: number): dict<any>
...
+    if lspserver.customProviding->has_key(domain)
+        && !lspserver.customProviding[domain](lspserver)
+      continue
+    endif
+
+    return lspserver
+  endfor
+
+  # No server found that supports 'domain'
+  return {}
```
andlrc: The idea behind the code is to use the first server that actually supports a given action. But also letting the user overriding which language server which should be used by providing a: `customProving` (⚠️ needs a better name) dictionary with functions that returns boolean if a language server should support a action.

autoload/lsp/lsp.vim
```diff
@@ -46,13 +46,26 @@ enddef
...
+        && lspserver.customRunIf != null_function
+        && !lspserver.customRunIf(bnr)
+      continue
+    endif
+
+    filteredServers->add(lspserver)
+  endfor
+
+  return filteredServers
 enddef
```
andlrc: I have added support for the configuration `runIf` since `rootSearch` cannot be used anymore to determinate if a LSP should be used or not.

doc/lsp.txt
```diff
@@ -1367,6 +1367,29 @@ which servers should be used for a given action.  The following actions are
...
+			filetype: ['javascript', 'typescript'],
+			path: '/usr/local/bin/typescript-language-server',
+			args: ['--stdio']
+		},
+		{
+			filetype: ['efm-langserver'],
+			path: '/usr/local/bin/efm-langserver',
+			args: [],
+			customProviding: {
+				documentFormatting: (lspserver) => true
```
andlrc, line 1403: Maybe this should simply be booleans instead of Funcrefs?

yegappan: Yes.  That will make it simple from a user perspective.  Users may not be aware of the Vim funcref syntax and it will look complicated.

yegappan: What do you think about using the name "features" instead of "customProviding"?

autoload/lsp/lsp.vim
yegappan: It will be complicated for a user to write a function (or a lambda) to determine which LSP server should be used.  It is easier for them to specify a list of files or directories to search for to enable a particular LSP server.

andlrc: The only way that we can reuse `rootSearch` is if we change it to include all servers which finds the same root, and exclude all other servers. Otherwise we need to find another name that can be used. `runIfAncestorDirectoriesContains` doesn't feel good 🤔

doc/lsp.txt
andlrc: > What do you think about using the name "features" instead of "customProviding"?

Certainly better than "customProviding". I'll change it and then we can think a bit more about it.

autoload/lsp/lsp.vim
andlrc: Actually come to think of it, it might be fine in the first iteration to include all LSP servers that have the same score, that will also make us backward compatible.

andlrc: > Actually come to think of it, it might be fine in the first iteration to include all LSP servers that have the same score, that will also make us backward compatible.

I have changed it to that now, and therefore removed `runIf` again.

andlrc:
@yegappan I think that this _might_ be ready for being merged, as:

1. It backward compatible.
2. It feels like everything is working for when running with one LSP servers for a given filetype.
3. It feels like everything is working when running with multiple LSP servers for a given filetype.

I'm happy because of 1. and 2.

andlrc:
@Shane-XB-Qian you seem like you use quite a few different language servers. Is it possible for you to try out this PR and see if things works as expected for you? 😄

maximyurevich:
Second server works only after the command LspServer restart

andlrc:
> Second server works only after command LspServer restart

You really need to provide some more context @maximyurevich. What servers, what config, etc, what does your debug information states, etc...

maximyurevich:
> > Second server works only after command LspServer restart
> 
> You really need to provide some more context @maximyurevich. What servers, what config, etc, what does your debug information states, etc...

I cant find LspServerDebug command

andlrc:
> > > Second server works only after command LspServer restart
> > 
> > 
> > You really need to provide some more context @maximyurevich. What servers, what config, etc, what does your debug information states, etc...
> 
> I cant find LspServerDebug command

It was moved to `:LspServer debug ...` in b721ab27e2e1951a48dd4ec58752f43af28c346b

PR #234, opened 2023-04-11
feat: Add custom completion kind

D4yvid:
I've tested it, and worked properly, i wanted this for too long time, so, i've implemented on my own

image

### DIFF
--- a/doc/lsp.txt
+++ b/doc/lsp.txt
@@ -444,7 +444,7 @@ snippetSupport		|Boolean| option.  Enable snippet completion support.
 			Need a snippet completion plugin like vim-vsnip.
 			By default this is set to false.
 						*lsp-opt-ultisnipsSupport*
-ultisnipsSupport		|Boolean| option.  Enable SirVer/ultisnips support.
+ultisnipsSupport	|Boolean| option.  Enable SirVer/ultisnips support.
 			Need a snippet completion plugin SirVer/ultisnips.
 			By default this is set to false.
 						*lsp-opt-usePopupInCodeAction*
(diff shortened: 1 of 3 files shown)


doc/lsp.txt
```diff
@@ -444,14 +444,20 @@ snippetSupport		|Boolean| option.  Enable snippet completion support.
...
 usePopupInCodeAction    |Boolean| option.  When using the |:LspCodeAction|
 			command to display the code action for the current
 			line, use a popup menu instead of echoing.
 			By default this is set to false.
+						*lsp-opt-customCompletionKinds*
+customCompletionKinds   |Boolean| option.  If you set this to true, you can set
+			custom completion kinds using the option completionKinds.
+						*lsp-opt-completionKinds*
+completionKinds		|Dictionary| option. Here you can set custom completion
+			kinds. To view all completion kind names, use :h lsp-custom-kinds
```
andlrc, line 460: I think that if you change it to `|lsp-custom-kinds|` then you can jump to it directly from the help page. `See |lsp-custom-kinds| for all completion kind names`.

D4yvid: Ok! I'll change it now

D4yvid: Done.

andlrc:
I get the following error while trying to create help tags:

```
:helptags ALL
E154: Duplicate tag "lsp-opt-completionKinds" in file /Users/and/.vim/pack/plugins/opt/andlrc-lsp/doc/lsp.txt
E154: Duplicate tag "lsp-opt-customCompletionKinds" in file /Users/and/.vim/pack/plugins/opt/andlrc-lsp/doc/lsp.txt
```

I think the following diff will resolve that:

```
diff --git a/doc/lsp.txt b/doc/lsp.txt
index 93a99e1..cbb8493 100644
--- a/doc/lsp.txt
+++ b/doc/lsp.txt
@@ -1233,8 +1233,8 @@ more information about the "Command" interface.
 
 When a completion popup is triggered, the LSP client will use a default kind
 list to show in the completion "kind" section, to customize it, you need to
-use the option *lsp-opt-customCompletionKinds* and set all custom kinds in the
-option *lsp-opt-completionKinds* . There is a table with all default LSP kinds:
+use the option |lsp-opt-customCompletionKinds| and set all custom kinds in the
+option |lsp-opt-completionKinds| . There is a table with all default LSP kinds:
 
  Kind Name		| Value
 ------------------------|--------------------
```

D4yvid:
Sorry, i'll apply those diffs, i've never used vim documentation

D4yvid:
Done!

PR #250, opened 2023-04-13
feat: make additionalTextEdits be possible to disable

### LINKED ISSUE #233: Golang double import with noNewlineInCmpletion: v:true
My \<cr\> bind 
inoremap <expr> \<cr\>    pumvisible() ? "\<C-y>": "\<cr>"

Shane-XB-Qian:
closes #233

### DIFF
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -33,6 +33,9 @@ export var lspOptions: dict<any> = {
   ignoreMissingServer: false,
   # Focus on the location list window after LspShowReferences
   keepFocusInReferences: false,
+  # If not apply LSP server-initiated actions after completion
+  # specially if some others plugins had registered to do that
+  noAdditionalTextEdits: false,
   # instead of the signature
   noDiagHoverOnLine: true,
   # Suppress adding a new line on completion selection with <CR>
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -33,9 +33,10 @@ export var lspOptions: dict<any> = {
   ignoreMissingServer: false,
   # Focus on the location list window after LspShowReferences
   keepFocusInReferences: false,
-  # If not apply LSP server-initiated actions after completion
-  # specially if some others plugins had registered to do that
-  noAdditionalTextEdits: false,
+  # If true, apply the LSP server supplied text edits after a completion.
+  # If a snippet plugin is going to apply the text edits, then set this to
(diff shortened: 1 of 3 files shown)


autoload/lsp/options.vim
```diff
@@ -33,6 +33,9 @@ export var lspOptions: dict<any> = {
   ignoreMissingServer: false,
   # Focus on the location list window after LspShowReferences
   keepFocusInReferences: false,
+  # If not apply LSP server-initiated actions after completion
+  # specially if some others plugins had registered to do that
+  noAdditionalTextEdits: false,
```
yegappan, line 38: Can you rename this to "completionTextEdit" and set the default to true?

autoload/lsp/options.vim
```diff
@@ -33,6 +33,9 @@ export var lspOptions: dict<any> = {
   ignoreMissingServer: false,
   # Focus on the location list window after LspShowReferences
   keepFocusInReferences: false,
+  # If not apply LSP server-initiated actions after completion
+  # specially if some others plugins had registered to do that
```
yegappan, line 37: Can you change the comment to:
```
# If true, apply the LSP server supplied text edits after a completion. 
# If a snippet plugin is going to apply the text edits, then set this to false to 
# avoid applying the text edits twice.
```

doc/lsp.txt
```diff
@@ -393,11 +393,17 @@ ignoreMissingServer	|Boolean| option.  Do not print a missing language
 						*lsp-opt-keepFocusInReferences*
 keepFocusInReferences	|Boolean| option.  Focus on the location list window
 			after LspShowReferences.
+			By default this is set to false.
+						*lsp-opt-noAdditionalTextEdits*
+noAdditionalTextEdits	|Boolean| option.  If not apply LSP server-initiated
+			actions after completion, specially if some others
+			plugins had registered to do that.
```
yegappan, line 400: Can you use the text from the above comment and update the help text?

yegappan (changes requested):
Thanks for the fix.

autoload/lsp/options.vim
Shane-XB-Qian: ok.

autoload/lsp/options.vim
Shane-XB-Qian: done.

doc/lsp.txt
Shane-XB-Qian: ok.

PR #254, opened 2023-04-14
Fix misc tidy or flaw

Shane-XB-Qian:


doc/lsp.txt
```diff
@@ -380,6 +380,18 @@ completionMatcher	|String| option.  Enable fuzzy or case insensitive
 			    case  - case sensitive matching (default).
 			    fuzzy - fuzzy match completion items.
 			    icase - ignore case when matching items.
+						*lsp-opt-noAdditionalTextEdits*
+completionTextEdit	|Boolean| option.  If true, apply the LSP server
```
andlrc, line 384: Can you change `lsp-opt-noAdditionalTextEdits` to `lsp-opt-completionTextEdit`?

Shane-XB-Qian: done.

autoload/lsp/util.vim
```diff
@@ -34,7 +34,13 @@ enddef
 
 # Empty out the LSP server trace logs
 export def ClearTraceLogs(fname: string)
-  writefile([], fname)
+  var fullname = $'{lsp_log_dir}{fname}'
+  # If file existed but not writable
+  if filewritable(fullname) != 1 && filereadable(fullname)
```
yegappan, line 39: If the log and error files are not writable, then TraceLog() function also will fail.  So checking whether the file is writable or not here is not sufficient.

The bug here is that the directory name is not prepended to the file name.  Can you change this function to:
```
    writefile([], $'{lsp_log_dir}{fname}'
```

autoload/lsp/lspserver.vim
```diff
@@ -155,12 +155,12 @@ def InitServer(lspserver: dict<any>, bnr: number)
   # Compute the rootpath (based on the directory of the buffer)
   var rootPath = ''
   var rootSearchFiles = lspserver.rootSearchFiles
-  var bufDir = bnr->bufname()->fnamemodify(':p:h')
+  var bufDir = bnr->bufname()->resolve()->fnamemodify(':p:h')
```
yegappan, line 158: If you resolve a symbolic link in Unix, then the root search will find the wrong root directory.  The symbolic link should not be resolved.

autoload/lsp/util.vim
Shane-XB-Qian: yes, the main point was it should be `fullname` (vs `fname`), 
// but before tracelog it would be clearlog first, so did a writable check (into to msg) was a hint, no harm.

autoload/lsp/lspserver.vim
Shane-XB-Qian: it did with `cwd` together, it should be fine, which just to avoid one is symblink, another was not, then it would be wrong.

Shane-XB-Qian: or whatever, now removed it.

PR #262, opened 2023-04-15
feat: make 'hover' be able to no-echo

Shane-XB-Qian:
// 'hover' perhaps was freq used,
// but its 'no found' echo msg may interrupt e.g signature msg (if set to echo at cmdline too).


andlrc:
You can also use `silent! LspHover` instead of adding the `!`.

andlrc:
I though that you could use `silent LspHover` but since the hover request is async it will not work.

autoload/lsp/lspserver.vim
```diff
@@ -832,7 +832,8 @@ def ShowHoverInfo(lspserver: dict<any>): void
   # interface HoverParams
   #   interface TextDocumentPositionParams
   var params = GetLspTextDocPosition(false)
-  lspserver.rpc_a('textDocument/hover', params, hover.HoverReply)
+  lspserver.rpc_a('textDocument/hover', params,
+			function(hover.HoverReply, [noEcho]))
```
andlrc, line 836: I assume that you can use a lambda instead:

```
lspserver.rpc_a('textDocument/hover', params, (lspserver, reply) => {
    hover.HoverReply(lspserver, reply, noEcho)
})
```

if you want `noEcho` to be the last argument instead of the first. I think it makes sense to strive to always have the `lspserver` and `reply` as the first arguments

test/clangd_tests.vim
```diff
@@ -975,7 +975,7 @@ def g:Test_LspHover()
   popup_close(p[0])
   cursor(7, 1)
   output = execute(':LspHover')->split("\n")
-  assert_equal('No hover messages found for current position', output[0])
+  assert_match('No hover messages found for current position', output[0])
```
andlrc, line 978: Can you add a test for `LspHover!` as well?

    output = execute(':LspHover!')->split("\n")
    assert_equal('', output[0])

autoload/lsp/lspserver.vim
Shane-XB-Qian: thx for tips. 
but i am not plan to refactor this code, i just kept the same code style like others in this repo.
// maybe we can refactor it later.

test/clangd_tests.vim
Shane-XB-Qian: as you said, it is async, actually i do not think that assert really worked.
// maybe we can refactor/remove such assert later.

andlrc: The tests are not async because g:LSP_Test is set to 1

Shane-XB-Qian: good catch.

autoload/lsp/hover.vim
```diff
@@ -72,12 +72,14 @@ enddef
 
 # process the 'textDocument/hover' reply from the LSP server
 # Result: Hover | null
-export def HoverReply(lspserver: dict<any>, hoverResult: any): void
+export def HoverReply(noEcho: bool, lspserver: dict<any>, hoverResult: any): void
```
yegappan, line 75: Can you change "noEcho" to "silent" and move this to the last parameter?  Most of the functions take lspserver as the first argument.

doc/lsp.txt
```diff
@@ -675,10 +675,12 @@ can map these commands to keys and make it easier to invoke them.
...
 
 						*:LspHover*
-:LspHover		Show the documentation for the symbol under the cursor
-			in a popup window. If you want to show the symbol
-			documentation in the preview window instead of in a
-			popup set >
+:LspHover[!]		Show the documentation for the symbol under the cursor
+			in a popup window. When "!" used, then no info message
+			about no hover found would be echoed, that helped to
+			less interrupt other cmd's messages. If you want to
```
yegappan, line 681: Can you change the new help text to:
```
When "!" is used, if no hover information is found at the current position, then the command will not echo any message (silent).
```

yegappan (changes requested):

Shane-XB-Qian:
> Can you change "noEcho" to "silent"


'silent is a vim keyword, put it as user func argument would make its syntax highligh looks incorrect and may a bit confused.

so I was thinking name it to 'noecho'.


> and move this to the last parameter?  Most of the functions take lspserver as the first argument.


yes. I think it should be. but to those funcs in this file actually they are all put additional args ahead than lspserver.

we thought perhaps we can refactor this later?

// it is not only this func put lspserver not the first argument. it req some trick to make it actually.

--
shane.xb.qian

andlrc:
> // it is not only this func put lspserver not the first argument. it req some trick to make it actually.

You can use a lambda as I showed here https://github.com/yegappan/lsp/pull/262#discussion_r1167563260

autoload/lsp/hover.vim
Shane-XB-Qian: done.
modified 'noecho' to 'silent'.
// as for arg place order, let's refactor it later.

doc/lsp.txt
Shane-XB-Qian: done. modified.

andlrc:
I added this PR https://github.com/yegappan/lsp/pull/266 which makes it possible to use `silent LspHover` as well as other modifiers for the preview window. I like that approach instead of `LspHover!`, as this can also be added to other commands easily.

Shane-XB-Qian:
> use a lambda

@andlrc i am not sure how it work actually, would you mind refactor this plus previous those same case all to that later?
// if you not mind, please open another PR to refactor it after this all merged.

Shane-XB-Qian:
> I added this PR #266 which makes it possible to use `silent LspHover` as well as other modifiers for the preview window. I like that approach instead of `LspHover!`, as this can also be added to other commands easily.

could we not do something repeatly and conflict each others? you added `showdiagcurrent[!]` with a `bang` approach, i just followed your style. now had to revert it back?!

andlrc:
> could we not do something repeatly and conflict each others? you added `showdiagcurrent[!]` with a `bang` approach, i just followed your style. now had to revert it back?!

The bang in `LspDiagCurrent!` is different as it, unlike `LspDiagCurrent`, dosn't scan forward for a diagnostic, and therefore instructs `LspDiagCurrent` to only show a diagnostic if it's under the cursor.

Shane-XB-Qian:
> It's therefore I purpose that we could use `<mods>` so that `silent LspHover` will actually behave as expected.

really? 
this feature just wish it's not to echo regular 'no found' such and perhaps its freq happened message,
but it did NOT prevent the msg e.g lsp server did not support hover, a.k.a it would still echo some message even it emit `LspHover!` with a `bang` with it.
clear?

andlrc:
> 

I understand fully, but I still think it's bad to use the bang for silencing an error message when things like `:h <mods>` exists.

Shane-XB-Qian:
ok, you can add all async commands in this plugin to support 'silent' <mods> later or whenever you like,

but this feature this PR wish to get hover error message still, 
vs/but JUST to prevent its regular and repeat and freq `not found` message.

autoload/lsp/hover.vim
```diff
@@ -72,12 +72,14 @@ enddef
 
 # process the 'textDocument/hover' reply from the LSP server
 # Result: Hover | null
-export def HoverReply(lspserver: dict<any>, hoverResult: any): void
+export def HoverReply(silent: bool, lspserver: dict<any>, hoverResult: any): void
```
yegappan, line 75: As much as possible, all the functions that deal with the server take the lspserver as the first argument.  Also I prefer the most significant input parameters to be earlier in the order of arguments.  Can you move the "bool" argument to the end?

Shane-XB-Qian: i think it should be, but i had explained, actually i do not know how.
please check: 
`export def DocSymbolReply(fname: string, lspserver: dict<any>, docsymbol: any)`
and
`def DocHighlightReply(bnr: number, lspserver: dict<any>, docHighlightReply: any): void`
etc.
would you mind refactor those all later by you or someone else?

Shane-XB-Qian:
i give up, since you preferred his style (again).
// closing.

PR #264, opened 2023-04-16
Fix tidy echom

Shane-XB-Qian:
~~// based on #262~~

### DIFF
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -41,8 +41,13 @@ def ProcessShowMsgNotif(lspserver: dict<any>, reply: dict<any>)
     # them.
     return
   endif
-  var mtype = LspMsgTypeToString(reply.params.type)
-  :echomsg $'Lsp({lspserver.name}):[{mtype}]: {reply.params.message}'
+  if reply.params.type == 1
+    util.ErrMsg($'Lsp({lspserver.name}) {reply.params.message}')
+  elseif reply.params.type == 2
+    util.WarnMsg($'Lsp({lspserver.name}) {reply.params.message}')
+  elseif reply.params.type == 3
+    util.InfoMsg($'Lsp({lspserver.name}) {reply.params.message}')
+  endif
 enddef
 
 # process a log notification message from the LSP server
@@ -123,7 +128,7 @@ def ProcessApplyEditReq(lspserver: dict<any>, request: dict<any>)
   endif
   var workspaceEditParams: dict<any> = request.params
   if workspaceEditParams->has_key('label')
-    :echomsg $'Workspace edit {workspaceEditParams.label}'
+    util.InfoMsg($'Workspace edit {workspaceEditParams.label}')
   endif
   textedit.ApplyWorkspaceEdit(workspaceEditParams.edit)
   # TODO: Need to return the proper result of the edit operation
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -34,7 +34,7 @@ enddef
 # Notification: window/showMessage
 # Param: ShowMessageParams
 def ProcessShowMsgNotif(lspserver: dict<any>, reply: dict<any>)
-  if reply.params.type == 4
+  if reply.params.type >= 4
     # ignore log messages from the LSP server (too chatty)
     # TODO: Add a configuration to control the message level that will be
     # displayed. Also store these messages and provide a command to display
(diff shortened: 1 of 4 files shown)


Shane-XB-Qian:
@yegappan ready to merge if ok.

autoload/lsp/handlers.vim
```diff
@@ -41,8 +41,13 @@ def ProcessShowMsgNotif(lspserver: dict<any>, reply: dict<any>)
...
   endif
-  var mtype = LspMsgTypeToString(reply.params.type)
-  :echomsg $'Lsp({lspserver.name}):[{mtype}]: {reply.params.message}'
+  if reply.params.type == 1
+    util.ErrMsg($'Lsp({lspserver.name}) {reply.params.message}')
+  elseif reply.params.type == 2
+    util.WarnMsg($'Lsp({lspserver.name}) {reply.params.message}')
+  elseif reply.params.type == 3
+    util.InfoMsg($'Lsp({lspserver.name}) {reply.params.message}')
+  endif
```
andlrc, line 50: The specifications states that there is a type = 4, called hint. I think you should add that as well, as well a fallback if type > 4

autoload/lsp/hover.vim
```diff
@@ -72,12 +72,14 @@ enddef
...
 # Result: Hover | null
-export def HoverReply(lspserver: dict<any>, hoverResult: any): void
+export def HoverReply(noEcho: bool, lspserver: dict<any>, hoverResult: any): void
   var [hoverText, hoverKind] = GetHoverText(lspserver, hoverResult)
 
   # Nothing to show
   if hoverText->empty()
-    util.WarnMsg($'No hover messages found for current position')
+    if !noEcho
+      util.InfoMsg($'No hover messages found for current position')
```
andlrc, line 81: I think this should stay as a warning, as to be consistent with all the other messages:

    util.WarnMsg('Error: No alternate file found')
    util.WarnMsg('Error: No references found')
    ...

autoload/lsp/handlers.vim
Shane-XB-Qian: the specification said it only had 4 level, but ok let me added a safeguard as well.

autoload/lsp/hover.vim
Shane-XB-Qian: i do not think it is a warn, but just a regular message, but ok, it's not a matter. modified.

autoload/lsp/hover.vim
```diff
@@ -72,12 +72,14 @@ enddef
 
 # process the 'textDocument/hover' reply from the LSP server
 # Result: Hover | null
-export def HoverReply(lspserver: dict<any>, hoverResult: any): void
+export def HoverReply(silent: bool, lspserver: dict<any>, hoverResult: any): void
```
yegappan, line 75: Can you remove the hover related changes from this PR?

autoload/lsp/textedit.vim
```diff
@@ -60,7 +60,6 @@ def Set_lines(lines: list<string>, A: list<number>, B: list<number>,
 
   var new_lines_len: number = new_lines->len()
 
-  #echomsg $"i_0 = {i_0}, i_n = {i_n}, new_lines = {string(new_lines)}"
```
yegappan, line 63: I left these debug messages in the plugin so that these can be used later to debug issues with the Set_lines() function.
Instead of removing these debug messages, we can change the call to :echomsg to instead use Debug() function.

yegappan (changes requested):

autoload/lsp/textedit.vim
Shane-XB-Qian: this PR based on #262

autoload/lsp/hover.vim
Shane-XB-Qian: this PR based on #262

autoload/lsp/textedit.vim
Shane-XB-Qian: the record still in the git history if you like to bring it back in case to debug or something.
and yes, if you preferred, i think we can add a util.DebugMsg() later, for tidy, so i just removed them.

autoload/lsp/hover.vim
Shane-XB-Qian: removed.

autoload/lsp/textedit.vim
Shane-XB-Qian: ok, reverted it as you wish.

PR #267, opened 2023-04-16
Raise an error if no hover server is found

andlrc:
The error message is something like: `Error: Language server for "xx" file type is not found`

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -738,8 +738,8 @@ enddef
 # Display the hover message from the LSP server for the current cursor
 # location
 export def Hover(cmdmods: string)
-  var lspserver: dict<any> = buf.CurbufGetServer('hover')
-  if lspserver->empty() || !lspserver.running || !lspserver.ready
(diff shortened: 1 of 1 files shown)


autoload/lsp/lsp.vim
```diff
@@ -738,8 +738,8 @@ enddef
 # Display the hover message from the LSP server for the current cursor
 # location
 export def Hover()
-  var lspserver: dict<any> = buf.CurbufGetServer('hover')
-  if lspserver->empty() || !lspserver.running || !lspserver.ready
+  var lspserver: dict<any> = buf.CurbufGetServerChecked('hover')
```
yegappan, line 741: Similar to the functionality offered in other IDEs like VSCode, I wanted to support using an autocmd to display the hover information under the current cursor:
```
autocmd CursorHold * LspHover
```
In this case, we don't want to display an error message every time Hover() is invoked.  That is why, the CurbufGetServer() is used here.

andlrc: Since #266 have been merged, I would think that the user can use:

    autocmd CursorHold * silent LspHover

PR #269, opened 2023-04-16
Add an silenceable via <mods> error message to LspHighlight

andlrc:
Add an error message to `:LspHighlight` when it's not possible to highlight the keyword, but allow the user to silence it with `:silent LspHighlight`


autoload/lsp/lspserver.vim
```diff
@@ -862,8 +862,12 @@ enddef
 
 # process the 'textDocument/documentHighlight' reply from the LSP server
 # Result: DocumentHighlight[] | null
-def DocHighlightReply(bnr: number, lspserver: dict<any>, docHighlightReply: any): void
+def DocHighlightReply(lspserver: dict<any>, docHighlightReply: any,
+                      bnr: number, cmdmods: string): void
   if docHighlightReply->empty()
+    if cmdmods !~ 'silent'
+      util.WarnMsg($'No highlight for the current position')
```
yegappan, line 869: When the "autoHighlight" option is set to true, this message will be displayed if the current position doesn't have a highlight.  So this message will be displayed many times.

andlrc: Makes sense, I have changed the code to send `silent` to the `LspDocHighlight` function from within the CursorMoved autocmd.

autoload/lsp/lsp.vim
```diff
@@ -395,7 +395,7 @@ def AddBufLocalAutocmds(lspserver: dict<any>, bnr: number): void
     acmds->add({bufnr: bnr,
 		event: 'CursorMoved',
 		group: 'LSPBufferAutocmds',
-		cmd: 'call LspDocHighlightClear() | call LspDocHighlight()'})
+		cmd: 'call LspDocHighlightClear() | call LspDocHighlight("silent")'})
```
andlrc, line 398: This silences the error for the autocmd.

Issue #260, opened 2023-04-15
Proposal: New option for alignment of virtual diagnostic text


berggeist:
I would like to propose a new option for the alignment of the virtual diagnostic text.

Currently the default behavior, if the existing option `showDiagWithVirtualText` is set to `v:true`, is that the diagnostic virtual text messages are displayed above the related lines, as can be seen in the following example:
![lsp-virtual-text-align-above](https://user-images.githubusercontent.com/24396556/232216748-807f2182-01be-44bf-82b7-e956b5494662.png)
In this case all diagnostic messages are truncated and each diagnostic message is displayed in its own virtual text line.

---

Now the proposal is, to introduce an additional option `diagVirtualTextAlign` that can take one of the following values:
- `above` , the default value
- `after`
- `below`

---

`above` is the default value if the new option `diagVirtualTextAlign` is not customized. In this case, the currently implemented behavior is preserved. An example is already given above.

---

`after` will display the virtual diagnostic text messages in the same line but after the related line of text. Here is an example:
![lsp-virtual-text-align-after](https://user-images.githubusercontent.com/24396556/232218143-840aeddb-f387-41b8-bcd3-fb45324a104a.png)
Now multiple diagnostic messages are joined within one virtual text line and prefixed by the according symbol for the severity. As the joined messages might become rather long, in this case, the virtual text is not truncated but wrapped.

---

`below` is more or less the opposite to `above`. I.e. in this case, the virtual text for the diagnostic messages are aligned below the related line of text:
![lsp-virtual-text-align-below](https://user-images.githubusercontent.com/24396556/232219249-9b105d4d-7d74-4fa6-ab4a-dea01e665d46.png)

---

I would like to hear your opinion, whether this proposal would be helpful and whether you agree that I create a PR for this new feature.

yegappan:
I think this option will be useful for some users who prefer the multiple diagnostics to be shown after the line.  Can you go ahead and create a PR for this?

andlrc:
> `below` is more or less the opposite to `above`. I.e. in this case, the virtual text for the diagnostic messages are aligned below the related line of text:
> ![lsp-virtual-text-align-below](https://user-images.githubusercontent.com/24396556/232219249-9b105d4d-7d74-4fa6-ab4a-dea01e665d46.png)

This is just nitpick, but shouldn't the two diagnostics be flipped around, so that "Cannot find name 'ddd' is rendered above "Variable 'ccc' implicitly ...."

[...]

PR #273, opened 2023-04-16
Adds option to align diagnostics virtual text

berggeist:
This is a PR that addresses #260 

This PR provides a new option that allows the configuration of the alignment of the virtual text, which is displayed if the option `showDiagWithVirtualText` is active.

This  is a (first) minimal solution in order to get this new option going. In my opinion without any risks, since the former implementation has been preserved as far as possible.

This PR does not (yet) address the remarks from @andlrc, concerning the order of messages for `below` and @Shane-XB-Qian concerning the limitation of displayed diagnostic messages. In order to resolve these remarks, a more complicated solution needs to be prepared. 

I want to ask you kindly to review all changes, as this is my first larger PR into this code repository.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -116,6 +116,19 @@ def DiagSevToInlineHLName(severity: number): string
   return typeMap[severity - 1]
 enddef
 
+def DiagSevToSymbolText(severity: number): string
+  var typeMap: list<string> = [
+    opt.lspOptions.diagSignErrorText,
+    opt.lspOptions.diagSignWarningText,
+    opt.lspOptions.diagSignInfoText,
+    opt.lspOptions.diagSignHintText
+  ]
+  if severity > 4
+    return opt.lspOptions.diagSignHintText
+  endif
+  return typeMap[severity - 1]
+enddef
+
 # Refresh the placed diagnostics in buffer 'bnr'
 # This inline signs, inline props, and virtual text diagnostics
 def DiagsRefresh(bnr: number)
@@ -141,6 +154,24 @@ def DiagsRefresh(bnr: number)
     return
   endif
 
+  # Initialize default/fallback properties for diagnostic virtual text:
+  var diag_align: string = 'above'
+  var diag_wrap: string = 'truncate'
+  var diag_symbol: string = '┌─'
+  var diag_padding: number = 0
+
+  if opt.lspOptions.diagVirtualTextAlign ==? 'below'
+    diag_align = 'below'
+    diag_wrap = 'truncate'
+    diag_symbol = '└─'
+    diag_padding = 0
+  elseif opt.lspOptions.diagVirtualTextAlign ==? 'after'
+    diag_align = 'after'
+    diag_wrap = 'wrap'
+    diag_symbol = 'E>'
+    diag_padding = 3
+  endif
+
   var signs: list<dict<any>> = []
   var diags: list<dict<any>> = diagsMap[bnr].sortedDiagnostics
   for diag in diags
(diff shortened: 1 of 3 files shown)


autoload/lsp/diag.vim
```diff
@@ -141,6 +154,24 @@ def DiagsRefresh(bnr: number)
     return
   endif
 
+  # Initialize default/fallback properties for diagnostic virtual text:
+  var diag_align: string = 'above'
+  var diag_wrap: string = 'truncate'
+  var diag_symbol: string = '┌─'
+  var diag_padding: number = 0
+
+  if opt.lspOptions.diagVirtualTextAlign ==? 'below'
```
andlrc, line 163: `==?` matches case insensitive, making `'BeLOw' ==? 'below', is this intentional?

autoload/lsp/diag.vim
```diff
@@ -141,6 +154,24 @@ def DiagsRefresh(bnr: number)
     return
   endif
 
+  # Initialize default/fallback properties for diagnostic virtual text:
+  var diag_align: string = 'above'
+  var diag_wrap: string = 'truncate'
+  var diag_symbol: string = '┌─'
+  var diag_padding: number = 0
```
andlrc, line 161: Seems like this variable is a bit useless as you can simply set `padding = 3` for the 'after' condition in the loop?

autoload/lsp/diag.vim
berggeist: Yes this was my intention to be more resilient against typos made by the user. Do you see any drawbacks?

andlrc: Only that `lsp-opt-completionMatcher` doesn't work like that. But we could make it work that way.  😄

autoload/lsp/diag.vim
berggeist: My intention here was to define all *magic* numbers and settings at the same place. But no problem for me to eliminate this variable `diag_padding` if this improves the code.

andlrc: I think it does, as it's the only case `diag_padding` is used is for `after` the two other cases computes the `padding` based on the visual size of the start character. Maybe even move that padding calculation to the else branch:

```
if diag_align == 'after'
  padding = 3
  ...
else
  padding = diag.range.start.character
  if padding > 0
    ...
  endif
endif

File: autoload/lsp/diag.vim [hunk 3]
```diff
@@ -162,14 +193,24 @@ def DiagsRefresh(bnr: number)
...
       if opt.lspOptions.showDiagWithVirtualText
+
         var padding = diag.range.start.character
         if padding > 0
           padding = strdisplaywidth(getline(diag.range.start.line + 1)[ : diag.range.start.character - 1])
         endif
+
+        var symbol = diag_symbol
+
+        if diag_align ==? 'after'
```
andlrc, line 204: Here `diag_align` is set from within the source code, so there cannot be any need for `==?` 😅

autoload/lsp/diag.vim
berggeist: I'm the newbie here in this code base and have to adhere to the legacies  :smile:
Therefore I leave the decision to you and @yegappan  :smile:

Shane-XB-Qian:
LGTM

// but they may prefer defined the init value to a dict, but not a matter, up to youself.

autoload/lsp/diag.vim
yegappan: The option value comparison should be case sensitive.  Can you use `==#` instead of `==?`?

Shane-XB-Qian: > I'm the newbie here in this code base and have to adhere to the legacies smile Therefore I leave the decision to you and @yegappan smile

as you said, it had no drawbacks, 
but just need to respect/keep the code style.
// which normally should be a common sense in the most open-source in the world, 

HOWEVER this is not the way here in this code base, @yegappan can break/revert their own created style/rule in anytime, and amazedly preferred to listen to someone no sense no why.......
but good news this is a open source, you can modify your copy/fork anytime, unless you want/like/thought it is worthy to share it with someone else then a PR.

PR #284, opened 2023-04-21
add words from current buffer to completion list

greeschenko:


andlrc:
I think this should be behind a flag, and I don't this this should affect people using the completion via Omni completion

autoload/lsp/completion.vim
```diff
@@ -170,6 +172,23 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     CompletionUltiSnips(prefix, items)
   endif
 
+  # add completion from current buf
+  var words = {}
+  var text = join(getline(1, '$'), "\n")
```
yegappan, line 177: It will be very expensive to process all the words in a buffer for every completion.  In particular, if the file is very large., it might take a quite a white.  Why can't the Vim built-in insert-mode completion mechanism (CTRL-N and CTRL-P) be used for this?

greeschenko: yes, we can always use C-x C-i or C-n to complete it, but it won't be in the autocomplete list, requiring additional user action

relatively to large list maybe I can exclude all the elements that are already in LSP server items or add to list only specific word such us strings , numbers, tag attributes etc.

but I don`t have any issues with 2-4k line files

greeschenko: and if we have another completion items C-n don`t work because  we just jump to next completion item 
only C-x C-i work in this case

greeschenko: another plus for me is use it with completionMatcher: "fuzzy" if I work with big files without LSP support (json config, docker-compose etc.). I can use C-n but in this case I spend more time for needed word search

greeschenko:
> I think this should be behind a flag, and I don't this this should affect people using the completion via Omni completion

make optional in https://github.com/yegappan/lsp/pull/284/commits/f2c8255cd0bb7a3be344b8341256f36f88f757dc

techntools:
I am including matches from current buffer in my plugin [completion-buzz](https://github.com/techntools/completion-buzz) 

I used function from [vim-CompleteHelper](https://github.com/inkarkat/vim-CompleteHelper) in the beginning but removed dependency on it by including library function in the plugin itself and removing code not needed.

That library functtion is fast. I do not get lag for more than 15000 lines of code for current buffer match.

But I am including only matches rather than collecting every word.

greeschenko:
> I am including matches from current buffer in my plugin [completion-buzz](https://github.com/techntools/completion-buzz)
> 
> I used function from [vim-CompleteHelper](https://github.com/inkarkat/vim-CompleteHelper) in the beginning but removed dependency on it by including library function in the plugin itself and removing code not needed.
> 
> That library functtion is fast. I do not get lag for more than 15000 lines of code for current buffer match.
> 
> But I am including only matches rather than collecting every word.

thanks, I'm currently looking for acceleration options and will definitely consider the solution from your plugin

autoload/lsp/completion.vim
```diff
@@ -170,6 +172,25 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     CompletionUltiSnips(prefix, items)
   endif
 
+  if opt.lspOptions.useBufferCompletion
+      # add completion from current buf
```
yegappan, line 176: Can you move the code inside this if statement into a separate function?

autoload/lsp/completion.vim
```diff
@@ -170,6 +172,25 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
...
+      var words = {}
+      var text = join(getline(1, '$'), "\n")
```
yegappan, line 178: Can you use the method calling syntax?

autoload/lsp/completion.vim
```diff
@@ -170,6 +172,25 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
...
+      for word in split(text, '\W\+')
```
yegappan, line 179: Instead of getting all the lines, joining them, and then splitting the words, it might be efficient to directly process the lines one at a time:
```
  for line in getline(1, '$')
    for word in line->split('\W\+')
```

doc/lsp.txt
```diff
@@ -498,6 +498,10 @@ usePopupInCodeAction    |Boolean| option.  When using the |:LspCodeAction|
 			command to display the code action for the current
 			line, use a popup menu instead of echoing.
 			By default this is set to false.
+						*lsp-opt-useBufferCompletion*
+useBufferCompletion     |Boolean| option. If enable adds current buffer words
+			to autocompletion list. Мay degrade vim performance.
+			By default this is set to false.
```
yegappan, line 504: Can you change the help text to:

If enabled, the words from the current buffer are added to the auto completion list.  This may degrade Vim performance as the current buffer contents are processed every time the completion menu is displayed.

autoload/lsp/completion.vim
```diff
@@ -170,6 +172,25 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
...
+          if !has_key(words, word) && len(word) > 1
```
yegappan, line 180: Note that numbers are also treated as word here.

yegappan (changes requested):

autoload/lsp/completion.vim
greeschenko: add fix in https://github.com/yegappan/lsp/pull/284/commits/b48c94b90e9da99d6fa271f0eba8400cb5d11d48

autoload/lsp/completion.vim
greeschenko: added fix in https://github.com/yegappan/lsp/pull/284/commits/b48c94b90e9da99d6fa271f0eba8400cb5d11d48

doc/lsp.txt
greeschenko: added fix in https://github.com/yegappan/lsp/pull/284/commits/b48c94b90e9da99d6fa271f0eba8400cb5d11d48

autoload/lsp/completion.vim
```diff
@@ -135,6 +137,26 @@ def CompletionUltiSnips(prefix: string, items: list<dict<any>>)
   endfor
 enddef
 
+# add completion from current buf
+def CompletionFromBuffer(items: list<dict<any>>)
+    var words = {}
+    for line in getline(1, '$')
+        for word in line->split('\W\+')
+            if !has_key(words, word) && len(word) > 1
```
yegappan, line 145: Can you change this to `if !words->has_key(word) && word->len() > 1`?

yegappan (changes requested):

autoload/lsp/completion.vim
greeschenko: done in https://github.com/yegappan/lsp/pull/284/commits/a7b2ba9f3f7f9d356f4cca8f38b834a62790444d

PR #297, opened 2023-05-08
Adapt LspFileTorUri for Cygwin

inetic:
Language servers are compiled to use the standard Windows paths (although only tested with `rust-analyzer`). But `vim` under `Cygwin` uses POSIX paths. Therefore the paths have to be converted somewhere and `LspFileToUri` seems like the ideal place for it.

The added code detects if the script is being run from `Cygwin` and if so, it convert paths of the form "/cygdrive/c/foo/bar" to "c:/foo/bar" and paths of the form "/home/pete/foo" to "C:/cygwin64/home/pete/foo".

Feel free to suggest changes as this is my first non-one-liner vim script.

### DIFF
--- a/autoload/lsp/util.vim
+++ b/autoload/lsp/util.vim
@@ -111,6 +111,11 @@ enddef
 export def LspFileToUri(fname: string): string
   var uri: string = fname->fnamemodify(':p')
 
+  if has("win32unix")
+    # We're in Cygwin
+    uri = CygwinToWindowsPath(uri)
+  endif
+
   var on_windows: bool = false
   if uri =~? '^\a:'
     on_windows = true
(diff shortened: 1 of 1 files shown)


autoload/lsp/util.vim
```diff
@@ -133,6 +138,27 @@ export def LspFileToUri(fname: string): string
   return uri
 enddef
 
+# Convert POSIX paths as used in Cygwin to native Windows paths
+def CygwinToWindowsPath(path: string): string
+  if path =~? '^\/cygdrive\/'
+    # Convert paths of the form "/cygdrive/c/foo/bar" to "C:/foo/bar"
+    return substitute(path, '^\/cygdrive\/\(\a\)\/', '\=submatch(1) .. ":/"', "")
```
yegappan, line 145: This plugin uses the method calling syntax.  Can you change the above to:
```
return path->substitute(....)
```

autoload/lsp/util.vim
```diff
@@ -133,6 +138,27 @@ export def LspFileToUri(fname: string): string
...
+  elseif path =~? '^\/'
+    # Convert paths of the form "/home/pete/foo" to "C:/cygwin64/home/pete/foo"
+    if len(g:cygwinroot) == 0
```
yegappan, line 148: Change this to `g:cygwinroot->len()`

autoload/lsp/util.vim
```diff
@@ -133,6 +138,27 @@ export def LspFileToUri(fname: string): string
...
+      # https://stackoverflow.com/a/7449029/273348
+      g:cygwinroot = substitute(system("reg query HKEY_LOCAL_MACHINE\\\\SOFTWARE\\\\Cygwin\\\\setup /v rootdir | grep rootdir"),
```
yegappan, line 150: For readability, can you assign the command to a local variable and then use it in the system() call?

yegappan (changes requested):

autoload/lsp/util.vim
```diff
@@ -133,6 +138,31 @@ export def LspFileToUri(fname: string): string
   return uri
 enddef
 
+# Convert POSIX paths as used in Cygwin to native Windows paths
+def CygwinToWindowsPath(path: string): string
```
Shane-XB-Qian, line 142: there should be direct way to convert convert such path in cygwin, if i didnot recall wrong.
// so most code of this func perhaps is not required.

// and other than this, not sure lsp server returned path was window or unix style path,
// somewhere perhaps should convert it oppositely.

inetic: > there should be direct way to convert convert such path in cygwin, if i didnot recall wrong.

Hm, you seem to be right, there is [`cygpath`](https://cygwin.com/cygwin-ug-net/cygpath.html). I'll try to replace my code with it.

yegappan: I have merged this PR.  Can you create a new PR with the change?

PR #298, opened 2023-05-11
Simplify cygwin path conversions

inetic:
As suggested by @Shane-XB-Qian [here](https://github.com/yegappan/lsp/pull/297/files#r1188445597), the `cygwin` path conversion code can be simplified by using `cygpath`.

### DIFF
--- a/autoload/lsp/util.vim
+++ b/autoload/lsp/util.vim
@@ -112,9 +112,9 @@ export def LspFileToUri(fname: string): string
   var uri: string = fname->fnamemodify(':p')
 
   if has("win32unix")
-    # We're in Cygwin, we need to convert POSIX style paths to Windows style.
-    # `cygpath -m` converts paths of the form "/cygdrive/c/foo/bar" to "C:/foo/bar", and
-    # paths of the form "/home/pete/foo" to "C:/cygwin64/home/pete/foo".
+    # We're in Cygwin, convert POSIX style paths to Windows style.
+    # The substitution is to remove the '^@' escape character from the end of
+    # line.
     uri = system($'cygpath -m {uri}')->substitute('^\(\p*\).*$', '\=submatch(1)', "")
   endif
 
--- a/autoload/lsp/util.vim
+++ b/autoload/lsp/util.vim
@@ -112,8 +112,10 @@ export def LspFileToUri(fname: string): string
   var uri: string = fname->fnamemodify(':p')
 
   if has("win32unix")
-    # We're in Cygwin
-    uri = CygwinToWindowsPath(uri)
+    # We're in Cygwin, we need to convert POSIX style paths to Windows style.
+    # `cygpath -m` converts paths of the form "/cygdrive/c/foo/bar" to "C:/foo/bar", and
+    # paths of the form "/home/pete/foo" to "C:/cygwin64/home/pete/foo".
+    uri = system($'cygpath -m {uri}')->substitute('^\(\p*\).*$', '\=submatch(1)', "")
   endif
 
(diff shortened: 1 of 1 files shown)


autoload/lsp/util.vim
```diff
@@ -112,8 +112,10 @@ export def LspFileToUri(fname: string): string
   var uri: string = fname->fnamemodify(':p')
 
   if has("win32unix")
-    # We're in Cygwin
-    uri = CygwinToWindowsPath(uri)
+    # We're in Cygwin, we need to convert POSIX style paths to Windows style.
+    # `cygpath -m` converts paths of the form "/cygdrive/c/foo/bar" to "C:/foo/bar", and
```
Shane-XB-Qian, line 116: not sure the comment was accurate or not, since i recall not all possible it was /cygdrive/ header, it was not always was cygwin64/, but if just an example, then ok.

autoload/lsp/util.vim
```diff
@@ -112,8 +112,10 @@ export def LspFileToUri(fname: string): string
...
+    # paths of the form "/home/pete/foo" to "C:/cygwin64/home/pete/foo".
+    uri = system($'cygpath -m {uri}')->substitute('^\(\p*\).*$', '\=submatch(1)', "")
```
Shane-XB-Qian, line 118: this substitute seems not very efficient or necessary?

Shane-XB-Qian (commented):
i have no Windows pc anymore, please try to self-verify a bit if all worked ok.

autoload/lsp/util.vim
inetic: The substitution is to remove end of line escape characters. Quick internet searching did not show me a better alternative.

autoload/lsp/util.vim
inetic: > i have no Windows pc anymore, please try to self-verify a bit if all worked ok.

I did.

> not sure the comment was accurate or not, since i recall not all possible it was /cygdrive/ header, it was not always was cygwin64/, but if just an example, then ok.

I'm not sure I understand, can you please rephrase? Quick test in bash under cygwin:

```
$ cygpath -m /cygdrive/c/foo/bar
C:/foo/bar
$ cygpath -m /home/pete/foo/bar
C:/cygwin64/home/pete/foo/bar
```

If you mean that the path where cygwin is installed is not always `C:/cygwin64` then yes, that's true. I would hope in that case `cygpath` would return the actual correct path.

autoload/lsp/util.vim
Shane-XB-Qian: i think this perhaps can be refined a bit, but maybe not a matter.

autoload/lsp/util.vim
Shane-XB-Qian: if it is not sure, then perhaps can just remove those comments, 
since the command name supposed had already identify your code purpose.

Shane-XB-Qian: e.g just `convert POSIX style path to Windows style`

autoload/lsp/util.vim
inetic: Can you suggest how? As I mentioned in the other PR, I'm not a seasoned vim scripter and google search did not provide better alternatives.

Shane-XB-Qian: let it be, not a matter.

Shane-XB-Qian: let it be, not a matter.

Shane-XB-Qian:
@yegappan up to you, merge if you felt ok.

PR #301, opened 2023-05-14
Always switch to correct loclist after running `:LspShowDiags`

### LINKED ISSUE #300: Diagnostics error list is not switched to when it already exists
## Expected behaviour

Every time `:LspDiagShow` is run, the `Language Server Diagnostics` error list for that buffer is switched to.

## Actual behaviour

The `Language Server Diagnostics` error list is not switched to when it already exists and there is another error list in the loclist stack.

## Steps to reproduce:

```
:LspDiagShow
:lclose
<place cursor on symbol>
:LspShowReferences
:LspDiagShow
```

newtonne:
Fixes #300

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -404,6 +404,9 @@ def DiagsUpdateLocList(bnr: number): bool
     setbufvar(bnr, 'LspQfId', getloclist(0, {id: 0}).id)
   endif
 
+  var LspQfNr: number = getloclist(0, {id: LspQfId, nr: 0}).nr
+  silent! exe $':{LspQfNr} lhistory'
+
   return true
 enddef
 
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -421,7 +421,7 @@ export def ShowAllDiags(): void
     return
   endif
 
-  # make the error list the active one and open it
+  # make the diagnostics error list the active one and open it
   var LspQfId: number = getbufvar(bufnr(), 'LspQfId', 0)
   var LspQfNr: number = getloclist(0, {id: LspQfId, nr: 0}).nr
   exe $':{LspQfNr} lhistory'
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -410,12 +410,6 @@ def DiagsUpdateLocList(bnr: number): bool
     setbufvar(bnr, 'LspQfId', getloclist(0, {id: 0}).id)
   endif
 
-  # if pre-existing error list, make current
-  if LspQfId != 0
-    var LspQfNr: number = getloclist(0, {id: LspQfId, nr: 0}).nr
-    silent! exe $':{LspQfNr} lhistory'
-  endif
-
   return true
 enddef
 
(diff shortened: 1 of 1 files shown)


autoload/lsp/diag.vim
```diff
@@ -404,6 +404,9 @@ def DiagsUpdateLocList(bnr: number): bool
     setbufvar(bnr, 'LspQfId', getloclist(0, {id: 0}).id)
   endif
 
+  var LspQfNr: number = getloclist(0, {id: LspQfId, nr: 0}).nr
```
Shane-XB-Qian, line 407: is it correct? what if lspqfid is 0? or what if current is the right list?

andlrc: I'm also a bit skeptical for whether or not this will work for all cases 🤔

newtonne: Yeah, I believe it does work because `LspQfId` is either:

- The id of the existing error list if one previously existed
- 0 (equivalent to current list in `getloclist` `id` parameter) if this is a new error list

In the first case, it will make the existing list current. In the second case, it will make the new list the current list again since new lists are already made current when they are created. This is fine but unnecessary. I'll make it clearer.

autoload/lsp/diag.vim
```diff
@@ -404,6 +404,12 @@ def DiagsUpdateLocList(bnr: number): bool
     setbufvar(bnr, 'LspQfId', getloclist(0, {id: 0}).id)
   endif
 
+  # if pre-existing error list, make current
```
Shane-XB-Qian, line 407: previously existed? :rofl:

Shane-XB-Qian (commented):
i knew that possibly qf list was changed, but your code i just felt somehow incorrect.
perhaps should just switch it for it was really changed.

autoload/lsp/diag.vim
andlrc: Cool, thanks for elaborating. What will happen if the history is exhausted by a lot of `:lgrep` etc commands by the user? AFAIK the history only contains the last 10 location lists.

newtonne: That case is handled here: https://github.com/yegappan/lsp/blob/ce8f74885cd32d60e761bb979966e2728bd8eb6c/autoload/lsp/diag.vim#L368-L371

Since the error list has fallen out of the stack, it is created anew, which automatically makes it the current error list.

andlrc:
I think this PR should be merged

autoload/lsp/diag.vim
```diff
@@ -404,6 +404,12 @@ def DiagsUpdateLocList(bnr: number): bool
...
+  if LspQfId != 0
+    var LspQfNr: number = getloclist(0, {id: LspQfId, nr: 0}).nr
+    silent! exe $':{LspQfNr} lhistory'
```
yegappan, line 410: This code should be moved to the ShowAllDiags() function.  Otherwise whenever new diagnostics is received from the language server, the current location list will be changed.  If the user is using some other plugin (e.g. grep) to populate the location list, then this will lead to unexpected behavior.  In the ShowAllDiags() function, you can get the 'LspQfId' buffer-local variable for the current buffer and make that the current location list.

newtonne: Nice spot - thanks. I've made the change now.

PR #302, opened 2023-05-14
Add `keepFocusInDiags` option

newtonne:
This option serves a similar purpose to `keepFocusInReferences` but for `LspDiagShow` rather than `LspShowReferences`. Note that the default behaviour is changed by this PR as  `keepFocusInDiags` defaults to false.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -410,13 +410,12 @@ enddef
 # Display the diagnostic messages from the LSP server for the current buffer
 # in a location list
 export def ShowAllDiags(): void
-  var save_winid = win_getid()
-
   if !DiagsUpdateLocList(bufnr())
     util.WarnMsg($'No diagnostic messages found for {@%}')
     return
   endif
 
+  var save_winid = win_getid()
   :lopen
   if !opt.lspOptions.keepFocusInDiags
     save_winid->win_gotoid()
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -410,12 +410,17 @@ enddef
 # Display the diagnostic messages from the LSP server for the current buffer
 # in a location list
 export def ShowAllDiags(): void
+  var save_winid = win_getid()
+
   if !DiagsUpdateLocList(bufnr())
     util.WarnMsg($'No diagnostic messages found for {@%}')
     return
   endif
 
   :lopen
+  if !opt.lspOptions.keepFocusInDiags
+    save_winid->win_gotoid()
+  endif
 enddef
 
 # Display the message of 'diag' in a popup window right below the position in
--- a/test/clangd_tests.vim
+++ b/test/clangd_tests.vim
@@ -185,7 +185,7 @@ def g:Test_LspShowReferences()
   assert_equal([4, 6], [loclist[0].lnum, loclist[0].col])
   assert_equal([5, 2], [loclist[1].lnum, loclist[1].col])
   assert_equal([6, 6], [loclist[2].lnum, loclist[2].col])
-  :only
+  :lclose
   cursor(1, 5)
   :LspShowReferences
   assert_equal(1, getloclist(0)->len())
@@ -199,13 +199,13 @@ def g:Test_LspShowReferences()
   :LspShowReferences
   sleep 100m
   assert_equal('quickfix', getwinvar(winnr('$'), '&buftype'))
+  :cclose
   var qfl: list<dict<any>> = getqflist()
   assert_equal(3, qfl->len())
   assert_equal(bufnr(), qfl[0].bufnr)
   assert_equal([4, 6], [qfl[0].lnum, qfl[0].col])
   assert_equal([5, 2], [qfl[1].lnum, qfl[1].col])
   assert_equal([6, 6], [qfl[2].lnum, qfl[2].col])
-  :only
   cursor(1, 5)
   :LspShowReferences
   assert_equal(1, getqflist()->len())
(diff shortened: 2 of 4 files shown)


andlrc:
Is it possible to add a test that confirms this new behavior?

autoload/lsp/diag.vim
```diff
@@ -410,12 +410,17 @@ enddef
 # Display the diagnostic messages from the LSP server for the current buffer
 # in a location list
 export def ShowAllDiags(): void
+  var save_winid = win_getid()
```
Shane-XB-Qian, line 413: should move to/after `if !DiagsUpdateLocList(bufnr())`

Shane-XB-Qian (commented):
but mostly that's a default behavior which kept focus at qf/loc, not sure if it's worthy to add such option to every qf copen/lopen.

autoload/lsp/diag.vim
newtonne: Done 👍

andlrc:
It's a bit unfortunate that keepFocusInReferences and this new option keepFocusInDiags can't have the same default value.  

I would actually vote for breaking backward compatibility as to keep this two options have the same default value. 

I think the default values for both should be `false`.

newtonne:
I agree @andlrc . Will make the change.

test/clangd_tests.vim
```diff
@@ -380,7 +394,7 @@ def g:Test_LspDiag_Multi()
   endif
   assert_equal([1, 9, 'E'], [qfl[1].lnum, qfl[1].col, qfl[1].type])
   assert_equal([2, 9, 'E'], [qfl[2].lnum, qfl[2].col, qfl[2].type])
-  close
+  :lclose
```
newtonne, line 397: Need to use `:lclose` now since we're no longer focussed in the loclist.

test/clangd_tests.vim
```diff
@@ -263,7 +270,7 @@ def g:Test_LspDiag()
   assert_equal([3, 14, 'E'], [qfl[0].lnum, qfl[0].col, qfl[0].type])
   assert_equal([5, 2, 'W'], [qfl[1].lnum, qfl[1].col, qfl[1].type])
   assert_equal([7, 2, 'W'], [qfl[2].lnum, qfl[2].col, qfl[2].type])
-  close
+  :lclose
```
newtonne, line 273: Need to use `:lclose` now since we're no longer focussed in the loclist.

Shane-XB-Qian:
i disagree.
the default should not break most default behavior.
that keepFocusInReferences existed just because it was a async command, but normally it should be no needed such option.

newtonne:
It doesn't seem right to me that the default behaviour for `LspDiagShow` and `LspShowReferences` is different, which it currently is. This PR, as it stands, would fix that inconsistency, admittedly at the cost of changing existing behaviour. However, I don't feel like the change would be that disruptive in the grand scheme of things and the old behaviour would be a config change away.

andlrc:
TBH I don't really care which behavior is the default, but I think that we should strive to keep  `false`-values as defaults, and we should also strive to align these two options. So either:

    keepFocusInReferences = false
    keepFocusInDiags = false
    
Or

    keepBufferFocusWhenShowingRefs = false # this name stinks, but is trying to be "!keepFocusInReferences"
    keepBufferFocusWhenShowingDiags = false # this name stinks, but is trying to be "!keepFocusInDiags"

andlrc:
I think this PR should be merged, but keep in mind that it isn't backward compatible, even though the description states that.

newtonne:
Updated the description to note that there is a change to the default behaviour.

yegappan:
I think we should change the default behavior to move the focus to the location list window (set these two option default values to true).  When a user runs the "LspShowDiags" command, he most likely will want to jump to a particular diagnostic by selecting it in the location list window.  If the default behavior is not to move the focus to the location list window, then the user needs to take an additional step to jump to the location list window before selecting and jumping to a diagnostic.  Same applies for symbol references.  So it makes sense to change the default behavior to move the focus to the location list window.

newtonne:
No problem. Updated them both to default to true.

newtonne:
I couldn't get these golang tests to fail in the same way locally so would be grateful if  someone can point me in the right direction or feel free to fix it yourself.

andlrc:
> I couldn't get these golang tests to fail in the same way locally so would be grateful if someone can point me in the right direction or feel free to fix it yourself.

I think it's because the focus is not set in the loclist and qflist respectively, making the call to `bufnr()` returning a bufnr for those windows, instead of the buffer edited. One way around it is to store the bufnr running `:LspGotoImpl`:

```
diff --git a/test/gopls_tests.vim b/test/gopls_tests.vim
index 1b06e1b..bf00c89 100644
--- a/test/gopls_tests.vim
+++ b/test/gopls_tests.vim
@@ -16,6 +16,8 @@ echomsg systemlist($'{lspServers[0].path} version')
 # shown in a list or popup
 def g:Test_LspGoto()
   :silent! edit Xtest.go
+  var bnr = bufnr()
+
   sleep 200m
 
   var lines =<< trim END
@@ -62,7 +64,7 @@ def g:Test_LspGoto()
   var loclist: list<dict<any>> = getloclist(0)
   assert_equal('quickfix', getwinvar(winnr('$'), '&buftype'))
   assert_equal(2, loclist->len())
-  assert_equal(bufnr(), loclist[0].bufnr)
+  assert_equal(bnr, loclist[0].bufnr)
   assert_equal([9, 13, ''], [loclist[0].lnum, loclist[0].col, loclist[0].type])
   assert_equal([13, 13, ''], [loclist[1].lnum, loclist[1].col, loclist[1].type])
   lclose
@@ -75,7 +77,7 @@ def g:Test_LspGoto()
   var qfl: list<dict<any>> = getqflist()
   assert_equal('quickfix', getwinvar(winnr('$'), '&buftype'))
   assert_equal(2, qfl->len())
-  assert_equal(bufnr(), qfl[0].bufnr)
+  assert_equal(bnr, qfl[0].bufnr)
   assert_equal([9, 13, ''], [qfl[0].lnum, qfl[0].col, qfl[0].type])
   assert_equal([13, 13, ''], [qfl[1].lnum, qfl[1].col, qfl[1].type])
   cclose
```

yegappan:
@andlrc Can you create a PR with your suggested change to gopls_tests.vim?

andlrc:
@yegappan Sure.

newtonne:
Thanks for this 👍

PR #306, opened 2023-05-16
Custom ref highlight

andlrc:
Make it possible to define custom highlights for ref created via `:LspHighlight`.

Also prevent these highlights from overriding visual selection, hlsearch etc.

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -31,9 +31,13 @@ var ftypeServerMap: dict<list<dict<any>>> = {}
 var lspInitializedOnce = false
 
 def LspInitOnce()
-  prop_type_add('LspTextRef', {highlight: 'Search', override: true})
-  prop_type_add('LspReadRef', {highlight: 'DiffChange', override: true})
-  prop_type_add('LspWriteRef', {highlight: 'DiffDelete', override: true})
+  hlset([{name: 'LspTextRef', default: true, linksto: 'Search'}])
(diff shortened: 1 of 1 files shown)


autoload/lsp/lsp.vim
```diff
@@ -31,9 +31,13 @@ var ftypeServerMap: dict<list<dict<any>>> = {}
 var lspInitializedOnce = false
 
 def LspInitOnce()
-  prop_type_add('LspTextRef', {highlight: 'Search', override: true})
```
Shane-XB-Qian, line 34: could you keep 'override: true' ? though you had done it for 'LspDiagInlineError' etc,
but actually someone perhaps disagree with it, which preferred to had 'override: true' to work with 'cursorline'.

andlrc: It's unfortunate that vim doesn't allow us to override only the cursorline. It might make sense to add another option "propOverride", which when set will set the override flag on all props.

yegappan: If the 'override' attribute is not set for these text properties and the user has set the 'cursorline' option, then the references in the current line will not be highlighted or shown.  Won't that confuse the user?

andlrc: Possible, the alternative is that visual selection / search etc, will highlighted below the prop. So it's either show prop on top of everything, or show cursorline, visual selection, and search above the prop.

Shane-XB-Qian:
>  Won't that confuse the user?
yes, but this's been merged, so the answer is cursorline was **less** important?

PR #307, opened 2023-05-16
Add option "showDiagSource"

andlrc:
This will add an option to prefix each diagnostic message with the name of the language server from which the diagnostic is coming from.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -272,6 +272,9 @@ export def DiagNotification(lspserver: dict<any>, uri: string, newDiags: list<di
 
   var diagWithinRange: list<dict<any>> = []
   for diag in newDiags
+    if opt.lspOptions.showDiagSource
+      diag.message = $'{lspserver.name}: {diag.message}'
+    endif
     if diag.range.start.line + 1 > lastlnum
       # Make sure the line number is a valid buffer line number
       diag.range.start.line = lastlnum - 1
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -53,6 +53,9 @@ export var lspOptions: dict<any> = {
   # Suppress diagnostic hover from appearing when the mouse is over the line
   # Show a diagnostic message on a status line
   showDiagOnStatusLine: false,
+  # Show which language server a diagnostic is comming from, by prefixing the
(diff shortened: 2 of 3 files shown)


autoload/lsp/diag.vim
```diff
@@ -272,6 +272,9 @@ export def DiagNotification(lspserver: dict<any>, uri: string, newDiags: list<di
 
   var diagWithinRange: list<dict<any>> = []
   for diag in newDiags
+    if opt.lspOptions.showDiagSource
```
Shane-XB-Qian, line 275: you can make it just when/if it had multiple-server for that diag, then show source, otherwise just keep without that.
// if so no additional 'option' required
// , otherwise looks 'option' became too many.

andlrc: I don't see any problem is having another option, as long as it's well documented. If the source is automatically prefixed for multiple servers, then someone will request an option to remove the prefix.

Tachi107: > I don't see any problem is having another option, as long as it's well documented

I disagree. Having too many option has accessibility/usability drawbacks, and they are too often used as a mechanism to shift a responsibility to the user which the software writers could absorb instead by avoiding adding the option.

This doesn't apply to this patch in particular, it is more of a general rule; options are not "free", and should only be added if necessary.

With regards to this patch, I agree with @Shane-XB-Qian, and I would first add an automatic prefix mechanism, and I'd consider adding an option only if somebody actually asks for it.

andlrc: >  I'd consider adding an option only if somebody actually asks for it.

I would ask for it immediately, as I only want to prefix with the language server name in some cases.

berggeist: May I ask for a more expressive/meaningful name for this new option? 
As a user, for me it is not clear what is meant by```source``` in this context; is it the source-code that includes issues that raises the diagnostic messages?

By keeping to the description of this option

> prefix each diagnostic message with the name of the language
> server

something like ```pefixDiagMsgWithServername``` or ```prependServerToDiagMsg``` would be, at least in my opinion, far more expressive.

andlrc:
Closed as I like the flexibility of https://github.com/yegappan/lsp/pull/318  better

Issue #294, opened 2023-05-03
How to hide "Error: Unsupported notification message" for jdtls


maximyurevich:

yegappan:
Can you include the complete error message (including the notification message)?

maximyurevich:
> Can you include the complete error message (including the notification message)?

Error: Unsupported request message received from the LSP server (jdtls), message = {'id': '2', 'jsonrpc': '2.0', 'method': 'workspace/executeClientCommand', 'params': {'arguments': [], 'command': '_java.reloadBundles.command'}}

andlrc:
I'm confused as to if this is related to notifications or request replies in general? As the issue titles mentions notifications, while the copied error is related to the later.

andlrc:
FWIW then you can "silence" notifications by `:h lsp-cfg-customNotificationHandlers`.

Tachi107:
This might be somewhat related to https://github.com/yegappan/lsp/issues/125

girishji:
Should be fixed by #309

PR #309, opened 2023-05-17
Ignore non-standard request from eclipse java LS

girishji:
Eclipse java language server sends the 'workspace/executeClientCommand' request (to reload bundles) which is not in the LSP specification. Ignore this request and suppress the error message.

Fixed typos in lspserver.vim

M  autoload/lsp/handlers.vim
M  autoload/lsp/lspserver.vim

### DIFF
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -200,9 +200,17 @@ export def ProcessRequest(lspserver: dict<any>, request: dict<any>)
       'workspace/semanticTokens/refresh': ProcessUnsupportedReq
     }
 
+  # Explicitly ignored requests
+  var lspIgnoredRequestHandlers: list<string> =
+    [
+      # Eclipse java language server sends the 'workspace/executeClientCommand' 
+      # request (to reload bundles) which is not in the LSP specification.
+      'workspace/executeClientCommand',
+    ]
+
   if lspRequestHandlers->has_key(request.method)
     lspRequestHandlers[request.method](lspserver, request)
-  else
+  elseif lspIgnoredRequestHandlers->index(request.method) == -1
     util.ErrMsg($'Unsupported request message received from the LSP server ({lspserver.path}), message = {request->string()}')
   endif
 enddef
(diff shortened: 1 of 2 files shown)


andlrc:
I think that we should consider adding the same kind of support that have been added for notifications. I actually have a branch that adds that support somewhere.

Shane-XB-Qian:
https://github.com/yegappan/lsp/issues/294

andlrc:
I have now turned the branch that I talked about into a PR: https://github.com/yegappan/lsp/pull/310

girishji:
> I think that we should consider adding the same kind of support that have been added for notifications. I actually have a branch that adds that support somewhere.

This was my first thought. I am glad you have already worked on it. Otherwise this would've been my second PR. 

We also need to be able to ignore some of these non-standard requests though.

autoload/lsp/handlers.vim
```diff
@@ -200,9 +200,17 @@ export def ProcessRequest(lspserver: dict<any>, request: dict<any>)
       'workspace/semanticTokens/refresh': ProcessUnsupportedReq
     }
 
+  # Explicitly ignored requests
+  var lspIgnoredRequestHandlers: list<string> =
+    [
+      # Eclipse java language server sends the 'workspace/executeClientCommand' 
+      # request (to reload bundles) which is not in the LSP specification.
+      'workspace/executeClientCommand',
```
Shane-XB-Qian, line 208: not sure if you'd researched `workspace/executeClientCommand` what's that for?
some req msg though not a LSP specification, but that's useful, 
e.g `textDocument/switchSourceHeader` is only for clangd, but we had it.

// but we do truly require some code like this to ignore req msg IF/which it was really not useful.

girishji: I think `workspace/executeClientCommand` is for Eclipse IDE to reload some bundle with eclipse related packages. Document is scarce. If it turns out to be important then we can provide a callback through #310

andlrc:
> We also need to be able to ignore some of these non-standard requests though.

I think that's fair. As long as it's possible for the user to define them themselves, if needs be, if #310 is actually going to be merged.

PR #310, opened 2023-05-17
Add support for custom request handlers

andlrc:


girishji:
Is the API to send a request to the server exported? Many servers provide functionality not listed in LSP specs. User should be able to send a server-specific request and receive a response through callback and be able to do something with it.

girishji:
You could add a test. There is a test for `customNotificationHandlers` in test/clangd_tests.vim.

andlrc:
> You could add a test. There is a test for `customNotificationHandlers` in test/clangd_tests.vim.

I know, I wrote that test. I don't know any custom requests though.

I only posted this as a few have talked about some Java LSP sending unsupported requests. Feel free to adopt this PR and add tests / change it, if you feel something is missing, I'm not going to do much more to it.

autoload/lsp/handlers.vim
```diff
@@ -195,13 +191,13 @@ export def ProcessRequest(lspserver: dict<any>, request: dict<any>)
       'window/workDoneProgress/create': ProcessWorkDoneProgressCreate,
       'client/registerCapability': ProcessClientRegisterCap,
       'client/unregisterCapability': ProcessClientUnregisterCap,
-      'workspace/configuration': ProcessWorkspaceConfiguration,
-      'workspace/codeLens/refresh': ProcessUnsupportedReq,
-      'workspace/semanticTokens/refresh': ProcessUnsupportedReq
```
Shane-XB-Qian, line 200: i think you'd better keep this two request msg and this function, 
i guess yegappan kept it for backlog, perhaps they may impl one day.

andlrc: The code behaves the same, as it fall-though to the error message just below. If they are keept like that, then you wouldn't be able to implement your own handler.

andlrc: I don't think that I made myself clear, the code will behave exactly the same as if a handler isn't recognized the exact same error message, as the one printed in `ProcessUnsupportedReq` will be printed in line 202: https://github.com/yegappan/lsp/pull/310/files#diff-4a80aa048eaef05d2305e7a083faf0cf87c4b69a63e11faa37209fb3bb33ea72L206

Shane-XB-Qian: those two should be a remark, to indicate those two had not impl yet, clear?
at least i understood like that, or you can help impl those two request msg to common funcs now, then everything ok,, then you can have your customized funcs to those specific lang had specific msg.

Konfekt:
> I only posted this as a few have talked about some Java LSP sending unsupported requests. Feel free to adopt this PR and add tests / change it, if you feel something is missing, I'm not going to do much more to it.

Would you mind documenting how to use this? For example, according to https://github.com/puremourning/vimspector#java and https://urfoex.blogspot.com/2020/08/debugging-java-with-jdb-or-vim.html Java debugging requires to send a start request receiving a port

Issue #311, opened 2023-05-18
Error detected while compiling function


lucasgrvarela:
This error happens when opening a file for the first time. Tried with and without .vimrc and in both cases the error happened.
```                                                                                                                                                                        
Error detected while compiling function <SNR>25_NetrwBrowseChgDir[194]..<SNR>25_NetrwEditFile[7]..BufRead Autocommands for "*"..function lsp#lsp#AddFile[1]..lsp#buffer#BufHasLspServer[1]..lsp#buffer#BufLspServerGet:
line   16:
E1095: Unreachable code after :throw
Press ENTER or type command to continue
```

```
vim --version
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled May 17 2023 21:50:06)
Included patches: 1-1563
```

techntools:
Facing the same issue.

Try commenting return after ```:throw```

``` diff
diff --git a/autoload/lsp/buffer.vim b/autoload/lsp/buffer.vim
index c717e0d..bb67b5f 100644
--- a/autoload/lsp/buffer.vim
+++ b/autoload/lsp/buffer.vim
@@ -66,7 +66,7 @@ export def BufLspServerGet(bnr: number, feature: string = null_string): dict<any
   if !SupportedCheckFns->has_key(feature)
     # If this happns it is a programming error, and should be fixed in the source code
     :throw $'Error: ''{feature}'' is not a valid feature'
-    return {}
+    # return {}
   endif

   var SupportedCheckFn = SupportedCheckFns[feature]

```

Shane-XB-Qian:
the new recent vim9 patch restricted that, could you open a pr for that?
// or i believe yegappan would update it some days later.

yegappan:
The PR #314 is merged to address this issue.

PR #314, opened 2023-05-18
Remove return after throw

techntools:
Possible fix for #311


andlrc:
I think that this should be merged

PR #316, opened 2023-05-19
Examples to vim9, added index and some ascii artwork in docs

ubaldot:
The examples in the README were for the old vim, so I took the freedom to update them to vim9.
The same happened to the docs where I also found that there were a missing summary and I took the freedom to add a pinch of ascii art. 

I hope it won't bother! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ To register a LSP server, add the following lines to your .vimrc file (use only 
 vim9script
 
 # Clangd language server
-call LspAddServer([{
+LspAddServer([{
 	    name: 'clangd',
 	    filetype: ['c', 'cpp'],
 	    path: '/usr/local/bin/clangd',
@@ -63,7 +63,7 @@ call LspAddServer([{
 	  }])
 
 # Javascript/Typescript language server
-call LspAddServer([{
+LspAddServer([{
 	    name: 'typescriptlang',
 	    filetype: ['javascript', 'typescript'],
 	    path: '/usr/local/bin/typescript-language-server',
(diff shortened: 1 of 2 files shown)


girishji:
I think you don't need to quote dictionary keys in vim9. You may not need backslash line continuation as well.

ubaldot:
Right! Corrected. Please have a look. :)

ubaldot:
I also removed `call` for function calls as it is [is](https://vimhelp.org/motion.txt.html#is) discouraged in Vim9.

andlrc:
1. I think we should strive to keep most example snippets in legacy vimscript, as it seems that most people are still keeping their vimrc in that, which makes it easier to copy/paste snippets directly into your vimrc. At least examples in the README.md file.  
2. I guess the examples in the docs could be in vim9script.  
3. I don't like the ascii art, but that might just me me?
4. I don't think you should remove the license from the docs.
5. If we opt to ise vim9script examples everywhere, then `v:true` should be `true` and `v:false` -> `false`.

ubaldot:
1. Ah, ok. It is very stressed that you need vim9 that is why I thought to update everything but you actually have a good point because most likely many have their vimrc in the old format. 
2. Ok, that is updated, then. 
3. I don't know :D There will certainly be some who like and some who will not. But given that you are a major contributor I think we should do as you say (i.e. remove it)
4. Ok, I put it at the end because of a matter of practicality. For example, NERDTree has it at the end, but sure, we can put it back to the beginning. Shall that be before the summary? 
5. Correct. I never used the old vim script, sorry for having missed that! 

I am going to make the fixes right now...

andlrc:
4. I think that only @yegappan should move around the license :D

ubaldot:
4. ahaha good point! :D I placed where it was before.

andlrc:
I think this could be merged, if not, then I think that we should strive to keep all examples in the docs in legacy script syntax.

PR #318, opened 2023-05-20
Process diags

andlrc:
Add server configuration that is a function which can be used to manipulate diagnostics. The handler takes one argument which is the list of raw diagnostics and it expected to return a new list of diagnostics. 

I'm not really sure about the performance implications, if it's a big deal, then we could consider refactoring the handler to take each diagnostic individually and return a new one, returning `null` could work as a filter to remove the diagnostic. This would remove one pass though the diagnostics, but introduce many more function calls.


Tachi107:
> I'm not really sure about the performance implications

Does this feature have performance implications even if unused?

doc/lsp.txt
```diff
@@ -320,6 +320,44 @@ Additionally the following configurations can be made:
...
+
+		vim9script
+		g:LspAddServer([{
+			filetype: ['javascript', 'typescript'],
+			path: '/usr/local/bin/typescript-language-server',
+			args: ['--stdio'],
+			processDiagHandler: (diags: list<dict<any>>) => {
+				# Only include errors and warnings
+				return diags->filter((diag, ix) => {
+					return diag.severity <= 2
```
Shane-XB-Qian, line 340: technically maybe ok, but  i doubt is it really a right approach? 
i am feeling some kinds of risk which maybe used to manipulate sth if had such things... :rofl: 
// user maybe just copied cfg from somewhere/someone, but didnot understand what's that for, 
// a.k.a it maybe dangerous.

PR #319, opened 2023-05-20
WIP -Prioritize diags

andlrc:
DONT MERGE

This PR changes to a two pass approach over the diagnostics, as to avoid
complexity when computing the sign diagnostic:

For signs the diagnostic with the lowest (error is lowest) severity, in
a given line, is selected as being the severity to render in the sign
column.

For inline and virtual text diagnostics, the same is true, but here it
only happens if multiple diagnostics are at the same position (line +
character), this mostly happens if using multiple servers.

Avoid updating signs when the signcolumn is never shown

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -173,51 +173,99 @@ def DiagsRefresh(bnr: number)
 
   var signs: list<dict<any>> = []
   var diags: list<dict<any>> = diagsMap[bnr].sortedDiagnostics
-  for diag in diags
-    # TODO: prioritize most important severity if there are multiple diagnostics
-    # from the same line
-    var lnum = diag.range.start.line + 1
-    signs->add({id: 0, buffer: bnr, group: 'LSPDiag',
-				lnum: lnum,
-				name: DiagSevToSignName(diag.severity)})
+  var diaglen = diags->len()
+  var ix = 0
 
-    try
-      if opt.lspOptions.highlightDiagInline
-        prop_add(diag.range.start.line + 1,
-                  util.GetLineByteFromPos(bnr, diag.range.start) + 1,
-                  {end_lnum: diag.range.end.line + 1,
-                    end_col: util.GetLineByteFromPos(bnr, diag.range.end) + 1,
-                    bufnr: bnr,
-                    type: DiagSevToInlineHLName(diag.severity)})
+  # Place line based diagnostics
+  while ix < diaglen
+    var diag = diags[ix]
+
+    # Prioritize most important severity when there are multiple diagnostics in
+    # the same line
+    while ix + 1 < diaglen
+        && diag.range.start.line == diags[ix + 1].range.start.line
+
+      if diag.severity > diags[ix + 1].severity
+        diag = diags[ix + 1]
       endif
 
-      if opt.lspOptions.showDiagWithVirtualText
+      ix = ix + 1
+    endwhile
 
-        var padding: number
-        var symbol: string = diag_symbol
+    # increment for the iteration, place it as close to the top as possible,
+    # to avoid infinite loops, if "continue" is used.
+    ix = ix + 1
 
-        if diag_align == 'after'
-          padding = 3
-          symbol = DiagSevToSymbolText(diag.severity)
-        else
-          padding = diag.range.start.character
-          if padding > 0
-            padding = strdisplaywidth(getline(diag.range.start.line + 1)[ : diag.range.start.character - 1])
-          endif
+    var lnum = diag.range.start.line + 1
+    if signs->len() == 0 || signs[-1].lnum != lnum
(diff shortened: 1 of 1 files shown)


andlrc:
Need tests, the following might be a good test case for the sign selection:

```
#include <ctype.h>
#include <stdio.h>
void aFunc(void)
{
    printf("%s", z, tolower("aaa"));
}
```

As `z:5:18` is an error while `"aaa":5:29` is a warning.

autoload/lsp/diag.vim
```diff
@@ -171,55 +171,105 @@ def DiagsRefresh(bnr: number)
...
+          var padding: number
+          var symbol: string = diag_symbol
+
+          if diag_align == 'after'
+            padding = 3
+            symbol = DiagSevToSymbolText(diag.severity)
+          else
+            padding = diag.range.start.character
+            if padding > 0
+              padding = strdisplaywidth(getline(diag.range.start.line + 1))
```
berggeist, line 255: Seems that this line has been truncated or is incomplete here compared to the original line

```padding = strdisplaywidth(getline(diag.range.start.line + 1)[ : diag.range.start.character - 1])```

this part is lost from the end of the line:

```[ : diag.range.start.character - 1]```

File: autoload/lsp/diag.vim [hunk 2]
```diff
@@ -171,55 +171,105 @@ def DiagsRefresh(bnr: number)
...
+
+        ix = ix + 1
+      endwhile
+
+      # increment for the iteration, place it as close to the top as possible,
+      # to avoid infinite loops, if "continue" is used.
+      ix = ix + 1
+
+      var lnum = diag.range.start.line + 1
+      if signs->len() == 0 || signs[-1].lnum != lnum
```
berggeist, line 202: I do not understand this logic here ```signs[-1].lnum != lnum``` when could this not be true?

berggeist:
Hi @andlrc,
first my critical remarks (please don't feel offended, it's just my opinion :smiley:  ):

1. Looping twice (in the case ```signcolumn``` is enabled) through the complete list of diagnostics might slow down significantly the processing of diagnostics 
2. The additional complexity within ```DiagsRefresh``` might make the future maintenance of this code a kind of burden
3. If I do understand the logic concerning inline highlighting and showing virtual texts for diagnostic messages correctly, all but the diag. message with the highest severity will be dismissed. Thus information is lost. 

For 3.) In my opinion this is a bad idea, because different lsp- server might notify different helpful messages.

For 2.) maybe clever refactoring (extracting some logic into utility functions) might help

For 1.) just an idea: Wouldn't it be possible to skim just once through the  complete list of diagnostic messages and collect, maybe within an dict where the key is the line number, a structure that holds all needed information for further display? This dict can then be evaluated /processed for the individual GUI elements, like signs and highlighting.

autoload/lsp/diag.vim
andlrc: Thanks

autoload/lsp/diag.vim
andlrc: I think it was a relic from the attempt to have a one pass approach.

andlrc:
> 3. If I do understand the logic concerning inline highlighting and showing virtual texts for diagnostic messages correctly, all but the diag. message with the highest severity will be dismissed. Thus information is lost.

You have a point, with virtual text diagnostics. I as trying to fix the inline diagnostics, as currently if multiple diagnostics are at the same position, then multiple props are being added.  Same goes for the sign, multiple signs are being added for the same line. In both cases the last diagnostic wins, but that might not be the one with the most important severity.

> For 1.) just an idea: Wouldn't it be possible to skim just once through the complete list of diagnostic messages and collect, maybe within an dict where the key is the line number, a structure that holds all needed information for further display? This dict can then be evaluated /processed for the individual GUI elements, like signs and highlighting.

How would you make sure that you touch every element in your newly created dict? By iterating it? ;) Then you already had a two passes there. 

> And one question: How do you deal with diagnostics for overlapping regions with different severity, e.g. for highlighting.

We don't.

andlrc:
Come to think of it, another approach would be to, when sorting the diagnostics, to also sort on severity, so that error is sorted after warning. This would add the "correct" diagnostic at last for the sign, and for the inline highlight.

berggeist:
> How would you make sure that you touch every element in your newly created dict? By iterating it? ;) Then you already had a two passes there.

Yes of course, but with reduced number of contained items. And I was thinking ahead, too. By the use of such an additional structure it might be very easy in the future to add new features, like e.g. I just want to be notified about errors but no warnings, so I could implement a filter that additionally reduces the number of items within this structure.

berggeist:
All in all, it isn't my intention to block your efforts. But maybe it's worth to have a look how this is addressed in other lsp clients (if at all :smile:  ) But I don't have experience with other lsp-clients or editors :sob:

PR #322, opened 2023-05-23
Add propOverride which is forwarded to "prop_type_add"

andlrc:
This allows the user to make props override 'cursorline'


doc/lsp.txt
```diff
@@ -507,6 +507,10 @@ outlineOnRight		|Boolean| option.  Open the outline window on the
 						*lsp-opt-outlineWinSize*
 outlineWinSize		|Number| option.  The size of the symbol Outline
 			window.  By default this is set to 20.
+						*lsp-opt-propOverride*
+propOverride		|Boolean| option.  when TRUE textprops highlight
+			overrides any other, including 'cursorline' and Visual
+			By default this is set to false.
```
Shane-XB-Qian, line 513: if make it judge/work with `cursorlineopt` then no this option was required, it is 'ture' when no cursorline.
but simply it is ok, i think we can live with this option.

Shane-XB-Qian, line 513: if make it judge/work with `cursorlineopt` then no this option was required, it is 'ture' when no cursorline.
and wish it is default true, otherwise user maybe thought there is no diag (but actually it just been coverred).
but simply it is ok, i think we can live with this option.

andlrc: I made the default for `propOverride` equal:

```viml
&cursorline && &cursorlineopt =~ '\<line\>\|\<screenline\>\|\<both\>'
```

Shane-XB-Qian:
I mean we do not need this option if it behaved just rely on cursorlineopt, 

vs/but if you want this option existed, then wish this default is true, to be simple, just forget the cursorlineopt.

otherwise your modification made it more complex, since the cursorlineopt may changed in the middle.

doc/lsp.txt
Shane-XB-Qian: if you insisted this option required, then i think better be leave that cursorlineopt alone, just let user manager it via this option explicitly: which is directly and simply: 'true' or 'false'.

andlrc:
> I mean we do not need this option if it behaved just rely on cursorlineopt,
> 
> vs/but if you want this option existed, then wish this default is true, to be simple, just forget the cursorlineopt.

Having a sane default value makes sense. Being able to customize the value also makes sense. I think this is the best of both worlds.

> otherwise your modification made it more complex, since the cursorlineopt may changed in the middle.

It needs to change between the plugin is loaded, and the first file is edited, as this where the prop types are added.

doc/lsp.txt
yegappan: It will be simpler from a user perspective, if this option is set to either true or false (instead of a Boolean expression).  Can you move the check for the `cursorline` and `cursorlineopt` into the plugin code?

andlrc: Like this: https://github.com/yegappan/lsp/pull/326 ?

Issue #299, opened 2023-05-14
Cygwin: slow or reordered letters as I type


inetic:
Darn, sorry about this, but there is some regression with [this commit](https://github.com/yegappan/lsp/commit/acfd1408f2b076c7f883cb74aa2bf451219a78d8). It is that typing got a lot slower in a better case. In worse cases - as I type - I see some typed characters being reordered or duplicated. They stay as such until I switch buffers.

I think it's got to do with the fact that calling an external command makes the function slower. If that's the case than the problem was probably there even with the previous cygwin-path-change implementation in [this code path](https://github.com/yegappan/lsp/commit/acfd1408f2b076c7f883cb74aa2bf451219a78d8#diff-89400e0c68cc41fcdae808ce36f81438f12b87b4e17c2210b5980f9fd9b9fa15L151-L160) (as it also calls an external command).

I currently have a task in my work where I need to use Windows, so I can try to look into it, but would need some guidance or suggestions on what to try.

andlrc:
That's OK, these things happens when evolving software. I suspect that the same few files are send though `util.LspFileToUri` many many times, it might be worth adding a naive caching around the external commands, something like:

    var resolvedUris = {}
    export def LspFileToUri(fname: string): string
      if resolvedUris->has_key(fname)
        return resolvedUris[fname]
      endif
      
      ...
      ...

      resolvedUris[fname] = uri
      return uri

inetic:
This helped, thanks. Here is the [PR](https://github.com/yegappan/lsp/pull/328), let me know if you'd like to include something in it.

yegappan:
The PR #328 is merged.

PR #328, opened 2023-05-28
Speed up cygwin path transformation.

inetic:
Reported in issue #299

### DIFF
--- a/autoload/lsp/util.vim
+++ b/autoload/lsp/util.vim
@@ -117,15 +117,13 @@ export def LspFileToUri(fname: string): string
     return resolvedUris[fname_full]
   endif
 
-  var uri: string
+  var uri: string = fname_full
 
   if has("win32unix")
     # We're in Cygwin, convert POSIX style paths to Windows style.
     # The substitution is to remove the '^@' escape character from the end of
     # line.
-    uri = system($'cygpath -m {fname_full}')->substitute('^\(\p*\).*$', '\=submatch(1)', "")
-  else
-    uri = fname_full
+    uri = system($'cygpath -m {uri}')->substitute('^\(\p*\).*$', '\=submatch(1)', "")
   endif
 
   var on_windows: bool = false
--- a/autoload/lsp/util.vim
+++ b/autoload/lsp/util.vim
@@ -107,8 +107,14 @@ export def LspUriRemote(uri: string): bool
   return uri =~ '^\w\+::' || uri =~ '^[a-z][a-z0-9+.-]*://'
 enddef
 
+var resolvedUris = {}
+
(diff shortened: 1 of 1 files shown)


autoload/lsp/util.vim
```diff
@@ -107,8 +107,14 @@ export def LspUriRemote(uri: string): bool
...
 
+var resolvedUris = {}
+
 # Convert a Vim filename to an LSP URI (file://<absolute_path>)
 export def LspFileToUri(fname: string): string
+  if resolvedUris->has_key(fname)
+    return resolvedUris[fname]
+  endif
+
   var uri: string = fname->fnamemodify(':p')
```
Shane-XB-Qian, line 118: the key of dict perhaps should be this full_path, but just a feel.

andlrc: Good point.

inetic: Yeah, I think you're right. There is a potential issue if one would `cd` into a different directory with the same structure.

inetic:
I wonder if maybe another thing you may want here is to execute this caching only if we're running under cygwin?

autoload/lsp/util.vim
```diff
@@ -111,17 +111,21 @@ var resolvedUris = {}
...
-  if resolvedUris->has_key(fname)
-    return resolvedUris[fname]
+  var fname_full: string = fname->fnamemodify(':p')
+
+  if resolvedUris->has_key(fname_full)
+    return resolvedUris[fname_full]
   endif
 
-  var uri: string = fname->fnamemodify(':p')
+  var uri: string
```
andlrc, line 120: I would assign `fname_full` to `uri` here instead of in the else block, as it's so far away that it's hard to realize that `uri` have been assigned.

autoload/lsp/util.vim
```diff
@@ -111,17 +111,21 @@ var resolvedUris = {}
...
 
   if has("win32unix")
     # We're in Cygwin, convert POSIX style paths to Windows style.
     # The substitution is to remove the '^@' escape character from the end of
     # line.
-    uri = system($'cygpath -m {uri}')->substitute('^\(\p*\).*$', '\=submatch(1)', "")
+    uri = system($'cygpath -m {fname_full}')->substitute('^\(\p*\).*$', '\=submatch(1)', "")
```
andlrc, line 126: ... then use `uri` instead of `fname_full` here

autoload/lsp/util.vim
```diff
@@ -111,17 +111,21 @@ var resolvedUris = {}
...
+  else
+    uri = fname_full
```
andlrc, line 128: ... and remove this else clause.

PR #329, opened 2023-05-31
Skip invalid servers instead of aborting

vimpostor:
Right now with `ignoreMissingServer` set, when calling `LspAddServer()` with a list of servers where the very first entry is invalid, but the remaining ones are valid, no servers are added.

It seems highly unintuitive to me that the `ignoreMissingServer` option only silences the warning, but still aborts the whole call even if the remaining servers in the list are valid.

For example, I am calling:
```vim
au VimEnter * call LspAddServer([
	\ #{ name: 'nix', filetype: ['nix'], path: 'nil' },
	\ #{ name: 'bash', filetype: ['sh'], path: 'bash-language-server', args: ['start'] },
	\ #{ name: 'cpp', filetype: ['c', 'cpp'], path: 'clangd', args: ['--background-index'] },
	\ #{ name: 'haskell', filetype: ['haskell'], path: 'haskell-language-server', args: ['--lsp'] },
	\ #{ name: 'qml', filetype: ['qml'], path: 'qmlls6' },
	\ #{ name: 'rust', filetype: ['rust'], path: 'rust-analyzer', syncInit: 1 },
\ ])
```

But the nix language server is only available, when I am inside a `nix-shell`, which means that as a sideeffect all other LSPs are not added as well, even though they are available.

This patch fixes this by just continuing with the next element in the list, if the option is set.

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -598,13 +598,13 @@ export def AddServer(serverList: list<dict<any>>)
       if !opt.lspOptions.ignoreMissingServer
         util.ErrMsg($'LSP server {server.path} is not found')
       endif
-      return
+      continue
     endif
     var args: list<string> = []
     if server->has_key('args')
       if server.args->type() != v:t_list
         util.ErrMsg($'Arguments for LSP server {server.args} is not a List')
-        return
+        continue
       endif
       args = server.args
     endif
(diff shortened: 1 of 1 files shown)


andlrc:
I think this is fair. But I also think we should not abort when one server is not found. Instead it should work as it does when the the filetype or path is missing:

```
    if !server->has_key('filetype') || !server->has_key('path')
      util.ErrMsg('LSP server information is missing filetype or path')
      continue
    endif
```

And same goes for the three other checks in the function, simply substitute `return` with `continue`.

But that might just be me?

vimpostor:
> Or at least make all configuration errors + missing path behave the same.

OK, I think that makes sense. This makes this independent of ignoreMissingServer too and should be a better default.

PR #333, opened 2023-06-01
Fix bug where buffer completion spams

girishji:
Steps to reproduce:
- Enable buffer completion
- Set the matching option to 'case'
- Search for 'xyz' where 'xy' has completions
- Backspace to remove 'z'
- You'll see all items from current buffer show up in completion list that are not related to 'xy'

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -228,8 +228,9 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
       # snippet completion.  Needs a snippet plugin to expand the snippet.
       # Remove all the snippet placeholders
       d.word = MakeValidWord(d.word)
-    elseif !lspserver.completeItemsIsIncomplete
-      # Don't attempt to filter on the items, when "isIncomplete" is set
+    elseif !lspserver.completeItemsIsIncomplete || opt.lspOptions.useBufferCompletion
+      # Filter items only when "isIncomplete" is set (otherwise server would
+      #   have done the filtering) or when buffer completion is enabled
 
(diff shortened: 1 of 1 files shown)


girishji:
Another issue: buffer completion stutters when file gets over 3000 lines. Maybe keywords should be cached unless user types a space of newline.

autoload/lsp/completion.vim
```diff
@@ -228,8 +228,9 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
       # snippet completion.  Needs a snippet plugin to expand the snippet.
       # Remove all the snippet placeholders
       d.word = MakeValidWord(d.word)
-    elseif !lspserver.completeItemsIsIncomplete
-      # Don't attempt to filter on the items, when "isIncomplete" is set
+    elseif !lspserver.completeItemsIsIncomplete || opt.lspOptions.useBufferCompletion
+      # Filter items only when "isIncomplete" is set (otherwise server would
+      #   have done the filtering) or when buffer completion is enabled
```
Shane-XB-Qian, line 233: || opt.lspOptions.useBufferCompletion ?

i felt i have no confident on this, or if this needed, then how about those snip compl etc options?

// bufcompl had been noticed that's risk (or perf impact) option which had noticed in the doc, so i think user should have awareness be careful to enable it.

Shane-XB-Qian:
if this required, how about snip compl option?
e.g ultisnip compl would `var txt = parts[0]->readfile()[parts[1]->str2nr() : parts[1]->str2nr() + 20]` readfile and parse for each items.

Issue #148, opened 2023-03-12
[Feature Request] Make possible to extend features with custom plugins


rcasta74:
I have just seen that a commit was added to define custom notification handlers; it would be nice to have a more generic way to:
1) override any handler
2) add a new lsp request

Examples of usage:
1) override `ShowReferences` to use `fzf` plugin (this will give a smoother look is `fzf` is already used for other commands)
2) add `java/classFileContents` request for jdt-ls (when decompiler boundle is enable in jdt-ls, it can be used to get the decompiled source for a reference with `jdt://` URI)

rcasta74:
> 1. I think that it could make sense to be able to provide functions for rendering the references

Yeah, I was thinking to something like this. The function `lspserver.ShowReferences` currently send the reqest to the server and then it processes the reply calling the function `symbol.ShowReferences`. If instead it calls a registered function and the possibility to override the registered function is given, we can customize the way how the reply is processed.
`ShowReferences` was only an example, it could be done for all lsp requests or for a subset of them that could make sense to be overriden.

> 2. How would this work in a general purpose language server client? Is it a custom code action?

No, `java/classFileContents` is not a code action but a custom request available only for jdt-ls: the client send the request to the server and the server reply with the text of decompiled class.
Anyway looking deeper in the code it seems that adding a new request is already possible: with the exported function `buffer.CurbufGetServer` I can get the lspserver for the current buffer and then I have access  to the function `rpc` to send the request. I will give it a try in the next days.

yegappan:
It is good idea to add hooks to allow using fzf with the language server results.  One problem I see is that the fzf custom hook (fzf#run) allows jumping to a file but not to a specific location in the file.  Need to figure out a way to jump to a particular location in the selected file.

yegappan:
I have been thinking about separating the UI layer from the language server interface.  This will allow exposing APIs directly to the language server interface.  For users who want the LSP functionality out of the box, the built-in UI will be sufficient.  The advanced users can directly use the language server API and customize the UI.

sigzegv:
Is extension possible new or not yet ? didn't saw anything in documentation.

jclsn:
At least the second point is now possible via https://github.com/yegappan/lsp/pull/640

PR #334, opened 2023-06-02
Implement Ale integration support

vimpostor:
This adds an `aleSupport` option that is off by default. If enabled, the plugin will send all diagnostics to [Ale](https://github.com/dense-analysis/ale) which allows them to be displayed together with all other linter warnings and improves integration, e.g. all diagnostics will be available in the location list regardless of whether they come from Ale or if they come from lsp.

This is the equivalent to the `"diagnostic.displayByAle"` option in coc.nvim.

Here is a demo where all linter warnings and LSP warnings are displayed nicely together:
242944188-4a232dc7-10bf-4352-98db-7c2716d24a29

The first warning comes from clazy (a linter, that is not available as LSP and only in Ale) and the second and third warning come from lsp, which were then sent to Ale for display.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ Some of the LSP plugin features can be enabled or disabled by using the LspOptio
 Here is an example of configuration with default values:
 ```viml
 call LspOptionsSet(#{
+	\   aleSupport: v:false,
 	\   autoComplete: v:true,
 	\   autoHighlight: v:false,
 	\   autoHighlightDiags: v:true,
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -82,6 +82,10 @@ export def InitOnce()
   hlset([{name: 'LspDiagVirtualText', default: true, linksto: 'LineNr'}])
   prop_type_add('LspDiagVirtualText', {highlight: 'LspDiagVirtualText',
                                        override: true})
+
+  if opt.lspOptions.aleSupport
+    autocmd_add([{group: 'LspAleCmds', event: 'User', pattern: 'ALEWantResults', cmd: 'AleHook(g:ale_want_results_buffer)'}])
+  endif
 enddef
 
 # Sort diagnostics ascending based on line and character offset
@@ -235,6 +239,29 @@ def DiagsRefresh(bnr: number)
   signs->sign_placelist()
 enddef
 
+# Sends diagnostics to Ale
+def SendAleDiags(bnr: number, timerid: number)
+  if !diagsMap->has_key(bnr)
+    return
+  endif
+
+  # Conver to Ale's diagnostics format (:h ale-loclist-format)
+  ale#other_source#ShowResults(bnr, 'lsp', diagsMap[bnr].sortedDiagnostics->mapnew((_, v) => {
+     return {text: v.message,
+             lnum: v.range.start.line + 1,
+             col: util.GetLineByteFromPos(bnr, v.range.start) + 1,
+             end_lnum: v.range.end.line + 1,
+             end_col: util.GetLineByteFromPos(bnr, v.range.end) + 1,
+             type: "EWIH"[v.severity - 1]}
+  }))
+enddef
+
+# Hook called when Ale wants to retrieve new diagnostics
+def AleHook(bnr: number)
+  ale#other_source#StartChecking(bnr, 'lsp')
+  timer_start(0, function('SendAleDiags', [bnr]))
+enddef
+
 # New LSP diagnostic messages received from the server for a file.
 # Update the signs placed in the buffer for this file
 export def ProcessNewDiags(bnr: number)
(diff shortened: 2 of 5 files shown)


berggeist:
Just asking, couldn't the same be achieved without tight integration of ALE into this lsp-plugin?

In a very early stage of this lsp-plugin, I was not really happy with the UI of diagnostics implemented in this lsp-plugin (of course this has changed :smile: ) therefore by using a combination of the option `autoPopulateDiags` I filled all diagnostics into the locations list and by listening to the autocommand `LspDiagAttached` I displayed the diagnostic messages in my own UI. I.e without touching the code of this lsp-plugin. When I quickly skim through your code it seem for me the same logic, just integrated into the internals of the lsp-code.

Of course by providing a new option this would be far easier than writing yourself a couple of viml code in your vimrc, but on the other side creates dependencies to ALE.

girishji:
You can add multiple servers to same filetype, sort of like null-ls in nvim. I think this is what I remember from reading the documentation sometime ago. You could add `clazy` to work with main language server. That way clazy diags are also handled by this plugin.

My point is since there is already a way (IIRC), why complicate this plugin?

To satisfy everyone, I think there could be callbacks embedded in this plugin for not just diagnostics but also for completion. This plugin would provide adequate implementation (good for 90% users), but specialized use cases can take advantage of callbacks.

berggeist:
As I understood @vimpostor `clazy` does not provide a lsp-server interface for its linter results and that's the reason he's going to use ALE as _main  UI_ and he wants to _forward_ other lsp-servers diagnostic messages, that are _caught_ by this lsp-client-plugin, to ALE. At least that is my understanding, but I don't know `clazy` at all.

According to this issue #148  there is already some, at least, _brainstorming_ about providing an extension interface for this lsp-plugin by custom plugins. And in my opinion that is the right way ahead, before this lsp-code is too tightly integrated with other vim plugins, e.g. as it already happened with vsnip integration and ultisnip integration.

But let's @yegappan decide where to go

Shane-XB-Qian:
ale itself had lsp functionality too, that maybe looks a bit duplicated.

vs now this lsp diag UI feat i think had been refined workable mostly, 
// though sometime diag prop highlight may disappeared somehow..

maximyurevich:
we can use efm-langserver instead ALE

vimpostor:
I understand that this functionality can be achieved without Ale, but this is not just about getting the clazy linter to run together with the LSP, it's also about the extra functionality of Ale and to be honest about the fact that Ale works ootb with pretty much everything.

I thought it wouldn't be a problem to add Ale support directly in the plugin, as this plugin already contains integrations with a lot of other plugins, e.g. UtilSnips or vsnip so I honestly don't see the difference here.

But I also understand how this can quickly get bloated so I guess the decision is with @yegappan.

> we can use efm-langserver instead ALE

I didn't know about [efm-langserver](https://github.com/mattn/efm-langserver), thanks, it looks like an editor-independent variant of neovim's null-ls. It feels a bit weird to shoehorn every linter and formatter into an LSP though, so I am not sure if it works as well as Ale.

I also prefer the diagnostics display of Ale over the one's from this plugin, e.g. I like the option to only show the virtual text on the line where the cursor is, but this can of course be improved.

Shane-XB-Qian:
> I thought it wouldn't be a problem to add Ale support directly in the plugin, as this plugin already contains integrations with a lot of other plugins, e.g. UtilSnips or vsnip so I honestly don't see the difference here.

the difference is usnip and vsnip just a single function with an option to toggle.
vs this spread a bit (though not much), that maybe a problem later no one knew how to maintain those code. :rofl: 

i actually like your idea, but looks ale used to integrate others into itself, vs here this perhaps wish can be lightweight.
// at least i wish this lsp codebase as lightweight as possible.
// if this integration can be limited to single place, or let's help to improve current diag of this lsp codebase instead. :smile:

berggeist:
> but looks ale used to integrate others into itself

... more the other way around. ALE's documentation describes its API for _being integrated into other lsp clients_ here:

https://github.com/dense-analysis/ale/blob/7021ed0c6859ec8b1bf5012d1276d54a786449c3/doc/ale.txt#L256

and when I understand @vimpostor 's implementation correctly, he integrated this ALE API into the lsp-code. So nothing wrong per se.

The only thing I'm asking myself, is it the right way to add more and more dependencies into the lsp code. 

One main reason for me, using this lsp-code is its small and understandable code base, which makes it, in my opinion, easy to maintain and to extend according to new language server features. But this is only my humble opinion as mainly a user of this plugin and it is by no means my intention to block any contribution to this plugin.

andlrc:
> One main reason for me, using this lsp-code is its small and understandable code base, which makes it, in my opinion, easy to maintain and to extend according to new language server features. But this is only my humble opinion as mainly a user of this plugin and it is by no means my intention to block any contribution to this plugin.

Yes this is the main reason for me picking up this LSP as well.

vimpostor:
What is the opinion on instead extending the `LspDiagsUpdated` autocmd such that it is possible to retreive all diagnostics without having to use the hack with setting `autoPopulateDiags` and then having to go through the location list.

This would avoid hardcoding any Ale stuff in this plugin, but would give the user the power to handle the diagnostics in any arbitrary way.

It would be sufficient to expose `diagsMap[bnr].sortedDiagnostics` behind a function `LspGetDiags(bnr)` for the user.

PR #346, opened 2023-06-18
fix: may check buf feat before it's ready

Shane-XB-Qian:
// supposed this fix there no side effect,
// or code err if buf lsp server actually not ready.


Shane-XB-Qian:
not sure the side effect, so throwing warning msg instead, let's see the occurrence.

Shane-XB-Qian:
let it be...

PR #347, opened 2023-06-25
Make buffer completion responsive for large files

girishji:
Since buffer completion processes the current buffer everytime user types something, it will degrade the experience for large files.

This change adds a timeout to buffer completor function. Processing current buffer is cut short when timeout is exceeded. Setting timeout to 0 will revert back to existing behaviour.

Default is set to 100 ms, good for scanning 6000 lines on M1 macbook.

It is possible to get fancy by scanning locality of cursor first but such complication may not be worth the complexity.

Tested on >20k line files (I have to open these large C files filled with hw specs occasionally).

M  autoload/lsp/options.vim
M  doc/lsp.txt

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -114,22 +114,30 @@ enddef
 
 # add completion from current buf
 def CompletionFromBuffer(items: list<dict<any>>)
-    var words = {}
-    for line in getline(1, '$')
-        for word in line->split('\W\+')
-            if !words->has_key(word) && word->len() > 1
-                words[word] = 1
-                items->add({
-                    label: word,
-                    data: {
-                        entryNames: [word],
-                    },
-                    kind: 26,
-                    documentation: "",
-                })
-            endif
-        endfor
+  var words = {}
+  var start = reltime()
+  var linenr = 1
+  for line in getline(1, '$')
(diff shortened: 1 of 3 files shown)


autoload/lsp/completion.vim
```diff
@@ -114,22 +114,30 @@ enddef
...
+	  data: {
+	    entryNames: [word],
+	  },
+	  kind: 26,
+	  documentation: "",
+	})
+      endif
     endfor
+    # Check every 200 lines if timeout is exceeded
+    if opt.lspOptions.bufferCompletionTimeout > 0 && linenr % 200 == 0 &&
```
yegappan, line 135: For optimization reasons, can you store the value of opt.lspOptions.bufferCompletionTimeout in a local variable outside the for loop and then use that local variable here?  Otherwise, Vim will recompute the Dict index every time.  You can check the output of the `:disassemble` command for the difference in the number of instructions.

girishji: I was thinking it will optimize dict key search, but it does not. There is LOADSCRIPT and USEDICT instructions inside for-loop. So you're correct.

girishji:
Would `searchpos()` be more appropriate since it has timeout option? It also searches around cursor position which is more relevant than searching top-to-bottom.

PR #357, opened 2023-07-08
fix: inlay hints maybe a bit delayed if sync init

Shane-XB-Qian:


autoload/lsp/inlayhints.vim
```diff
@@ -157,6 +157,14 @@ export def BufferInit(lspserver: dict<any>, bnr: number)
 		group: 'LspInlayHints',
 		cmd: $'LspInlayHintsUpdateStop({bnr})'})
 
+  # Inlay hints maybe a bit delayed if it was a sync init lsp server.
+  if lspserver.syncInit
+    acmds->add({bufnr: bnr,
+		  event: ['User'],
+		  group: 'LspAttached',
+		  cmd: $'LspInlayHintsChanged({bnr})'})
```
yegappan, line 165: The LspAttached autocmd handler needs to be executed only once for this buffer.  So can you add `once: v:true` to the autocmd Dict?

Shane-XB-Qian: 1, i donot know why, but once added `once: true`, that au may not be kicked-off.
2, normally bufinit would/should be done only once when addFile()?? unless manually `doau` it again.
3, `exists('#User#LspAttached')` seems should be `exists('#LspAttached#User')` but actually worked, weird.

yegappan: If `once` is not used, then the autocmd will be present forever (even though it will not be invoked).
If you run the `:autocmd` command, you will see too many of these.

Are you saying that if you set `once` to `true`, then the autocmd is not invoked?

Shane-XB-Qian: > If `once` is not used, then the autocmd will be present forever (even though it will not be invoked).
> If you run the `:autocmd` command, you will see too many of these

yes, since it's on buf, if that's a concern, then maybe a bit annoyed, but:


> Are you saying that if you set `once` to `true`, then the autocmd is not invoked?

yes, au was not invoked at all, no even once (nothing).

Shane-XB-Qian:
r.i.p bram.

yegappan:
I don't know why you closed this PR.  I intend to continue to develop and maintain the LSP plugin.

CoelacanthusHex:
Hope this PR can be merged soon. Without this, my rust-analyzer doesn't until I execute `:LspInlayHint enable` manually.

PR #362, opened 2023-07-24
Markdown parser enhancement

rcasta74:
Handle tabs and soft/hard line breaks


yegappan:
Thanks for the patch.  With this patch, it looks like the length calculated for some text properties is -1, leading to an `invalid argument` error.  This can be reproduced using the following steps:

```
-------- x.txt ----------------
From `<stdio.h>`  
Write formatted output to STREAM.  
This function is a possible cancellation point and therefore not marked with \_\_THROW.
-------- x.txt ----------------

$ vim
 :pedit x.txt
 :wincmd P
 :set filetype=lspgfm
```

You should see the `E474: Invalid argument` error with the above steps.

rcasta74:
Found the error, sorry for the mistake.
In the new PR I have fixed an old error as well.

PR #363, opened 2023-07-24
Fix markdown parser

rcasta74:
Compute correct position of the line break and of the end of the text property when splitting the line.
Add the correct list of text properties for heading block.

### DIFF
--- a/autoload/lsp/markdown.vim
+++ b/autoload/lsp/markdown.vim
@@ -428,9 +428,9 @@ def SplitLine(line: dict<any>, indent: number = 0): list<dict<any>>
     props: []
   }
   for prop in line.props
-    if prop.col + prop.length < pos
+    if prop.col + prop.length - 1 < pos + 1
       cur_line.props->add(prop)
-    elseif prop.col >= pos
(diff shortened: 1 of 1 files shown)


rcasta74:
ok, I missed that right operand of slice was inclusive in vimscript.
I'm using this function
```
vim9script

def g:Test_md()
  var text = getline(1, '$')
  var winid = text->popup_dialog(
    {moved: 'any',
     close: 'click',
     fixed: true,
     maxwidth: 80,
     minwidth: 80,
     border: [0, 1, 0, 1],
     borderchars: [' ']})
  win_execute(winid, $'setlocal ft=lspgfm')
enddef
```
to test my changes in a popup, but there the character is not visible so I missed it :)

I was merging the new commit with the old one, but I have seen later that you have already merged it.
Now I have messed up my branch, so I will open the new PR later, or if you prefer you can fix it by yourself:
https://github.com/yegappan/lsp/blob/c96fdfd73a38039b20d4e50e62607b39459be688/autoload/lsp/markdown.vim#L423
it should be `pos - 1` instead

PR #373, opened 2023-08-01
fix: keep omnicompl really default enable

Shane-XB-Qian:
`omnicompl` option was default `true` really, the recent change made it be possible `false`.
// this to fix it.

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -571,14 +571,9 @@ export def AddServer(serverList: list<dict<any>>)
       util.ErrMsg('LSP server information is missing filetype or path')
       continue
     endif
+
     # Enable omni-completion by default
-    var omnicompl_def: bool = false
-    if opt.lspOptions.omniComplete == true
-	|| (opt.lspOptions.omniComplete == null
-	    && !opt.lspOptions.autoComplete)
-      omnicompl_def = true
-    endif
-    server.omnicompl = server->get('omnicompl', omnicompl_def)
+    server.omnicompl = server->get('omnicompl', opt.lspOptions.omniComplete)
 
     if !server.path->executable()
       if !opt.lspOptions.ignoreMissingServer
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -73,9 +73,8 @@ export var lspOptions: dict<any> = {
   # Suppress adding a new line on completion selection with <CR>
   noNewlineInCompletion: false,
 
-  # Omni-completion support.  To keep backward compatibility, this option is
-  # set to null by default instead of false.
-  omniComplete: null,
+  # Omni-completion support.
+  omniComplete: true,
 
   # Open outline window on right side
   outlineOnRight: false,


yegappan:
There is a deliberate reason why I made this change.  This change allows omni-completion to be disabled independent of auto-completion.  Previously, when auto-completion is disabled, omni-completion is enabled by default.  There are use cases where a user needs to disable both auto-completion and omni-completion globally.  With your change, these features cannot be independently enabled or disabled.

Shane-XB-Qian:
> There are use cases where a user needs to disable both auto-completion and omni-completion globally. With your change, these features cannot be independently enabled or disabled.

it can be both independently enabled or disabled, only different with your last change is to make it be really default true.

Shane-XB-Qian:
> There are use cases where a user needs to disable both auto-completion and omni-completion globally.
> With your change, these features cannot be independently enabled or disabled.

it can be both independently enabled or disabled,
only different with your last change is to make it be really default true, and really independent with auto-compl.

-- 
shane.xb.qian

yegappan:
>
> you changed this several times, last time before latest one you modified it
> to if auto-complete disabled then omnicompl enabled, the latest change you
> recovered it to independent auto-complete and omnicompl which both can be
> enabled, and added a vim session level omnicomplete option.
>

The changes I made didn't change the default behavior. The behavior previously
was:

1. Auto completion is enabled by default and omni-complete is not enabled.
2. If the autocomplete option is set to false, then omni-complete is enabled.
3. If autocomplete is set to false and the per-server omnicompl flag is
   set to false, then omni-completion is disabled for that server.

My changes added another option to enable or disable omni-completion.
If this broke something else, let me know.

>
> now this/my
> change kept it and same like the original design, but get default from vim
> session level omnicomplete option too, however your latest change make that
> default 'false' which server level omnicompl maybe false, which changed the
> consistent default behavior, this change/my pr just to fix that to keept it
> really default true, unless user changed vim session level omnicomplete
> option value, and didnot change omnicompl for that server.
>

Currently to change the autocomplete and omnicomplete option values, the
Vim instance need to be restarted.  This makes it difficult to test these
options.  I am working on changes to dynamically enable or disable these
options.  To support this, the default option value should be set to
null and not true.  So we can detect whether the user specified the
option value or the default value is used.

Shane-XB-Qian:
> The changes I made didn't change the default behavior. The behavior previously
> was:
> 
> 1. Auto completion is enabled by default and omni-complete is not enabled.
> 2. If the autocomplete option is set to false, then omni-complete is enabled.
> 3. If autocomplete is set to false and the per-server omnicompl flag is
>    set to false, then omni-completion is disabled for that server.
> 
> My changes added another option to enable or disable omni-completion.
> If this broke something else, let me know.

no, if you remembered it was me design and add this 'omnicompl' cfg at very early of this repo startup.
at that time, you named it '7x24 compl' instead of auto-compl with omnifunc was set too,
i was adding this 'omnicompl' to allow disable it.
so previous behavior was those can be both disabled or enabled.


> Currently to change the autocomplete and omnicomplete option values, the
> Vim instance need to be restarted.  This makes it difficult to test these
> options.  I am working on changes to dynamically enable or disable these
> options.  To support this, the default option value should be set to
> null and not true.  So we can detect whether the user specified the
> option value or the default value is used.

i felt you're confused omnicompl with feature 'completion', this is just all about 'omnifunc',
so how come this related to 'restart'?

this change is just to make auto-compl and omni-compl both be independent each other.

only different at begining of this repo startup was omnicompl was for ft level,
now since there is multiple-servers feat now, adding omnicompl to one server but another one not seems maybe confused.
// not checking that code detail now, but i guess it would be since it was not designed for multiple-servers.
a.k.a : to multiple-servers should be avoiding to add omnicompl option (unless both false or true for those servers)

as for your new added `omnicomplete`, that's a vim session level, it's ok if you like to global disable omni-compl,
but to keep consistent, it should be default true.

-- 
shane.xb.qian

Shane-XB-Qian:
r.i.p bram.

Andilutten:
I can confirm the fact that omnicomplekion is broken for me in master branch (#380), this branch works like a charm tho.

Andilutten:
What is the status of this ? 
I find it kind of frustrating not being able to make omnicompletion work :O

yegappan:
> What is the status of this ? I find it kind of frustrating not being able to make omnicompletion work :O

I will look into this over this weekend.

Shane-XB-Qian:
it actually didnot bother me, but how about just merging it? :smile:
// your previous modification i felt perhaps was a bit too tricky, 
// a.k.a making life easy, and you can get some rest as well.. :rofl:

TheDreadedAndy:
I'm not convinced this patch is actually necessary. AFAIK, the current version is behaving as the documentation says:

- If the server specifies omnicompl, that is used
- If the global config does, that is used.
- Otherwise, the value is !cfg.autoComplete

I struggled with issue #380 earlier tonight, and the actual problem (for both me and the OP of that issue) seems to be that LspAddServer was called before LspOptionSet, which prevents the value of cfg.omniComplete from being used (since LspAddServer is the function that inspects it).

Shane-XB-Qian:
this PR was to make it easy, tho not have to.
// and https://github.com/yegappan/lsp/pull/475#discussion_r1527156125

PR #374, opened 2023-08-06
Get hover from multiple lsps

MartinAskestad:
I don't really intend for this to be merged, but it could perhaps be a stepping stone for a more robust solution.

What I've done is I have the Hover function retrieve a list of LSPs for the current buffer that have the hover feature. And then loop through these and request the hover information. For the LSP that can't find hover info for the current position, it will just print out an error saying it can't find any info. This is the same behavior as today. But if another LSP finds something to display in the hover popup, it will display this.

So my test consists of angular-language-server, typescript-language-server and vscode-html-language-server. In a typescript document with inline html, I can hover on the html-properties and angular-ls will respond with a popup. But if I hover on something in typescript, typescript-ls will respond. Works as indented.

Here I'm hovering on a html property within a string in typescript, and angular-ls responds.
image

Here I'm hovering over a typescript property and typescript-ls responds.
image

The warning on the bottom comes from the other LSP that didn't find any hover info. This should be pretty easy to hide?

There is however an issue when two or more LSPs can show hover on the same position. In html-documents both vscode-html-ls and angular-ls can show hover on html-tags. Vscode has a lot of info on what kind of tag you are hovering over and also you can hover on attribute names. Angular-ls just says what kind of tag it is. But the overlap can give a result you didn't want, so here a mechanism for prioritizing the LSP is needed. I couldn't really figure out a decent way to do that yet.

Again, I don't really expect you to merge this, but what do you think if the idea that two or more LSPs can respond on the same feature on the same document? If this would work on things like completion and such it would really be terrific for my when I work with my angular projects in particular.

### DIFF
--- a/autoload/lsp/buffer.vim
+++ b/autoload/lsp/buffer.vim
@@ -59,6 +59,54 @@ var SupportedCheckFns = {
   workspaceSymbol: (lspserver) => lspserver.isWorkspaceSymbolProvider
 }
 
+# Returns the LSP servers for the buffer "bnr".  If "feature" is specified,
+# then returns the LSP servers that provides the "feature".
+# Returns an  empty list if the server is not found.
+export def BufLspServerGetList(bnr: number, feature: string = null_string): list<dict<any>>
+  if !bufnrToServers->has_key(bnr)
+    return []
+  endif
+
+  if bufnrToServers[bnr]->empty()
+    return []
+  endif
+
+  if feature == null_string
+    return bufnrToServers[bnr]
+  endif
+
+  if !SupportedCheckFns->has_key(feature)
+    # If this happns it is a programming error, and should be fixed in the
+    # source code
+    :throw $'Error: ''{feature}'' is not a valid feature'
+  endif
+
+  var SupportedCheckFn = SupportedCheckFns[feature]
+
+  var possibleLSPs: list<dict<any>> = []
+
+  for lspserver in bufnrToServers[bnr]
+    if !lspserver.ready || !SupportedCheckFn(lspserver)
+      continue
+    endif
+
+    possibleLSPs->add(lspserver)
+  endfor
+
+  if possibleLSPs->empty()
+    return []
+  endif
+
+  # LSP server is configured to be a provider for "feature"
+  possibleLSPs->filter((_, lsp) => lsp.features->get(feature, false))
+
+  # Return the LSP servers that supports "feature" and doesn't have it
+  # disabled
+  possibleLSPs->filter((_, lsp) => lsp.featureEnabled(feature))
+
+  return possibleLSPs
+enddef
+
 # Returns the LSP server for the buffer "bnr".  If "feature" is specified,
 # then returns the LSP server that provides the "feature".
 # Returns an empty dict if the server is not found.
@@ -142,6 +190,12 @@ export def BufLspServersGet(bnr: number): list<dict<any>>
   return bufnrToServers[bnr]
 enddef
 
+# Returns the LSP servers for the current buffer with the optionally "feature".
+# Returns an empty list if the server is not found.
+export def CurbufGetServerList(feature: string = null_string): list<dict<any>>
+  return BufLspServerGetList(bufnr(), feature)
+enddef
+
 # Returns the LSP server for the current buffer with the optionally "feature".
 # Returns an empty dict if the server is not found.
 export def CurbufGetServer(feature: string = null_string): dict<any>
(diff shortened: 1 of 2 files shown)


autoload/lsp/buffer.vim
```diff
@@ -59,6 +59,54 @@ var SupportedCheckFns = {
...
 
+# Returns the LSP servers for the buffer "bnr".  If "feature" is specified,
+# then returns the LSP servers that provides the "feature".
+# Returns an  empty list if the server is not found.
+export def BufLspServerGetList(bnr: number, feature: string = null_string): list<dict<any>>
+  if !bufnrToServers->has_key(bnr)
+    return []
+  endif
+
+  if bufnrToServers[bnr]->empty()
```
yegappan, line 70: Can you merge this "if" check with the previous one?

autoload/lsp/buffer.vim
```diff
@@ -59,6 +59,54 @@ var SupportedCheckFns = {
...
+  if possibleLSPs->empty()
+    return []
+  endif
+
+  # LSP server is configured to be a provider for "feature"
+  possibleLSPs->filter((_, lsp) => lsp.features->get(feature, false))
+
+  # Return the LSP servers that supports "feature" and doesn't have it
+  # disabled
+  possibleLSPs->filter((_, lsp) => lsp.featureEnabled(feature))
```
yegappan, line 105: This test and the test on line 101 do the same.

autoload/lsp/buffer.vim
```diff
@@ -160,6 +214,29 @@ export def BufHasLspServer(bnr: number): bool
   return !lspserver->empty()
 enddef
 
+# Returns the LSP servers for the current buffer with the optinally "feature" if
```
yegappan, line 217: Type: "optinally" -> "optional"

autoload/lsp/lsp.vim
```diff
@@ -764,12 +764,14 @@ enddef
...
-  var lspserver: dict<any> = buf.CurbufGetServerChecked('hover')
-  if lspserver->empty()
+  var lspservers: list<dict<any>> = buf.CurbufGetServerCheckedList('hover')
+  if lspservers->empty()
     return
   endif
 
-  lspserver.hover(cmdmods)
+  for lspserver in lspservers
+    lspserver.hover(cmdmods)
```
yegappan, line 773: If one LSP server provides the hover information, then there is no need to query other servers for hover information.

yegappan:
The warning message if the hover information is not found can be changed to display only if all the LSP servers return empty information.

Also, after valid information is returned by an LSP server, there is no need to query other subsequent servers (as this will slow down the operation).

This approach can be used for synchronous LSP requests.  But for asynchronous LSP requests this approach will not work.  Completion currently uses asynchronous RPC to get the completion matches from the LSP server.

autoload/lsp/lsp.vim
MartinAskestad: Doesn't thst check for hover for the whole document and not the current location? Since different LSPs can give more or less relevant info depending on location, each needs to be queried?

yegappan: How will we display the hover information in a popup from different LSP servers?  We can display the information from only one LSP server.

MartinAskestad: That is something I still haven't solved. In my example above, angular-ls gave a reply when I triggered hover on the html-code that is inside a string. And the reason there is also a warning about documentation not found, that is typescript-ls answering. The same with the other example where typescript-ls answered with a good reply but angular-ls answered documentation not found.

In those cases, the easiest solution is probably to just nog show the warning. if no LSP gives an answer, perhaps not show anything at all? Perhaps that is fine?

When two or more LSPs actually gives an answer, now you probably have to go by priority and pick the first LSP added to your config. And that is how it works today I believe?

I think there are a few good reasons why someone might want to use several LSPs on the same document. I personally would be helped by having typescript and angular. I know I probably would want to have typescript and eslint diagnostics on the same document. I still haven't gotten eslint to work, but when it works I would have syntax diagnostics from typescript and code style diagnostics from eslint. This would be very useful. But for that to work you have to get the response from all the LSPs, not just the first one.

andlrc:
@MartinAskestad I like this idea. One major problem in implementing this is that `:LspHover` is async as it requests the server for hover information. Which means that we would proper have to refactor quite a bit of code to be able to only request hover from just enough servers to get a valid result.

MartinAskestad:
Full disclaimer, I'm not that versed in vim-script. But if there is something similar to JavaScripts Promise.all, it would be pretty trivial to call several LSPs asyncly and handle the results.

PR #383, opened 2023-08-29
fix: ignore `eslint/status` and `taplo/didChangeSchemaAssociation`

maximyurevich:
Ignore `eslint/status` and `taplo/didChangeSchemaAssociation` unsupported notification messages

### DIFF
--- a/autoload/lsp/handlers.vim
+++ b/autoload/lsp/handlers.vim
@@ -116,7 +116,9 @@ export def ProcessNotif(lspserver: dict<any>, reply: dict<any>): void
       # pyright language server notifications
       'pyright/beginProgress',
       'pyright/reportProgress',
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
what is `taplo`?
could you add some comment for prompt?
at least it was not pyright.

--
shane.xb.qian

maximyurevich:
> what is `taplo`? could you add some comment for prompt? at least it was not pyright.
> […](#)
> -- shane.xb.qian

taplo is lsp for toml language server

Shane-XB-Qian:
for less used lsp server better be using self-custom filter.
// I recall it can do that.
// meanwhile it should add some comment in code for such less used lsp server if really added such into codebase.

--
shane.xb.qian

PR #386, opened 2023-09-01
fix: remove less used filter

Shane-XB-Qian:
// for unsupported msg of less used lsp server better be using self-custom filter

refine https://github.com/yegappan/lsp/pull/383#issuecomment-1702008031


Shane-XB-Qian:
and https://github.com/yegappan/lsp/pull/387 ?

PR #387, opened 2023-09-03
fix(handlers): ignore `sqlLanguageServer.finishSetup`

maximyurevich:


Shane-XB-Qian:
IMO not worth to add unsupported msg of less used lsp server into common main filter, you can add it via self-custom filter.
// have you seen comment at last similar PR?

--
shane.xb.qian

maximyurevich:
> IMO not worth to add unsupported msg of less used lsp server into common main filter, you can add it via self-custom filter. // have you seen comment at last similar PR?
> […](#)
> -- shane.xb.qian

Why taplo, sqlls or eslint are less used?

yegappan:
The LSP plugin should work without major issues with as many language servers as possible out of the box.   So I will merge this PR.

Shane-XB-Qian:
so you plan to add all "you" think unsupported msg of various lsp servers like hard code like odd code into codebase?
that's the so-called "out of box"?

--
shane.xb.qian

Issue #367, opened 2023-07-28
Auto import not working (tsserver)


basilgood:
Thanks for lsp plugin.
I have been using it for some time and I was very satisfied.
But recently it no longer auto imports.
Please help me bring auto import back.

vimrc:
```vim
  vim9script
  packadd minpac
  minpac#init()
  minpac#add('k-takata/minpac', {'type': 'opt'})
  minpac#add('yegappan/lsp', {'type': 'opt'})
  minpac#add('dense-analysis/ale', {'type': 'opt'})

  silent! packadd lsp
  var lspServers = [
    {
      name: 'tsserver',
      filetype: ['javascript', 'typescript'],
      path: 'typescript-language-server',
      args: ['--stdio'],
    },
  ]

  call LspOptionsSet({
    aleSupport: true,
    noNewlineInCompletion: true,
    showDiagInPopup: false,
    echoSignature: true,
    usePopupInCodeAction: true,
  })
  call LspAddServer(lspServers)

  silent! packadd ale
  g:ale_disable_lsp = 1
  g:ale_virtualtext_cursor = 0
  g:ale_sign_error = '▸▸'
  g:ale_sign_warning = '▸ '
  g:ale_sign_info = '▸ '
  g:ale_echo_msg_error_str = 'E'
  g:ale_echo_msg_warning_str = 'W'
  g:ale_echo_msg_info_str = 'I'
  g:ale_echo_msg_format = '[%linter%] %s [%severity%]'
  g:ale_sh_shfmt_options = '-i 2 -ci -sr'
  g:ale_fixers = {
    sh: ['shfmt'],
    javascript: ['prettier'],
    typescript: ['prettier'],
    html: ['prettier'],
    css: ['prettier'],
    json: ['fixjson'],
    nix: ['alejandra'],
    yaml: ['yamlfix'],
    rust: ['rustfmt'],
  }

  nnoremap ]d <Plug>(ale_next_wrap)
  nnoremap [d <Plug>(ale_previous_wrap)
  nnoremap Q :ALEFix<cr>

```

https://github.com/yegappan/lsp/assets/16057481/299e0081-fe61-43d2-bcc8-646a6c43a7f8

This is from LspServer debug on.
```Auto import from ''loglevel''```  it's there but no autoimport.

```md
[...]

PR #389, opened 2023-09-03
Add support for CompletionItem commands

vimpostor:
This patch implements support for the optional "command" field in the `CompletionItem` response to a "completionItem/resolve" request: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

Some language servers (e.g. haskell-language-server) use this field to apply other edits to the document (e.g. auto-importing functions).
While the spec technically says that "additionalTextEdits" should be used for that usecase (also see #367), apparently not all servers do and I can see the reason, because haskell-language-server also wants to show an "window/showMessage" info message to the user about the auto-import.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -7,6 +7,7 @@ import './buffer.vim' as buf
 import './options.vim' as opt
 import './textedit.vim'
 import './snippet.vim'
+import './codeaction.vim'
 
 # per-filetype omni-completion enabled/disabled table
 var ftypeOmniCtrlMap: dict<bool> = {}
@@ -590,12 +591,17 @@ def LspCompleteDone(bnr: number)
     # additional text edits.  So try to resolve the completion item now to get
     # the text edits.
     completionData = lspserver.resolveCompletion(completionData, true)
-    if completionData->get('additionalTextEdits', {})->empty()
-      return
+    if !completionData->get('additionalTextEdits', {})->empty()
+      textedit.ApplyTextEdits(bnr, completionData.additionalTextEdits)
     endif
   endif
 
-  textedit.ApplyTextEdits(bnr, completionData.additionalTextEdits)
+  if completionData->has_key('command')
+    # Some language servers (e.g. haskell-language-server) want to apply
+    # additional commands after completion.
+    codeaction.DoCommand(lspserver, completionData.command)
+  endif
+
 enddef
 
 # Initialize buffer-local completion options and autocmds
--- a/autoload/lsp/textedit.vim
+++ b/autoload/lsp/textedit.vim
@@ -213,7 +213,6 @@ export def ApplyWorkspaceEdit(workspaceEdit: dict<any>)
     return
   endif
 
-  var save_cursor: list<number> = getcurpos()
   for [uri, changes] in workspaceEdit.changes->items()
     var bnr: number = util.LspUriToBufnr(uri)
     if bnr == 0
@@ -224,8 +223,6 @@ export def ApplyWorkspaceEdit(workspaceEdit: dict<any>)
     # interface TextEdit
     ApplyTextEdits(bnr, changes)
   endfor
-  # Restore the cursor to the location before the edit
-  save_cursor->setpos('.')
 enddef
 
 # vim: tabstop=8 shiftwidth=2 softtabstop=2


autoload/lsp/completion.vim
```diff
@@ -590,12 +590,17 @@ def LspCompleteDone(bnr: number)
...
     completionData = lspserver.resolveCompletion(completionData, true)
-    if completionData->get('additionalTextEdits', {})->empty()
-      return
+    if !completionData->get('additionalTextEdits', {})->empty()
+      textedit.ApplyTextEdits(bnr, completionData.additionalTextEdits)
     endif
   endif
 
-  textedit.ApplyTextEdits(bnr, completionData.additionalTextEdits)
+  if completionData->has_key('command')
```
Shane-XB-Qian, line 598: what if it had both (just in case) ?

Shane-XB-Qian: and is it possible or what if it was delayed (or not) to get  it like additionaltextedits ?

autoload/lsp/completion.vim
```diff
@@ -590,12 +590,17 @@ def LspCompleteDone(bnr: number)
...
+    # Some language servers (e.g. haskell-language-server) want to apply
+    # additional commands after completion.
+    lspserver.executeCommand(completionData.command)
```
Shane-XB-Qian, line 601: perhaps better to use `codeaction.DoCommand()` instead.

autoload/lsp/completion.vim
vimpostor: > what if it had both (just in case) ?

Then both should be used. The command can be something completely unrelated to the additionaltextedit and the LSP spec mandates that we apply the command.

> and is it possible or what if it was delayed (or not) to get it like additionaltextedits ?

Do you know a specific language server that does this or is this just speculation? Unlike the textedit, the command name is not expensive to compute, so I don't know any reason why a server would delay it.

autoload/lsp/completion.vim
vimpostor: Why? In this case it ends up just calling lspserver.executeCommand() anyway?

autoload/lsp/completion.vim
Shane-XB-Qian: ok, if you were sure about it.

autoload/lsp/completion.vim
Shane-XB-Qian: yes, but just gave a self-custom entry to user if that alternative way.
or thought that not necessary? i donot know actually, for this `command`, 
but for uniform, so far there was no where directly call lspserver.execuateCommand

vimpostor:
@yegappan I have also removed the cursor-reset after an workspaceEdit action. It causes problems, when a workspaceEdit inserts or deletes line, as we will then likely end up on a different logical line than it was before. If there was any reason for the cursor-reset, let me know.

yegappan:
As long as the cursor is restored to the previous insertion end point so that the user can continue to enter text after a completion, it is fine to remove the cursor-reset code.  I have also seen some problems in correctly restoring the cursor if an import is automatically added or some lines are removed.  I will merge this and if we see any additional issues with cursor restoration, we can then address them.

Shane-XB-Qian:
and the logic seems wrong like before for textadditionaledits : 
seems it would NOT apply it when it do had it (vs delayed) now.

// BTW: @yegappan looks you're a bit hurry to merge recently except my PR. :smile:

vimpostor:
> and the logic seems wrong like before for textadditionaledits : seems it would NOT apply it when it do had it (vs delayed) now.

I don't really understand what you are trying to say, but if you know about a specific language server that provides `command` only delayed, then we can amend the logic like for `additionalTextEdits`. But to me this sounds like speculation right now and would be a wasted `resolveCompletion` request.

Shane-XB-Qian:
1. i meant i was concerning the cursor pos when/if they were both (additionaltextedits and command) set from lsp server specially if one was or both was delayed to process, but you said generally it was not possible happened, then ok, but just worry the cursor pos would be not correct.

2. here : 
``` vim
  if !completionData->has_key('additionalTextEdits')
    # Some language servers (e.g. typescript) delay the computation of the
```
it would NOT apply the additionaltextedits when it do had it (vs delayed) now.
// this logic seems wrong (not like before).

vimpostor:
> it would NOT apply the additionaltextedits when it do had it (vs delayed) now. // this logic seems wrong (not like before).

I understand your point now, sry that if-case somehow slipped through. I fixed it in #390

Shane-XB-Qian:
simple way should be just keeping additionalTextEdits not touch and leave it alone,
but just adding `command` before it, like following:

// if so, you above removed cursor save/restore code perhaps just leave it there too.

```vim
index 0d64e2b..b7480e6 100644
@@ -583,28 +583,29 @@ def LspCompleteDone(bnr: number)
   var completionData: any = v:completed_item->get('user_data', '')
   if completionData->type() != v:t_dict
       || !opt.lspOptions.completionTextEdit
     return
   endif

+  if completionData->has_key('command')
+    # Some language servers (e.g. haskell-language-server) want to apply
+    # additional commands after completion.
+    codeaction.DoCommand(lspserver, completionData.command)
+  endif
+
   if !completionData->has_key('additionalTextEdits')
     # Some language servers (e.g. typescript) delay the computation of the
     # additional text edits.  So try to resolve the completion item now to get
     # the text edits.
     completionData = lspserver.resolveCompletion(completionData, true)
-    if !completionData->get('additionalTextEdits', {})->empty()
-      textedit.ApplyTextEdits(bnr, completionData.additionalTextEdits)
+    if completionData->get('additionalTextEdits', {})->empty()
+      return
     endif
   endif

-  if completionData->has_key('command')
-    # Some language servers (e.g. haskell-language-server) want to apply
-    # additional commands after completion.
-    codeaction.DoCommand(lspserver, completionData.command)
-  endif
-
+  textedit.ApplyTextEdits(bnr, completionData.additionalTextEdits)
 enddef

 # Initialize buffer-local completion options and autocmds
 export def BufferInit(lspserver: dict<any>, bnr: number, ftype: string)
   if !lspserver.isCompletionProvider
     # no support for completion

```

// but anyway, if they (additionalTextEdits and command) were both there, i am curious
// cursor would be correct? since one of them may changed the pos stored in 'user_data'
//
// or perhaps generally they actually not possible do the both at same time.

-- 
shane.xb.qian

Shane-XB-Qian:
all in all, your PR was just modifying following before additionaltextedits :

+  if completionData->has_key('command')
+    # Some language servers (e.g. haskell-language-server) want to apply
+    # additional commands after completion.
+    codeaction.DoCommand(lspserver, completionData.command)
+  endif

would you mind to change it to this way? it would be the less change and less impact.
// and not remove that cursor save/restore code if so, unless really confirm it required.

vimpostor:
> // if so, you above removed cursor save/restore code perhaps just leave it there too.

It was intentionally removed to avoid the logical line changing after the user confirmed the completion. That was an actual problem that **was** happening with haskell-language-server and other LSPs as well. Can you describe any scenario where there is buggy behaviour because the cursor is not manually reset? Personally I don't know any such scenario.

> but just adding `command` before it

I am not entirely sure, but to me it sounds like `command` should be applied **after** `additionalTextEdits`, if both are present. In any case with #390 merged, from a code-perspective there is not really this difference between the two anymore, they are both equally short now.

PR #392, opened 2023-09-13
Add support for "[uint, uint]" as label

### LINKED ISSUE #391: svelteserver errors for open brackets `(`
Screenshot from 2023-09-13 11-24-16
Screenshot from 2023-09-13 11-23-43

andlrc:
Closes #391

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -477,7 +477,9 @@ def g:LspOmniFunc(findstart: number, base: string): any
 
     var [triggerKind, triggerChar] = GetTriggerAttributes(lspserver)
     if triggerKind < 0
-      return -1
+      # previous character is not a keyword character or a trigger character,
+      # so cancel omni completion.
+      return -2
     endif
 
     # first send all the changes in the current buffer to the LSP server
@@ -509,7 +511,7 @@ def g:LspOmniFunc(findstart: number, base: string): any
       return v:none
     endif
 
-    var res: list<dict<any>> = lspserver->get('completeItems', [])
+    var res: list<dict<any>> = lspserver.completeItems
     var prefix = lspserver.omniCompleteKeyword
 
     # Don't attempt to filter on the items, when "isIncomplete" is set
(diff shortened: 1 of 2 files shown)


autoload/lsp/signature.vim
```diff
@@ -92,9 +92,15 @@ export def SignatureHelp(lspserver: dict<any>, sighelp: any): void
   if sig->has_key('parameters') && sighelp->has_key('activeParameter')
     var params_len = sig.parameters->len()
     if params_len > 0 && sighelp.activeParameter < params_len
-      var label: string = sig.parameters[sighelp.activeParameter].label
-      hllen = label->len()
-      startcol = text->stridx(label)
+      // label = string | [uint, uint]
```
Shane-XB-Qian, line 95: 1, '//' is wrong comment string;
2, '[unit, unit]' was that really matching the lsp specification?

Shane-XB-Qian (changes requested):

autoload/lsp/signature.vim
```diff
@@ -92,9 +92,15 @@ export def SignatureHelp(lspserver: dict<any>, sighelp: any): void
...
-      var label: string = sig.parameters[sighelp.activeParameter].label
-      hllen = label->len()
-      startcol = text->stridx(label)
+      # label = string | [uint, uint]
+      var label = sig.parameters[sighelp.activeParameter].label
+      if label->type() == v:t_string
+        hllen = label->len()
+        startcol = text->stridx(label)
+      else
+        startcol = label[0]
```
yegappan, line 101: According to the spec:

The offsets are based on a UTF-16 string representation as `Position` and `Range` does.

So this needs a conversion to a UTF-8 offset.

andlrc: Good point. I'll not have time until next week to look at this. I also think it's worth investigating if [`ClientCapabilities.general.positionEncodings`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#positionEncodingKind) should be taken into account.

PR #396, opened 2023-09-19
fix: inlay hints maybe emit too early

### LINKED ISSUE #395: Inlay hints callback corruption?
I've noticed a weird behaviour with the gopls language server. I can consistently reproduce it on bigger projects, where I assume the `textDocument/inlayHint` request takes a long time.

The first symptom was that the workspace never finished loading so no LSP functionality worked. But I was receiving this error:

```
Error: request textDocument/inlayHint failed (no result)
```

It seemed weird that the result would be empty, but no error from the server so I extended the error message like so:

```diff
-util.ErrMsg($'request {method} failed (no result)')
+util.ErrMsg($'request {method} failed (no result): {reply->string()}')
```

And it revealed something curious. This is the error I'm getting:

```
Error: request textDocument/inlayHint failed (no result): 
{'method': 'workspace/configuration', 'jsonrpc': '2.0', 'id': 2, 'params': {'items': [{'section': 'gopls'}]}}
```

So I'm receiving a `workspace/configuration` response in the `textDocument/inlayHint` callback??

---

```
var servers = [
  {
    name: 'gopls',
    path: 'gopls',
    args: ['serve'],
    filetype: ['go', 'gomod', 'gowork', 'gotmpl', 'gohtmltmpl', 'gotexttmpl'],
    syncInit: true,
    debug: true,
    traceLevel: 'verbose',
    rootSearch: ['go.work', 'go.mod'],
    workspaceConfig: {
      gopls: {
        analyses: {
          nilness: true,
          unusedparams: true,
          unusedwrite: true,
          useany: true,
        },
        hoverKind: 'FullDocumentation',
        linksInHover: false,
        gofumpt: true,
        completeUnimported: true,
        # semanticTokens: false,
        staticcheck: true,
        usePlaceholders: true,
        completionDocumentation: true,
        codelenses: {
          generate: true,
          test: true,
          run_vulncheck_exp: true,
        },
        hints: {
          assignVariableTypes: false,
          compositeLiteralFields: true,
          compositeLiteralTypes: false,
          constantValues: true,
          functionTypeParameters: true,
          parameterNames: true,
          rangeVariableTypes: true,
        },
      }
    }
  }
]

var opts = {
  autoComplete: true,
  autoHighlight: true,
  highlightDiagInline: true,
  showDiagOnStatusLine: true,
  showInlayHints: true,
  snippetSupport: true,
  vsnipSupport: true,
}

g:LspAddServer(servers)
g:LspOptionsSet(opts)
```

Shane-XB-Qian:
closes #395

### DIFF
--- a/autoload/lsp/inlayhints.vim
+++ b/autoload/lsp/inlayhints.vim
@@ -95,7 +95,7 @@ def LspInlayHintsUpdate(bnr: number)
     # When running tests, update the inlay hints immediately
     InlayHintsTimerCb(lspserver, bnr, -1)
   else
-    timerid = timer_start(800, function('InlayHintsTimerCb', [lspserver, bnr]))
+    timerid = timer_start(600, function('InlayHintsTimerCb', [lspserver, bnr]))
     setbufvar(bnr, 'LspInlayHintsTimer', timerid)
   endif
 enddef
--- a/autoload/lsp/inlayhints.vim
+++ b/autoload/lsp/inlayhints.vim
@@ -76,7 +76,7 @@ enddef
 
 # Update all the inlay hints.  A timer is used to throttle the updates.
 def LspInlayHintsUpdate(bnr: number)
-  if !bnr->getbufvar('LspInlayHintsNeedsUpdate', true)
+  if !bnr->getbufvar('LspInlayHintsNeedsUpdate', false)
     return
   endif
 
@@ -87,7 +87,7 @@ def LspInlayHintsUpdate(bnr: number)
   endif
 
   var lspserver: dict<any> = buf.BufLspServerGet(bnr, 'inlayHint')
-  if lspserver->empty()
+  if lspserver->empty() || !lspserver.ready
     return
   endif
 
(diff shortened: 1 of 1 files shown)


yegappan:
Based on the comments in #395, it looks like this doesn't really fix that problem.

autoload/lsp/inlayhints.vim
```diff
@@ -87,15 +87,15 @@ def LspInlayHintsUpdate(bnr: number)
   endif
 
   var lspserver: dict<any> = buf.BufLspServerGet(bnr, 'inlayHint')
```
yegappan, line 89: The BufLspServerGet() function already checks whether the LSP server is ready or not.  It returns the LSP server that is ready.  So there is no need to check for the ready flag below.

Shane-XB-Qian:
> Based on the comments in #395, it looks like this doesn't really fix that problem.

no, if no this, regardingless project size, it would be high possibility failed,
seems gopls really was a bit delay (not really ready) to response,
besides this is a refine to inlay hints code anyway,
so this PR is required.

-- 
shane.xb.qian

autoload/lsp/inlayhints.vim
Shane-XB-Qian: yea, but i was thinking it was missing something, or anyway inlay hint was not ready anyway.

Shane-XB-Qian:
thanks to that patch on channel, i reduced the value of delay a bit, should be a balanced value now, wish so.

Shane-XB-Qian:
or let it be, since that channel patch perhaps was better or good enough.

PR #397, opened 2023-09-19
fix: move additionalTextEdits switch

Shane-XB-Qian:
refine #389

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -582,7 +582,6 @@ def LspCompleteDone(bnr: number)
 
   var completionData: any = v:completed_item->get('user_data', '')
   if completionData->type() != v:t_dict
-      || !opt.lspOptions.completionTextEdit
     return
   endif
 
@@ -593,6 +592,7 @@ def LspCompleteDone(bnr: number)
     completionData = lspserver.resolveCompletion(completionData, true)
   endif
   if !completionData->get('additionalTextEdits', {})->empty()
+	\ && opt.lspOptions.completionTextEdit
     textedit.ApplyTextEdits(bnr, completionData.additionalTextEdits)
   endif
 
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -586,6 +586,7 @@ def LspCompleteDone(bnr: number)
   endif
 
   if !completionData->has_key('additionalTextEdits')
+	\ && opt.lspOptions.completionTextEdit
     # Some language servers (e.g. typescript) delay the computation of the
     # additional text edits.  So try to resolve the completion item now to get
     # the text edits.


Shane-XB-Qian:
perhaps this is ok to merge too, it is to make that switch only control the additionalTextedit, and it was for that too, last PR make it lost scope, this is to fix that.

yegappan:
If 'completionTextEdit' option is set, then the expectation is that a snippet completion plugin will expand the snippet and apply the text edits.  So I am not sure the `command` in the completion reply can be applied using codeaction.DoCommand() if this option is set.  Is it possible to test this change with a snippet plugin to verify this?

Shane-XB-Qian:
> If 'completionTextEdit' option is set, then the expectation is that a snippet completion plugin will expand the snippet and apply the text edits.  So I am not sure the `command` in the completion reply can be applied using codeaction.DoCommand() if this option is set.  Is it possible to test this change with a snippet plugin to verify this?

supposed i got your concern: you're worrying that switch should toggle 'additionaltext' with 'command' together?
actually i donnot know but per general sense 'command' is not same scope with 'additionaltext',
and/but most specifical thing is i added this 'switch' was just for 'additionaltext'.
and i didnot see any lsp server combine them, but @vimposter said haskell did?
// BTW: i roughly check vsnip, looks it didnot mention any about 'command', so at least for now seems it is not part of it.

-- 
shane.xb.qian

autoload/lsp/completion.vim
```diff
@@ -593,6 +592,7 @@ def LspCompleteDone(bnr: number)
     completionData = lspserver.resolveCompletion(completionData, true)
   endif
   if !completionData->get('additionalTextEdits', {})->empty()
+	\ && opt.lspOptions.completionTextEdit
```
vimpostor, line 595: Moving the line here seems a bit fishy to me for another reason:

Suppose that `completionTextEdit == false`, then only guarding this if condition may perform redundant work. The delayed `lspserver.resolveCompletion` directly above this if-block might still be executed, but this is completely unnecessary as we won't use the result anywhere.

Shane-XB-Qian: well, simplest change i had said in https://github.com/yegappan/lsp/pull/389#issuecomment-1705627325 (but ignore that cursor issue, you had refined it), if just added and moved 'command' part before 'textedit', then this PR no required and nothing needed change;
i am still not sure what case required such 'command' action at here, could you verify your faced case to verify if it's ok to move that part to there (of course please be sure 'command' if was really required)?

vimpostor: > if just added and moved 'command' part before 'textedit', then nothing needed change

I already told you in my original PR that this is not possible, the command has to be applied after the textedit.

> if it's ok to move that part to there

With the latest commit your PR seems fine to me, although I am not an expert with snippet plugins, as I don't use any.

Shane-XB-Qian: > > if just added and moved 'command' part before 'textedit', then nothing needed change
> 
> I already told you in my original PR that this is not possible, the command has to be applied after the textedit.

yes, but i hear `I am not entirely sure` in that comment.

or now anyway seems only 'haskell' (as you said) worked like this, i have no idea any others, and seems you confirmed it should work like this way, then i think this PR is ready to go.

PR #398, opened 2023-09-21
Fixes duplicate suggestions for servers like volar-server

DanielViberg:
volar-server sends duplicates, items with the same kind and abbr. This addition makes sure there is only one of each instance. 

Note, if there are better properties to use for the comparison please change to those instead.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -296,6 +296,9 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     endif
 
     d.user_data = item
+
+    # Some servers like volar-server sends duplicates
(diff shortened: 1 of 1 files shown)


DanielViberg:
Needs more work on filtering procedure

autoload/lsp/completion.vim
```diff
@@ -296,6 +296,9 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     endif
 
     d.user_data = item
+
+    # Some servers like volar-server sends duplicates
+    completeItems = filter(completeItems, (_, i) => i.abbr != d.abbr && i.kind != d.kind)
```
Shane-XB-Qian, line 301: looks would be busy,
and specially dup was allowed and needed, e.g some snip item had same name and same kind,
or even different kind maybe same name.

PR #405, opened 2023-10-05
fix: send buffer text on textDocument/didSave

mhanberg:
This sends the full buffer text on the textDocument/didSave notification if the server's sync options specify it.

I'm not exactly sure how to unit test this, as I am a Neovim user, but I am attempting to get this LSP client
working with my new LSP for Elixir, [Next LS](https://github.com/elixir-tools/next-ls).

It uses the text property on this notification.

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -907,8 +907,11 @@ def DidSaveFile(lspserver: dict<any>, bnr: number): void
   # Notification: 'textDocument/didSave'
   # Params: DidSaveTextDocumentParams
   var params = {textDocument: {uri: util.LspBufnrToUri(bnr)}}
-  # FIXME: Need to set "params.text" when
-  # 'lspserver.caps.textDocumentSync.save.includeText' is set to true.
+
+  if lspserver.caps.textDocumentSync.save.includeText
+    params.textDocument.text = bnr->getbufline(1, '$')->join("\n") .. "\n"
+  endif
+
   lspserver.sendNotification('textDocument/didSave', params)
 enddef
 


mhanberg:
I tested this locally, seems to work.

autoload/lsp/buffer.vim
```diff
@@ -99,7 +99,7 @@ export def BufLspServerGet(bnr: number, feature: string = null_string): dict<any
 
   # LSP server is configured to be a provider for "feature"
   for lspserver in possibleLSPs
-    var has_feature: bool = lspserver.features->get(feature, false)
+    var has_feature: bool = lspserver.featureEnabled(feature)
```
Shane-XB-Qian, line 102: this is Not same meaning statement, it was going to check feature option.

autoload/lsp/lspserver.vim
```diff
@@ -906,9 +906,12 @@ def DidSaveFile(lspserver: dict<any>, bnr: number): void
 
   # Notification: 'textDocument/didSave'
   # Params: DidSaveTextDocumentParams
-  var params = {textDocument: {uri: util.LspBufnrToUri(bnr)}}
-  # FIXME: Need to set "params.text" when
-  # 'lspserver.caps.textDocumentSync.save.includeText' is set to true.
+  var params: dict<any> = {textDocument: {uri: util.LspBufnrToUri(bnr)}}
+
+  if lspserver.caps.textDocumentSync.save.includeText
```
Shane-XB-Qian, line 911: maybe need to check 'save' was a dict or not, and if it had includeText or not?

Shane-XB-Qian (changes requested):

autoload/lsp/buffer.vim
mhanberg: I think I understand now.

This first loop checks if any features have been _explicitly_ enabled by the user. 

The second loop checks if any default features are enabled?

I will revert this

mhanberg:
@Shane-XB-Qian pushed up changes for your comments, let me know what you think

PR #431, opened 2023-12-31
none: refine code action regex a bit

Shane-XB-Qian:
it maybe messed digit with str

### DIFF
--- a/autoload/lsp/codeaction.vim
+++ b/autoload/lsp/codeaction.vim
@@ -82,12 +82,13 @@ export def ApplyCodeAction(lspserver: dict<any>, actionlist: list<dict<any>>, qu
 
   var choice: number
 
-  if query =~ '^\d\+'	# digit
-    choice = query->str2nr()
-  elseif query =~ '^/'	# regex
-    choice = 1 + util.Indexof(actions, (i, a) => a.title =~ query[1 : ])
-  elseif query != ''	# literal string
-    choice = 1 + util.Indexof(actions, (i, a) => a.title[0 : query->len() - 1] == query)
+  var query_ = query->trim()
+  if query_ =~ '^\d\+$'	# digit
+    choice = query_->str2nr()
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
it maybe messed digit with str

autoload/lsp/codeaction.vim
```diff
@@ -82,12 +82,13 @@ export def ApplyCodeAction(lspserver: dict<any>, actionlist: list<dict<any>>, qu
 
   var choice: number
 
-  if query =~ '^\d\+'	# digit
-    choice = query->str2nr()
-  elseif query =~ '^/'	# regex
-    choice = 1 + util.Indexof(actions, (i, a) => a.title =~ query[1 : ])
-  elseif query != ''	# literal string
-    choice = 1 + util.Indexof(actions, (i, a) => a.title[0 : query->len() - 1] == query)
+  var query_ = query->trim()
```
yegappan, line 85: Instead of introducing another "query_" variable, can you reuse query itself here?

Shane-XB-Qian:
> Instead of introducing another "query_" variable, can you reuse query itself here?

I wish, but it cannot.
maybe it's another vim9 script bug/odd thing.

--
shane.xb.qian

yegappan:
> Instead of introducing another "query_" variable, can you reuse query itself here?
> I wish, but it cannot. maybe it's another vim9 script bug/odd thing.

If you use the 'query' variable, do you get an error?  What is the odd thing that you see?

Shane-XB-Qian:
> > > Instead of introducing another "query_" variable, can you reuse query itself here?
> 
> > I wish, but it cannot. maybe it's another vim9 script bug/odd thing.
> 
> If you use the 'query' variable, do you get an error?  What is the odd thing that you see?

yes, "cannot re-use def args", the err.
// have not you tried, or did not know?

-- 
shane.xb.qian

PR #432, opened 2024-01-02
Convert file path returned from LSP in Cygwin

matmatyk:
Conversion in the other direction is performed below, in `LspFileToUri` function. This direction was missing and this PR adds that.

### DIFF
--- a/autoload/lsp/util.vim
+++ b/autoload/lsp/util.vim
@@ -85,7 +85,12 @@ export def LspUriToFile(uri: string): string
   if uri_decoded =~? '^file:///\a:'
     # MS-Windows URI
     uri_decoded = uri_decoded[8 : ]
-    uri_decoded = uri_decoded->substitute('/', '\\', 'g')
+    if has("win32unix")  # we're in Cygwin
+      # The substitution is to remove the '^@' escape character from the end of line.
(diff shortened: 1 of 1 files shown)


autoload/lsp/util.vim
```diff
@@ -85,7 +85,12 @@ export def LspUriToFile(uri: string): string
   if uri_decoded =~? '^file:///\a:'
     # MS-Windows URI
     uri_decoded = uri_decoded[8 : ]
-    uri_decoded = uri_decoded->substitute('/', '\\', 'g')
+    if has("win32unix")  # we're in Cygwin
+      # The substitution is to remove the '^@' escape character from the end of line.
+      uri_decoded = system($'cygpath --unix {uri_decoded}')->substitute('^\(\p*\).*$', '\=submatch(1)', "")
```
yegappan, line 90: Invoking an external command for decoding every URI may incur a lot of overhead.  Is cygpath used to only the '^@' character from the end of the line?  Is it possible to remove it using trim()?

joseppc:
I think this PR could be closed, this might be fixed in 9a9c548c630982fbb843b18c9dd198f71856ed75 now

yegappan:
Thanks.  Closing this PR.  Please reopen this if this issue is not addressed in the latest version of the plugin.

Issue #433, opened 2024-01-02
Volar server return duplicate completion items


DanielViberg:
Using volar server in the provided test case yields duplicate completion items. 
[test.zip](https://github.com/yegappan/lsp/files/13813661/test.zip)

## How to replicate:
1. Open One.vue
2. Remove ccc variable on line 14
3. Write cc
4. Observe duplicates of completion items

Json encoded completionItems:
[complItems.json](https://github.com/yegappan/lsp/files/13813680/complItems.json)

PR #434, opened 2024-01-02
Added option for removing duplicate completion items

DanielViberg:
Fixes https://github.com/yegappan/lsp/issues/433

 Note, this solution seems to work for volar and php lsp, I'm unsure of how it will behave in other servers.
 Feedback i welcome on what keys should be used when comparing completion items.

### DIFF
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -135,7 +135,10 @@ export var lspOptions: dict<any> = {
   customCompletionKinds: false,
 
   # A dictionary with all completion kinds that you want to customize
-  completionKinds: {}
+  completionKinds: {},
+
+  # Filter duplicate completion items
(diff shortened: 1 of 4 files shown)


autoload/lsp/options.vim
```diff
@@ -135,7 +135,10 @@ export var lspOptions: dict<any> = {
   customCompletionKinds: false,
 
   # A dictionary with all completion kinds that you want to customize
-  completionKinds: {}
+  completionKinds: {},
+
+  # Filter duplicate completion items
+  filterCompletionDuplicates: false,
```
yegappan, line 141: Can you update the documentation with a description about this new option?

DanielViberg: I've added to the lsp-options

PR #437, opened 2024-01-04
none: tidy used item and trim tail spaces

Shane-XB-Qian:
- trim tail spaces
- pick up some to key

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -297,18 +297,16 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     endif
 
     # Dont include duplicate items
-    if lspOpts.filterCompletionDuplicates       
+    if lspOpts.filterCompletionDuplicates
       var key = d->get('word', '') ..
+                d->get('menu', '') ..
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
> Why "abbr" and "dup" fields are removed from the key?

- had 'word' already,
- 'dup' maybe all same val.

-- 
shane.xb.qian

yegappan:
> Why "abbr" and "dup" fields are removed from the key?
> - had 'word' already, - 'dup' maybe all same val.

it is better to keep `abbr` and `dup` to make sure there are no duplicates.
Note that a completion item may have `abbr` in addition to `word`.

girishji:
Please do not remove `dup`. If LSP sends a snippet back and if the `abbr` or `word` matches a buffer word completion or vsnip completion, it will not show.

Shane-XB-Qian:
> Please do not remove `dup`. 

remove where? if you meant in that compl items itself, then it should not and would not.

> If LSP sends a snippet back and if the `abbr` or `word` matches a buffer word completion or vsnip completion, it will not show.

are they same kind? was this really the problem?

PR #441, opened 2024-01-16
fix: hover popup-filter, remove redundant code

Song-Tianxiang:
The original few lines of code are redundant.
Just return false and let vim handle the other keys. The redundant code disabled mouse scrolling, and the popup will be closed automatically after some seconds (I don't know why it is closed automatically)

### DIFF
--- a/autoload/lsp/hover.vim
+++ b/autoload/lsp/hover.vim
@@ -90,11 +90,6 @@ def HoverWinFilterKey(hoverWin: number, key: string): bool
     keyHandled = true
   endif
 
-  if !keyHandled
-    # For any other key, close the hover window
-    hoverWin->popup_close()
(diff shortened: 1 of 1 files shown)


autoload/lsp/hover.vim
```diff
@@ -90,11 +90,6 @@ def HoverWinFilterKey(hoverWin: number, key: string): bool
     keyHandled = true
   endif
 
-  if !keyHandled
-    # For any other key, close the hover window
-    hoverWin->popup_close()
-  endif
-
```
yegappan, line 97: Pressing Escape should close the hover popup window.  With the above change, the escape key will not close the hover window.  An explicit check for `Esc` key should be added to close the popup window.

PR #449, opened 2024-02-06
Add: diagnostic virtual text wrapping mode option

TheDreadedAndy:
There's currently not a way to configure the wrapping mode of diagnostic virtual text. This prevents users from configuring diagnostics like the default in neovim, which is after the line containing the error and truncated when too long.

This PR adds a configuration option that lets the wrapping mode be forced to either 'wrap' or 'truncate', or kept as it was before this PR ('default', the default value, which uses 'wrap' for 'after' and 'truncate' for 'above' and 'below').

### DIFF
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -70,6 +70,10 @@ export var lspOptions: dict<any> = {
   # Allowed values: 'above' | 'below' | 'after' (default is 'above')
   diagVirtualTextAlign: 'above',
 
+  # Wrapping of virtual diagnostic text, when showDiagWithVirtualText is true.
+  # Allowed valuse: 'default' | 'truncate' | 'wrap' (default is 'default')
+  diagVirtualTextWrap: 'default',
+
   # Suppress adding a new line on completion selection with <CR>
   noNewlineInCompletion: false,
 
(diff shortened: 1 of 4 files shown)


autoload/lsp/options.vim
```diff
@@ -70,6 +70,10 @@ export var lspOptions: dict<any> = {
   # Allowed values: 'above' | 'below' | 'after' (default is 'above')
   diagVirtualTextAlign: 'above',
 
+  # Wrapping of virtual diagnostic text, when showDiagWithVirtualText is true.
+  # Allowed valuse: 'default' | 'truncate' | 'wrap' (default is 'default')
+  diagVirtualTextWrap: 'default',
```
yegappan, line 75: Can you update the help text with a description of this new option and also the README file?

TheDreadedAndy: Sure. I'll post here once I've got that finished.

TheDreadedAndy:
Just a note: It seems vims virtual text behavior with 'after' + 'truncate' is a little odd. It will keep appending to the line until the first text must be truncated, then any future text will go on the next line until it needs to be truncated, etc.

Example:
```
def foo(): E> An error E> Another err...
E> A third error E> A long fourth err...
```

autoload/lsp/options.vim
TheDreadedAndy: I've committed the documentation changes to my fork.

PR #475, opened 2024-03-13
bug: 'Lazy doc' completion broken for omnifunc

girishji:
pylsp was not displaying documentation from lsp server. this fixes the problem.
user does not have to enable omnicomplete for each file type when main option omniComplete is already enabled.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -641,7 +641,7 @@ export def BufferInit(lspserver: dict<any>, bnr: number, ftype: string)
     return
   endif
 
-  if !opt.lspOptions.autoComplete && !LspOmniComplEnabled(ftype)
+  if !opt.lspOptions.autoComplete && !LspOmniComplEnabled(ftype) && !opt.lspOptions.omniComplete
     # LSP auto/omni completion support is not enabled for this buffer
     return
   endif


girishji:
Thanks!

autoload/lsp/completion.vim
```diff
@@ -641,7 +641,7 @@ export def BufferInit(lspserver: dict<any>, bnr: number, ftype: string)
     return
   endif
 
-  if !opt.lspOptions.autoComplete && !LspOmniComplEnabled(ftype)
+  if !opt.lspOptions.autoComplete && !LspOmniComplEnabled(ftype) && !opt.lspOptions.omniComplete
```
Shane-XB-Qian, line 644: tho not sure your problem, but the code logic seems wrong:
- when `lspOptions.omniComplete` was true, 
- but `server.omnicompl` was false,
- then it should be still no omni compl there.
a.k.a when auto compl was false too, then that judgement should be `return` still.

Shane-XB-Qian (commented):
i am not going to raise another PR, please try to refine it by yourself. thanks.

autoload/lsp/completion.vim
```diff
@@ -641,7 +641,7 @@ export def BufferInit(lspserver: dict<any>, bnr: number, ftype: string)
...
     # LSP auto/omni completion support is not enabled for this buffer
     return
```
Shane-XB-Qian, line 646: and this comment.

autoload/lsp/completion.vim
Shane-XB-Qian: a.k.a when auto compl was false too, then ~~that judgement should be return still.~~ it already like its next comment said.

girishji:
You can refactor any way you want as long as we have these when `opt.lspOptions.omniComplete==true`

```
  if lspserver.completionLazyDoc
    # resolve additional documentation for a selected item
    acmds->add({bufnr: bnr,
                event: 'CompleteChanged',
                group: 'LSPBufferAutocmds',
                cmd: 'LspResolve()'})
  endif

  acmds->add({bufnr: bnr,
	      event: 'CompleteChanged',
	      group: 'LSPBufferAutocmds',
	      cmd: 'LspSetPopupFileType()'})

  # Execute LSP server initiated text edits after completion
  acmds->add({bufnr: bnr,
	      event: 'CompleteDone',
	      group: 'LSPBufferAutocmds',
	      cmd: $'LspCompleteDone({bnr})'})


```

girishji:
Or better still, export LspSetPopupFileType(), LspCompleteDone(), LspResolve().

Shane-XB-Qian:
> You can refactor any way you want as long as we have these when `opt.lspOptions.omniComplete==true`

i am not interesting, 
if you can confirm there no outage to set those completion settings when it should `return`, 
then please just re-word that comment, 
otherwise please try to fix your issue by "another" way.

Shane-XB-Qian:
https://github.com/yegappan/lsp/pull/426#discussion_r1432741825
https://github.com/yegappan/lsp/pull/428#issuecomment-1868244111

BTW: i still remember last time you modified that `kind`.

girishji:
> [#426 (comment)](https://github.com/yegappan/lsp/pull/426#discussion_r1432741825) [#428 (comment)](https://github.com/yegappan/lsp/pull/428#issuecomment-1868244111)
> 
> BTW: i still remember last time you modified that `kind`.

`kind` was the correct modification. If LSP server sends you an unexpected response, LSP client should deal with it gracefully. This is not open to debate.

Shane-XB-Qian:
> `kind` was the correct modification. If LSP server sends you an unexpected response, LSP client should deal with it gracefully. This is not open to debate.

there is a spec, and the `kind` value is 1~25 only, client no duty to handle all broken from some server.
wish you were reducing the possibilities of adding odd or wrong code here.

PR #482, opened 2024-03-18
Add LspDiag nextWrap and prevWrap command args

### LINKED ISSUE #413: LspNextDiagnostic at end of diagnostics
Hello, I would like to be able to find the next diagnostic after the last 
(like when vim continues search from top of the file). Can it be done?

nickspoons:
Add new `'nextWrap'` and `'prevWrap'` arguments for the `:LspDiag` command, allowing wrapping around to the first/last diagnostics instead of displaying the "No more diagnostics found" message.

Closes #413


yegappan:
Thanks for the patch.  Can you fix the following test failure?

```
command line..script /home/runner/work/lsp/lsp/test/runner.vim[48]..function <SNR>1_LspRunTests[20]..Test_LspDiagsSubcmd
line 4: Expected 'LspDiag first current here highlight last next prev show' but got 'LspDiag first current here highlight last next nextWrap prev prevWrap show'
Test_LspDiagsSubcmd(): FAIL

```

nickspoons:
Hi @yegappan, I've fixed the test by adding the missing args to the assertion.

test/clangd_tests.vim
```diff
@@ -1625,7 +1625,7 @@ def g:Test_LspDiagsSubcmd()
   new XLspDiagsSubCmd.raku
 
   feedkeys(":LspDiag \<C-A>\<CR>", 'xt')
-  assert_equal('LspDiag first current here highlight last next prev show', @:)
+  assert_equal('LspDiag first current here highlight last next nextWrap prev prevWrap show', @:)
   feedkeys(":LspDiag highlight \<C-A>\<CR>", 'xt')
```
yegappan, line 1629: Can you add new tests for the nextWrap and prevWrap functionality?

yegappan (changes requested):

test/clangd_tests.vim
nickspoons: I've added some commands and assertions to the existing `Test_LspDiag_Multi()` test, is that OK or do you want separate test functions?

PR #495, opened 2024-04-04
fix: wrap `didSaveFile` method with `try-finally` block

maximyurevich:
Solves issue: https://github.com/yegappan/lsp/issues/

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -335,7 +335,14 @@ def LspSavedFile(bnr: number)
   endif
 
   for lspserver in lspservers
-    lspserver.didSaveFile(bnr)
+    # TODO: implement `catch` block
+    # Wrap method with `try-finally` block to solve error: 
+    # `E716: Key not present in Dictionary: "supportsDidSave"`
+    try
+      lspserver.didSaveFile(bnr)
+    finally
+      return
+    endtry
   endfor
 enddef
 


autoload/lsp/lsp.vim
```diff
@@ -335,7 +335,14 @@ def LspSavedFile(bnr: number)
   endif
 
   for lspserver in lspservers
-    lspserver.didSaveFile(bnr)
+    # TODO: implement `catch` block
+    # Wrap method with `try-finally` block to solve error: 
+    # `E716: Key not present in Dictionary: "supportsDidSave"`
```
Shane-XB-Qian, line 340: how this would happen?

autoload/lsp/lsp.vim
```diff
@@ -335,7 +335,14 @@ def LspSavedFile(bnr: number)
...
+    try
+      lspserver.didSaveFile(bnr)
+    finally
+      return
```
Shane-XB-Qian, line 344: even it happened, just `return` or should `continue`?

Shane-XB-Qian (commented):
donot understand what this PR really solved, or worse.

autoload/lsp/lsp.vim
maximyurevich: I didn't dive into the entire code base, I just hid the errors that appeared during programming. Perhaps it would be worthwhile to implement the ability to hide errors using a variable.

Shane-XB-Qian: though @yegappan accepted your/this PR, but i think:
1. this didnot fix (or hid) the real err, since it perhaps was an init case, tho not sure how come that happened.
2. and perhaps worse, since if that really happened, i thought it should `continue` to next lspserver vs `return`.

PR #496, opened 2024-04-06
Add multiple language server association with syntax elements

indelog:
Improve multi-language server support by enabling the selection of the appropriate LSP based on syntax stack elements using a new configuration option for the LSP: `syntaxAssociatedLSP`, which accepts a list of syntax names obtainable through:

```vim
synstack(line, col)->map((_, v) => v->synIDattr('name'))
```

This enhancement allows by example, the simultaneous use of a PHP and HTML LSP in a PHP file when an LSP function is triggered. If the syntax stack contains an `htmlTag` pattern, the HTML LSP is called; otherwise, the PHP LSP is utilized. Furthermore, this improvement supports cases like embedding GraphQL blocks in TypeScript files.

Example usages include situations like the following configurations:

```vim
    call LspAddServer([
                \{
                \    'name': 'phpactor',
                \    'filetype': ['php'],
                \    'path': expand('~') . '/.local/bin/phpactor',
                \    'args': ['language-server'],
                \},
                \{
                \    'name': 'htmlls',
                \    'filetype': ['html', 'php'],
                \    'path': l:lib_path . 'node_modules/.bin/html-languageserver',
                \    'args': ['--stdio'],
                \    'initializationOptions': {'embeddedLanguages': {'css': v:true, 'javascript': v:true}},
                \    'syntaxAssociatedLSP': ['htmlTag', 'htmlError'],
                \    'runUnlessSearch': [
                \      'angular.json',
                \    ],
                \},
                \{
                \    'name': 'typescript-language-server',
                \    'filetype': ['typescript', 'javascript'],
                \    'path': l:lib_path . '/node_modules/.bin/typescript-language-server',
                \    'args': ['--stdio'],
                \},
                \{
                \    'name': 'graphql-language-server',
                \    'filetype': ['typescript', 'javascript', 'graphql'],
                \    'path': l:lib_path . '/node_modules/.bin/graphql-lsp',
                \    'args': ['server', '-m', 'stream'],
                \    'initializationOptions': {'diagnostics': 'true'},
                \    'syntaxAssociatedLSP': ['graphqlTemplateString'],
                \},
                \])
```

While acknowledging that this approach is relatively straightforward, I believe it remains effective for conventional use cases as long as the language server is designed to operate within the context of the current file.

### DIFF
--- a/.github/workflows/unitests.yml
+++ b/.github/workflows/unitests.yml
@@ -12,23 +12,26 @@ jobs:
     steps:
       - name: Install packages
         run: |
-          sudo apt update
+          sudo apt-get update
           # install clangd language server
-          sudo apt install -y clangd-15
+          sudo apt-get install -y clangd-15
           # install nodejs
-          sudo apt install -y curl
+          sudo apt-get install -y curl
           curl -fsSL https://deb.nodesource.com/setup_18.x | sudo bash -
-          sudo apt install -y nodejs
+          sudo apt-get install -y nodejs
           # install the typescript language server
           sudo npm install -g typescript-language-server typescript
           # install the golang language server
-          sudo apt install -y golang
-          sudo apt install -y gopls
+          sudo apt-get install -y golang
+          sudo apt-get install -y gopls
           # install the rust language server
-          sudo apt install -y cargo rust-src
+          sudo apt-get install -y cargo rust-src
           mkdir -p ~/.local/bin
           curl -L https://github.com/rust-lang/rust-analyzer/releases/latest/download/rust-analyzer-x86_64-unknown-linux-gnu.gz | gunzip -c - > ~/.local/bin/rust-analyzer
           chmod +x ~/.local/bin/rust-analyzer
+          # install markdown language server
+          curl -L https://github.com/artempyanykh/marksman/releases/latest/download/marksman-linux-x64 -o ~/.local/bin/marksman
+          chmod +x ~/.local/bin/marksman
       - name: Setup Vim
         uses: rhysd/action-setup-vim@v1
         id: vim
--- a/autoload/lsp/buffer.vim
+++ b/autoload/lsp/buffer.vim
@@ -106,6 +106,12 @@ export def BufLspServerGet(bnr: number, feature: string = null_string): dict<any
     endif
   endfor
 
+  # The LSP is explicitly associated to a specific syntax name within the syntax stack
+  var selectedLsp = SelectLSPBySyntaxNames(possibleLSPs)
+  if (!selectedLsp->empty())
+    return selectedLsp
+  endif
+
   # Return the first LSP server that supports "feature" and doesn't have it
   # disabled
   for lspserver in possibleLSPs
@@ -189,6 +195,49 @@ export def CurbufGetServerChecked(feature: string = null_string): dict<any>
   endif
 
   return lspserver
+enddef
+
+# Returns the selected LSP based on the syntax names stacked under the
+# current cursor position
+def SelectLSPBySyntaxNames(possibleLSPs: list<dict<any>>): dict<any>
+  var synnameStack = util.ListSynstackNamesAtPoint(line('.'), col('.'))->reverse()
+
+  if synnameStack->empty()
+    return {}
+  endif
+
+  # Initialize variables for tracking the selected LSP and the index of the
+  # matched word
+  # The syntax word statck is revesed so the word at a lower index is deeper
+  # in the syntax stack : use it in priority
+  var synWordIdx = 1000
+  var selected = {}
+
+  for server in possibleLSPs
+    if server.syntaxAssociatedLSP->empty()
+      continue
+    endif
+
+    # Loop through each syntax name in the stack
+    for idx in range(len(synnameStack))
+      # Skip this syntax name if it's not in the list of associated syntax
+      # names for this LSP
+      if server.syntaxAssociatedLSP->index(synnameStack[idx]) < 0
+        continue
+      endif
+
+      # Update the selected LSP and the index of the matched word if the
+      # syntax name has higher priority
+      if idx < synWordIdx
+        selected = server
+        synWordIdx = idx
+	# Break out of the loop once the LSP has been selected
+        break
+      endif
+    endfor
+  endfor
+
+  return selected
 enddef
 
 # vim: tabstop=8 shiftwidth=2 softtabstop=2
--- a/doc/lsp.txt
+++ b/doc/lsp.txt
@@ -1900,6 +1900,44 @@ everything else: >
 		},
 	])
 <
+Language servers can also be configured to associate themselves with specific
+syntax elements. If a syntax element is detected at the current location, the
+corresponding language server will take precedence over the standard ordering.
+
+To specify a language server for a particular syntax element, use the
+`syntaxAssociatedLSP` property in the configuration object passed to
+`LspAddServer`. The value of `syntaxAssociatedLSP` should be a string or list
+of strings representing the desired syntax elements.
+>
+	vim9script
+
+	# Add two LSP configurations for handling JavaScript, TypeScript, and
+	# GraphQL files.
+	g:LspAddServer([
+		{
+			# Configuration for Typescript-language-server:
+			# Used for general purposes when working with '.js'
+			# and '.ts' files.
+			filetype: ['javascript', ''typescript'],
+			path: 'typescript-language-server',
+			args: ['--stdio']
+		},
+		{
+			# Configuration for GraphQL-language-server:
+			# Specifically bound to 'graphqlTemplateString' syntax
+			# element. Will handle requests regarding GraphQL
+			# template literals.
+			filetype: ['javascript', ''typescript', 'graphql'],
+			path: 'graphql-lsp',
+			args: ['server', '-m', 'stream'],
+			syntaxAssociatedLSP: ['graphqlTemplateString'],
+		}
+	])
+
+<
+To discover the current syntax stack and determine the appropriate value for
+`syntaxAssociatedLSP`, you can employ `LspUtilGetCurrentSynStack` command. 
+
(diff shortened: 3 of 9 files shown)


.github/workflows/unitests.yml
```diff
@@ -12,20 +12,20 @@ jobs:
     steps:
       - name: Install packages
         run: |
-          sudo apt update
+          sudo apt-get update
```
Shane-XB-Qian, line 15: seems unnecessary?

autoload/lsp/buffer.vim
```diff
@@ -170,6 +176,7 @@ export def CurbufGetServerChecked(feature: string = null_string): dict<any>
     return {}
   endif
 
+
```
Shane-XB-Qian, line 179: redundant blank line.

doc/lsp.txt
```diff
@@ -1900,6 +1900,44 @@ everything else: >
...
+		},
+		{
+			# Configuration for GraphQL-language-server:
+			# Specifically bound to 'graphqlTemplateString' syntax
+			# element. Will handle requests regarding GraphQL
+			# template literals.
+			filetype: ['javascript', ''typescript', 'graphql'],
+			path: 'graphql-lsp',
+			args: ['server', '-m', 'stream'],
+			syntaxAssociatedLSP: ['graphqlTemplateString'],
```
Shane-XB-Qian, line 1933: how reliable of this? and need to dyn change/config the syntax value?

.github/workflows/unitests.yml
```diff
@@ -36,6 +36,12 @@ jobs:
           version: ${{ matrix.vim }}
       - name: Checkout LSP plugin Code
         uses: actions/checkout@v4
+      - name: Prepare Tests
+        run: |
+          # install deps for the dummy lsp
+          cd ./test/dummy-lsp/
+          npm install
+          cd -
```
Shane-XB-Qian, line 44: perhaps just using markdown lsp? vs/but not need to make up a dummy lsp.

Shane-XB-Qian:
thanks, seems a good try, 
but not sure how reliable was it, and need to dyn change/config the syntax value? seems a little not easy to use.

BTW: truly there now still some problem so far no good answer, e.g:
1. like this one, how to easily handle embedded code.
2. some lsp handle for same ft, e.g yaml and ansible was both to 'yml' ft.

.github/workflows/unitests.yml
indelog: This is a minor fix. Using the apt command in scripts is discouraged, and it is recommended to use apt-get instead. According to the apt manual:

> The apt(8) commandline is designed as an end-user tool and its behavior might change between versions. Although it aims to maintain backward compatibility, this is not guaranteed if a modification appears advantageous for interactive use.

That's why a warning message is issued when apt is employed in a script.

However, this fix can be safely ignored; nevertheless, there isn't a reason to ignore it.

doc/lsp.txt
indelog: I don't foresee a situation where we need to dynamically change the syntax. Can you provide an example to help me understand what you mean?

.github/workflows/unitests.yml
indelog: I merely wanted a simple LSP setup for a very basic use case, keeping the unit tests equally simple.

The dummy LSP can be configured to return varying responses to hover requests, thus serving as a convenient way to mimic different LSPs using just one. Without this approach, you would have had to rely on two distinct LSP implementations—for example, starting with a PHP file, configuring both `phpactor` and `html-languageserver`, and checking whether the correct LSP responds upon querying an `htmlTag`. I find this complicates the tests unnecessarily.

Should implementing a fake LSP for simple mocking in tests prove problematic, replacing it with actual LSP instances wouldn't be an issue.

Do note that the dummy LSP could potentially be expanded for conducting additional tests. This idea serves as a suggestion rather than a requirement.

indelog:
With this proposal, I do not claim to fully address the complex challenge of multi-LSP, but rather aim to enhance existing options with a simple solution built upon Vim's inherent syntax identification capabilities.

This works smoothly in cases where the LSP is designed to operate within the given context. As demonstrated, the proposed solution excels in cases like PHP files with embedded HTML or TypeScript files featuring GraphQL templates. By maintaining the primary LSP as the default and placing it first in the LSP list, complementary LSPs are included only when encountered within a specific syntax group. Notably, the script processes the complete syntax stack of the current word, starting from the highest level.

Admittedly, cases similar to the YAML Ansible scenario cannot be resolved solely through the suggested approach, as decisions concerning the appropriate LSP cannot depend exclusively on file syntax. In such circumstances, tradeoffs must be made, leaning on other configuration options like `runIfSearch` or `features`. For instance, choosing the Angular LSP over the HTML LSP requires the following configuration:

```vim
    call LspAddServer([
                \{
                \    'name': 'angular',
                \    'filetype': ['html'],
                \    'path': g:lib_path . 'node_modules/.bin/ngserver',
                \    'args': [
                \      '--stdio',
                \      '--tsProbeLocations',
                \      g:lib_path . 'node_modules/typescript/lib/',
                \      '--ngProbeLocations',
                \      g:lib_path . 'node_modules/@angular/language-server/bin/',
                \    ],
                \    'initializationOptions': {'diagnostics': 'true'},
                \    'runIfSearch' : [
                \      'angular.json',
                \    ],
                \},
                \{
                \    'name': 'htmlls',
                \    'filetype': ['html', 'php'],
                \    'path': g:lib_path . 'node_modules/.bin/html-languageserver',
                \    'args': ['--stdio'],
                \    'initializationOptions': {'embeddedLanguages': {'css': v:true, 'javascript': v:true}},
                \    'syntaxAssociatedLSP': ['htmlTag', 'htmlError'],
                \    'runUnlessSearch': [
                \      'angular.json',
                \    ],
                \},
                \])

```

.github/workflows/unitests.yml
Shane-XB-Qian: it has pipeline here, maybe it's ok.

.github/workflows/unitests.yml
Shane-XB-Qian: you use a markdown file which embedded clang/golang etc code block to test, which those had been config into this test, if so, the code base would be clean and easy.

doc/lsp.txt
Shane-XB-Qian: i meant the syntax value was not reliable, or dynamic, perhaps if there was a way to auto detect the syntax value (which it belonging) would be great, otherwise e.g markdown file, there perhaps embedded some unconcern code block.

Shane-XB-Qian:
> With this proposal, I do not claim to fully address the complex challenge of multi-LSP, but rather aim to enhance existing options with a simple solution built upon Vim's inherent syntax identification capabilities.

i know, so i am just discussing, not against it, but just the detection on syntax val seems manually seems was not reliable, or IF it would be auto then would be great.

as for 'ansible', there was no fixed file like `'angular.json'`, but shared yaml lsp with same ft, that's another knew problem.

autoload/lsp/buffer.vim
Shane-XB-Qian: i saw your commit, as i knew yegappan would not squash the commits when merging, maybe you need to squash it later, otherwise you would see the commits history (and tho the contribute count too) a little redundant too, :smile:
// just an irrelevant comment, and it's on yegappan.

.gitignore
```diff
@@ -6,3 +6,4 @@ test/X*.cpp
 # test/Xtest.c
 # test/Xtest.cpp
 test/results.txt
+test/dummy-lsp/node_modules/
```
Shane-XB-Qian, line 9: if so, no such dir or not req now.

.github/workflows/unitests.yml
indelog: I modified the unit tests according to recommendations, removing the dummy LSP. Although these tests aren't as extensive, they ensure the right LSP is invoked, which should be enough. Testing real LSP execution isn't feasible as clangd doesn't understand Markdown, and ultimately, testing real LSP execution turns out to be unnecessary because selecting the appropriate LSP suggests that everything else will most likely fall into place.

indelog:
Manually configuring the syntax may seem non-optimal. But this can have certain advantages, notably allowing the user to adapt their configuration as they wish without having to worry about whether the LPS plugin supports the scenario or not.

On the other hand, this can in fact sometimes be tricky. I actually got trapped with the new unit tests because on my machine a C language block is recognized with the syntax name `markdownHighlight_c` whereas with the version of vim used by the CI is `markdownHighlightc`. This is why I added the small utility command `LspUtilGetCurrentSynStack` to help find which syntax name used for `syntaxAssociatedLSP`.

Shane-XB-Qian:
> Any idea suggested?

i like your 'try', but so far no good answer, wish there some more convenience or reliable way.

jclsn:
Would this fix https://github.com/yegappan/lsp/issues/521 ?

briancarbone:
@indelog It's stating there are merge conflicts, but I'm curious whether this PR would make it easier to have tailwindcss's LS and typescript's LS coexist. Would you rebase it?

briancarbone:
I rebased master unto this here: https://github.com/briancarbone/lsp/tree/syntax-compl-layering.

Overall, the intention behind this PR is really smart. Not all features of LSP can be merged or layered, but diagnostics merging and completion layering makes a lot of sense. This PR in particular makes completion layering explicit by syntax region, allowing me to, for example, set TailwindCSS to take over completions within `tsxString` (which I demonstrated in a video using the branch linked above). The concept of surgically poking holes through to an underlying language server for specific syntax regions does seem superior, upon preliminary consideration, to a more generalized merging where all results are collated.

@yegappan Do you think @indelog's implementation makes sense for completion layering? I'm using this PR and it's solved a major pain point for me.

https://github.com/user-attachments/assets/108108d5-98c6-447d-ac98-af7243ef3e4a

PR #499, opened 2024-04-07
fix: 'GetLineByteFromPos' maybe return val overflow

Shane-XB-Qian:
// supposed it is ok to cover like that..


Shane-XB-Qian:
this overflow maybe a bug from specific server, so perhaps let it be...

Shane-XB-Qian:
um.. i changed my mind, perhaps keep this open whatever it would be going, in case it do required (to fix)...

Shane-XB-Qian:
test was flaky, seems not regarding to this change.

Shane-XB-Qian:
image

PR #534, opened 2024-06-26
Fix crash with neocmakelsp + ale

vimpostor:
This fixes the following error by falling back to `info` severity:

```
Error detected while processing function <SNR>66_Output_cb[5]..lsp#handlers#ProcessMess
ages[32]..lsp#handlers#ProcessNotif[52]..<SNR>67_ProcessDiagNotif[2]..lsp#diag#DiagNoti
fication[72]..lsp#diag#ProcessNewDiags[5]..<SNR>69_SendAleDiags[14]..<lambda>46:
line    6:
E716: Key not present in Dictionary: "severity"
```


Tachi107:
> This fixes the following error by falling back to `info` severity

According to the LSP specification, the default should be `Error`:

> The diagnostic's severity. To avoid interpretation mismatches when a server is used with different clients it is highly recommended that servers always provide a severity value. If omitted, it’s recommended for the client to interpret it as an Error severity.

PR #540, opened 2024-07-24
Use a copy of the server list when restarting

### LINKED ISSUE #378: lsp server re-start/re-effect
// continue for https://github.com/yegappan/lsp/issues/368#issuecomment-1659605692 :

```log
c, if you have time, [#338](https://github.com/yegappan/lsp/issues/338) maybe more valuable
   // though not only py env dyn change issue,
   // e.g go.mod (and/or its import pkg) changed, seems re-effect the changes seems do was an issue,
   // 'lspserver restart' seems not help some.
   // i was observing as well, no clue so far.
```

seems after lsp restart the server[s] didnot really restart (via checking the pid)
but simple cases to repro is: 
1. an opened py file if a pkg did not install (via pip) but imported, then/after installed and lsp restart, you will find env did not take effect still.
2. an opened go file if its imported pkg was not completed, then/after installed and update go.mod, and even lsp restart, you would find it may not take effect.

i think IF no chance/possible to dynamicly take effect, at least make lsp restart work.
// just a mark, @yegappan or someone else if you have time, take a look & check further.

nickspoons:
If the original server list is used, it is cleared during the step of removing filetype buffers, so no servers are actually restarted, as the `lspservers` list is empty by the time the `for` loop is reached.

Fixes #378

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -543,7 +543,7 @@ enddef
 
 # Restart the LSP server for the current buffer
 def RestartServer()
-  var lspservers: list<dict<any>> = buf.CurbufGetServers()
+  var lspservers: list<dict<any>> = buf.CurbufGetServers()->copy()
   if lspservers->empty()
     util.WarnMsg($'No Lsp servers found for "{@%}"')
     return


Shane-XB-Qian:
thx, 
by the way: if so, do you think the 'copy()' in `RemoveFile()` func still required?

nickspoons:
> do you think the 'copy()' in `RemoveFile()` func still required?

I don't know, I tried removing it and it doesn't seem to have broken anything, but I haven't tried to exhaustively chase down code paths and make sure it doesn't cause problems elsewhere. Probably safer just to leave it for now.

Shane-XB-Qian:
> > do you think the 'copy()' in `RemoveFile()` func still required?
> 
> I don't know, I tried removing it and it doesn't seem to have broken anything, but I haven't tried to exhaustively chase down code paths and make sure it doesn't cause problems elsewhere. Probably safer just to leave it for now.

yes, per your PR, i was reading it, then found that 'copy()' in `RemoveFile()` was weird, i donot understand why it was there or any useful.. :shrug:
// maybe it was a copy/paste or refactor code, then move the 'copy()' to wrong place.

-- 
shane.xb.qian

PR #546, opened 2024-08-10
add lsp format operator mapping

Konfekt:
As convenience complementing, say `xnoremap  gw  :LspFormat<cr>`

### DIFF
--- a/plugin/lsp.vim
+++ b/plugin/lsp.vim
@@ -74,7 +74,20 @@ command! -nargs=0 -bar LspDiagShow lsp.ShowDiagnostics()
 command! -nargs=0 -bar LspDiagHere lsp.JumpToDiag('here')
 command! -nargs=0 -bar LspDocumentSymbol lsp.ShowDocSymbols()
 command! -nargs=0 -bar LspFold lsp.FoldDocument()
+
 command! -nargs=0 -bar -range=% LspFormat lsp.TextDocFormat(<range>, <line1>, <line2>)
+function! LspFormatFunc(type, ...) abort
+  if a:0
+    " Visual mode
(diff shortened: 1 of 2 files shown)


plugin/lsp.vim
```diff
@@ -74,7 +74,19 @@ command! -nargs=0 -bar LspDiagShow lsp.ShowDiagnostics()
...
 command! -nargs=0 -bar -range=% LspFormat lsp.TextDocFormat(<range>, <line1>, <line2>)
+def LspFormatFunc(type: string, visual_mode = v:false)
+  if visual_mode
+    exe "normal! gv:LspFormat\<cr>"
+  elseif type ==# 'line'
+    exe "normal! '[V']:LspFormat\<cr>"
+  elseif type ==# 'char'
+    exe "normal! `[v`]:LspFormat\<cr>"
+  endif
+enddef
```
Shane-XB-Qian, line 87: seems a bit mess, are you sure the logic was correct?
btw: the range format normally required lsp server gave it.
vs to local, not sure if it was ok, or if you made it really work, maybe ok too? :shrug:

Konfekt: Works for me and is a suggestion. Feel free to improve, I am stuck with Legacy Vim script for the moment in view of the divergent eco system. Likely the function has to go into autoload, if ever accepted.

Shane-XB-Qian: there is lsp#lsp#FormatExpr() already, though req lsp server gave the range format.

Konfekt: In fact, I am not sure if the lsp server understands the range correctly yet; for example, formatting a single line of a comment with clangd formats the whole comment

Issue #113, opened 2022-11-15
textDocument/didChange does not follow the server sync mode


andirady:
The lsp plugin always submit in Full mode although the server responded with textDocumentSync Incremental.

See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentSyncKind

yegappan:
This support has now been added in 53b3cfa9fba3f2a08d4921424ec6b4623b047997.  This can be enabled by setting the `incrementalSync` option to `true`.

PR #551, opened 2024-08-21
Introduce incremental doc sync

DanielViberg:
Introduced line-based document synchronization during didChange request.

### DIFF
--- a/.github/workflows/unitests.yml
+++ b/.github/workflows/unitests.yml
@@ -27,7 +27,8 @@ jobs:
           # install the rust language server
           sudo apt install -y cargo rust-src
           mkdir -p ~/.local/bin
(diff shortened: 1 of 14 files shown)


DanielViberg:
Looking into the failed tests

DanielViberg:
Fix https://github.com/yegappan/lsp/issues/113
When i sort out the tests

DanielViberg:
Not sure why rust_tests fails on WaitForDiags line 11: Expected 1 but got 0.

pusewicz:
@DanielViberg @yegappan Is there anything that we can do to push this forward?

DanielViberg:
There are cases where the state is not correctly sent to the server. Trying to find the root cause/edge case.

PR #556, opened 2024-09-08
Add missing "g:" to call to LspAddServer

castedo:
Simple typo I needed to fix to make it work.


castedo:
Although ... perhaps
```
call LspAddServer(lspServers)
```
would be more understandable. I'm new to vimscript9 so I'm not sure what style is better.

Issue #553, opened 2024-08-29
Auto Completion only on triggerChar


felixendres:
Hi.
Thanks for the plugin, it's awesome!

I would like to configure completion such that it is only triggered automatically when I type a trigger character (e.g. ".", ">" for C++). If I need completions at other times, I can use C-X C-O.

I think this cannot be done with the current configuration options. I can get the behavior I want by modifying (or adding another) [check for triggerKind after this line](https://github.com/yegappan/lsp/blob/89d5350d756a9bd19bfc3ac0e30c53228e0b28d6/autoload/lsp/completion.vim#L478).

If you are fine with it, I'd like to contribute this as a PR:
- Create a new option "autoCompleteOnTriggerCharOnly" with a default of v:false
- Create a check after the linked line to return without completion if the above option is true but the trigger kind is something other than 2.
- Document the new option

Would you merge that? Is there anything else I need to consider?
Best regards.

Shane-XB-Qian:
that part of code, which that func to get trigger kind maybe not complete correct, since now it not only kind 1 or 2, but also 3, so even you added such feat or flag, the result maybe weird still.

Shane-XB-Qian:
that part of code, which that func to get trigger kind maybe not complete
correct, since now it not only kind 1 or 2, but also 3, so even you added such
feat or flag, the result maybe weird still.

-- 
shane.xb.qian

Shane-XB-Qian:
i meant you either refine that func (and related others code) to make it work fine with all kinds, or just disable auto-compl, then using omni-compl for all cases including trigger char cases.

Shane-XB-Qian:
seems you are mis-understanding, i meant `trigger kind` was not only 1, 2 but also 3 now per spec.
// if you cannot refine that, seems that PR was unnecessary.

-- 
shane.xb.qian

Shane-XB-Qian:
> Why do you mention trigger kind  3?

Because you mentioned in the OP:
```
get the behavior I want by modifying (or adding another) [check for triggerKind after this line](https://github.com/yegappan/lsp/blob/89d5350d756a9bd19bfc3ac0e30c53228e0b28d6/autoload/lsp/completion.vim#L478)
```

> It doesn't exist [in GetTriggerAttributes()](https://github.com/yegappan/lsp/blob/master/autoload/lsp/completion.vim#L446-L466) on the master branch

it is in the spec.


// fine, that's it.

-- 
shane.xb.qian

PR #557, opened 2024-09-12
Add options to limit auto completion and buffer completion

felixendres:
Add option to limit usage of buffer/buffer completion

Auto completion can now be set to be triggered on specific trigger types. This is useful to auto-complete on trigger characters only, e.g. when typing a dot to access a class member, or a colon for a namespaced symbol. See documentation of option autoCompleteTriggers in options.vim.

When completing a file name or a class method, buffer completion makes the list longer and provides little value. The option bufferCompletionTriggers allows to turn buffer completion of for specific trigger types.

The first part (auto completion) was discussed in #553. I made the PR accordingly.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -182,7 +182,7 @@ export def CompletionReply(lspserver: dict<any>, cItems: any, trigger_type: numb
     snippet.CompletionVsnip(items)
   endif
 
-  if lspOpts.useBufferCompletion
+  if lspOpts.useBufferCompletion && count(lspOpts.bufferCompletionTriggers, trigger_type) > 0
     CompletionFromBuffer(items)
   endif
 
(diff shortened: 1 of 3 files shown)


Shane-XB-Qian:
had replied at OP ticket, if you just want to toggle that buf/auto-compl, there are options can do that already.
// so supposed you donot need to dup the wheel or mess the code.

-- 
shane.xb.qian

Shane-XB-Qian:
> But I want them only in specific cases:
> - Auto completion shall only happen after typing trigger characters and shall not include buffer completions.
> - Buffer completions shall only be included in the result when typing words and triggering omni-completion manually

fine, it burn my head, i donot want to sort up those compl one or two or some others to others depended one/some to one/some.
// but anyway, it is your choice, whatever...

-- 
shane.xb.qian

PR #558, opened 2024-09-17
Do LspAttached autocmd in the buffer its actually called for

64-bitman:
Issue: opening a session file with multiple buffers open would only set bufnr for the buffer that is focused from the start. This prevents buffer local mappings and autocmds from working properly.

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -434,7 +434,7 @@ def BufferInit(lspserverId: number, bnr: number): void
     endfor
 
     if exists('#User#LspAttached')
-      doautocmd <nomodeline> User LspAttached
+      execute ':' .. bnr .. 'bufdo doautocmd <nomodeline> User LspAttached'
     endif
   endif
 enddef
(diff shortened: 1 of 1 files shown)


autoload/lsp/lsp.vim
```diff
@@ -434,7 +434,7 @@ def BufferInit(lspserverId: number, bnr: number): void
     endfor
 
     if exists('#User#LspAttached')
-      doautocmd <nomodeline> User LspAttached
+      win_execute(bufwinnr(bnr), 'doautocmd <nomodeline> User LspAttached', silent)
```
Shane-XB-Qian, line 437: this seems incorrect, winnr maybe changed or not existed.

64-bitman: Any other ideas?

GreenCourt:
Thanks for the patch.

I tried it, but it does not trigger the LspAttached (even if the window exists).
And I had to fix the code where the variable `silent` wasn’t defined in the patch.

mmrwoods:
Is this fixed by #645?

64-bitman:
> Is this fixed by #645?

Seems so

yegappan:
Thanks for the confirmation.  I will close this PR.  Please try out the latest plugin which also adds the LspAttachedContext global variable containing the buffer number.

PR #562, opened 2024-09-28
let LS detect workspace root after first file opened

### LINKED ISSUE #512: lua lsp uses always current work dir when LSP initalized when editing file
According to the LSP logs, the root directory is always that of the current working directory when the LSP was added:
In `/path/to/lua-language-server/log/` a log file for this working directory shows up containing

```
  rootPath = "/home/konfekt",
  rootUri = "file:///home/konfekt/",
```

after Vim was started in `~/`, editing a file in a project dir below `~/`, throwing an error that more than 100000 files were scanned.

Instead, one would expect the rootPath to be set to the current working directory.

According to https://luals.github.io/wiki/faq/#why-is-the-server-scanning-the-wrong-folder

> When not using Visual Studio Code, check the rootUri field in your configuration - some clients allow users to customize this value. If the value is correct or cannot be modified, please report the issue to the developer of the client you are using.

I am using latest LSP master and Vim 9.1.369 on Ubuntu 22.04

Konfekt:
Only set workspace root once the first file of that file type is opened; otherwise the workspace root often is `$HOME` or more generally different from the repo of the file (say when using Gvim, or generally opening a file in a Git repo different from the current work dir) and the LS scans too many files

Partially resolves [0] as ideally this workspace root would automatically change with the CWD, in particular after loading a session (`:h mksession`)

[0]: https://github.com/yegappan/lsp/issues/512

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ $ vim -u NONE -c "helptags $HOME/.vim/pack/downloads/opt/lsp/doc" -c q
 ```
 
 After installing the plugin using the above steps, add the following line to
-your $HOME/.vimrc file:
+your `$HOME/.vimrc` file:
 
 ```viml
 packadd lsp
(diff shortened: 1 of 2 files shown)


Konfekt:
If initializing the LS for a language as soon as a buffer of that language is opened sounds sensible, then the cleaner solution would be officializing that list of LSs (g:LspServers) and let the plug-in take care of adding the LSP and all commands buffer-locally. If this also sounds sensible, I could craft such a PR.

Issue #366, opened 2023-07-26
omnifunc/LspOmniFunc does not filter completion options


dunric:
When `autoComplete` is set to false, manual invocation of a completion uses `omnifunc` buffer setting which defaults to `LspOmniFunc`. However it completely ignores already typed chars and offers full list of completion options from the server.

For example, in a rust filetype buffer using rust-analyzer as a lsp server, invoking omnifunc completion for a string literal followed with `.to`

```rust
"hello".to
```

offers unfiltered and unsorted list of completions:
```
replace
replacen
to_lowercase
to_uppercase
repeat
to_ascii_lowercase
…
```

instead of only words starting with `to`

```
to_lowercase
to_uppercase
to_ascii_lowercase
to_ascii_upperrcase
to_string
to_owned
```

When `autoComplete` is on it does not happen, because there is no existing prefix by which options are filtered.

Expected behavior: already typed chars are used for completion options.

Completion matcher is *"case"* for case sensitive search.

girishji:
This is because the results are not sorted correctly from rust-analyzer. You can try this plugin https://github.com/girishji/vimcomplete and see if it solves your problem.

PR #568, opened 2024-11-02
Remove check for 'isIncomplete' on completion

saccarosium:
Hi,
this remove the check for 'isComplete' when trying to do completion.

The main issue is that some lsps, such as rust analyzer (#366), will set this option and if the language server returns an unfiltered list, the user will receive completion completely unrelated to what they typed before.

### Before

[Screencast_20241103_002049.webm](https://github.com/user-attachments/assets/441be262-f63c-4e04-8073-3d137a6a7f4b)
 
### After

[Screencast_20241103_002234.webm](https://github.com/user-attachments/assets/adfa4f7e-021e-4e61-ba0a-536dcb5c793e)

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -514,8 +514,7 @@ def g:LspOmniFunc(findstart: number, base: string): any
     var res: list<dict<any>> = lspserver.completeItems
     var prefix = lspserver.omniCompleteKeyword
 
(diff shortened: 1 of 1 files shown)


saccarosium:
@yegappan does this PR have any issues / blockers?

Josegodi (approved):

PR #572, opened 2024-11-10
Do not check for digits in request.id

### LINKED ISSUE #571: Request ID validation
When using `zls`, an error is thrown [here](https://github.com/yegappan/lsp/blob/e34c17e7f37244ad361fb9af09fa755e21051cb3/autoload/lsp/lspserver.vim#L333). The reason is because `zls` is not sending a string that contains digits as can be seen [here](https://github.com/zigtools/zls/blob/6e0e7a89eb893650a44721c0199a88d8c17b8b61/src/Server.zig#L729) as an example.

The specification says that the request's ID can be a [number *or* a string](https://github.com/zigtools/zls/blob/6e0e7a89eb893650a44721c0199a88d8c17b8b61/src/Server.zig#L729), but, as fat as I can see it doesn't say that this ID must contain digits (but it would make sense). 

I came up with this "workaround":

```diff
diff --git a/autoload/lsp/lspserver.vim b/autoload/lsp/lspserver.vim
index 7304b00..1d4b0ed 100644
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -326,11 +326,9 @@ enddef

 # send a response message to the server
 def SendResponse(lspserver: dict<any>, request: dict<any>, result: any, error: dict<any>)
-  if (request.id->type() == v:t_string
-       && (request.id->trim() =~ '[^[:digit:]]\+'
-           || request.id->trim()->empty()))
+  if (request.id->type() == v:t_string && request.id->trim()->empty())
     || (request.id->type() != v:t_string && request.id->type() != v:t_number)
-    util.ErrMsg('request.id of response to LSP server is not a correct number')
+    util.ErrMsg('request.id of response to LSP server must be a number or a string')
     return
   endif
   var resp: dict<any> = lspserver.createResponse(
```

What are your thoughts?

julien:
Fixes #571

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -326,11 +326,9 @@ enddef
 
 # send a response message to the server
 def SendResponse(lspserver: dict<any>, request: dict<any>, result: any, error: dict<any>)
-  if (request.id->type() == v:t_string
-	&& (request.id->trim() =~ '[^[:digit:]]\+'
-	    || request.id->trim()->empty()))
+  if (request.id->type() == v:t_string && request.id->trim()->empty())
     || (request.id->type() != v:t_string && request.id->type() != v:t_number)
-    util.ErrMsg('request.id of response to LSP server is not a correct number')
+    util.ErrMsg('request.id of response to LSP server must be a number or a string')
     return
   endif
   var resp: dict<any> = lspserver.createResponse(


autoload/lsp/lspserver.vim
```diff
@@ -326,11 +326,9 @@ enddef
...
-  if (request.id->type() == v:t_string
-	&& (request.id->trim() =~ '[^[:digit:]]\+'
-	    || request.id->trim()->empty()))
+  if (request.id->type() == v:t_string && request.id->trim()->empty())
     || (request.id->type() != v:t_string && request.id->type() != v:t_number)
-    util.ErrMsg('request.id of response to LSP server is not a correct number')
+    util.ErrMsg('request.id of response to LSP server must be a number or a string')
     return
   endif
   var resp: dict<any> = lspserver.createResponse(
```
Shane-XB-Qian, line 334: your change here made `request.id->str2nr()` maybe always `0`
maybe you can try to refine it in the future~

julien: yes, i know. concerning this, maybe it's be possible to generate a number based on the string that is passed as the id ... i'll have to think about it but i agree that this isn't the best way to fix the issue.

julien: What do you think you something like this?

```diff
diff --git a/autoload/lsp/lspserver.vim b/autoload/lsp/lspserver.vim
index 1d4b0ed..40f0c0b 100644
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -332,7 +332,7 @@ def SendResponse(lspserver: dict<any>, request: dict<any>, result: any, error: d
     return
   endif
   var resp: dict<any> = lspserver.createResponse(
-           request.id->type() == v:t_string ? request.id->str2nr() : request.id)
+           request.id->type() == v:t_string ? GenerateNumber(request.id) : request.id)
   if error->empty()
     resp->extend({result: result})
   else
@@ -1856,6 +1856,18 @@ def GetUniqueServerId(): number
   return UniqueServerIdCounter
 enddef

+# Generates a number given a string
+def GenerateNumber(input: string): number
+  var length = len(input)
+
+  var hash = 0
+  for char in split(input, '\zs')
+    hash += char2nr(char)
+  endfor
+
+  return length * 1000 + hash
+enddef
+
 export def NewLspServer(serverParams: dict<any>): dict<any>
   var lspserver: dict<any> = {
     id: GetUniqueServerId(),

```

Shane-XB-Qian:
> What do you think you something like this?
> -           request.id->type() == v:t_string ? request.id->str2nr() : request.id)
> +           request.id->type() == v:t_string ? GenerateNumber(request.id) : request.id)

if i recall correctly, it should reply the id which server gave you.

-- 
shane.xb.qian

julien:
Gotcha, thanks!

Issue #581, opened 2024-12-18
Improve complete-popup on vertical screens


dezza:
I suggest this can be improved:
![Screenshot_20241218_215359](https://github.com/user-attachments/assets/9caaf044-0baf-4396-ad0f-a7b17412e877)
I looked around for settings and none seem like a permanent fix (_except switching to preview window obviously_). Yes you can set `pumheight=20` which will limit it vertically, but not horizontally in a sensible way afaik. It is dictated by the item length for each row `word, kind, info`, (functionname, args, markdown doc)

One could make a function for how occupied the screen is by the popup from the cursor, to act on it in completion.
```vim
def ColumnPct(in: number): number
  var onePct: float = ( &columns / 100.0 )
  var retPct: number = float2nr( in / onePct )

  return retPct
enddef
```

https://github.com/yegappan/lsp/blob/main/autoload/lsp/completion.vim#L315
If columns occupied by screen is 70%+
```vim
  if (ColumnPct(col('.') + longestMenu + longestWord) > 69)
    completeItems->map(((k, v) => (v->extend({'menu': ''}))))
  endif
```
As an example, just to make a screenshot to compare without the function-args signature.
![Screenshot_20241218_230457](https://github.com/user-attachments/assets/9b93f6f8-5a78-476a-bb54-6a68e70133d1)

The best ideas I could come up with quick:

1. **Append the function-args removed from screenshot above into the actual info popup `d.info` as a separate header in the top.** (https://github.com/yegappan/lsp/blob/main/autoload/lsp/completion.vim#L281-L288)
I suggest a [lsp-options](https://github.com/yegappan/lsp/blob/main/doc/lsp.txt#L445-L698) for toggling this on/off by default. Some other lsp plugins have this. I would probably opt for this myself. Straight-forward way to quickly save a lot of rows for vertical screens.

2. **Limit the func-name, func-args row in length, cut off by padding "`..`" to the right or both names** and functions-args in length. It should be determined by both cursor pos, window size (columns) and the longest line-length in the info popup, which would ensure it never limits unnecessarily and always maximizes width.

Any other ideas?

PR #582, opened 2024-12-29
Add option to condense completion menu items (would address #581)

berggeist:
This will add a new option `condensedCompletionMenu`, which allows to condense (or minimize) the text (width) of the completion menu items:
The default behavior for the completion menu items does not only display the word that will be completed but also additional details, like completion kind and detailed information, provided by the language LSP server. Typically, especially for typed languages for functions their call parameters with all their types are included. This makes the display of the completion menu items surprisingly overwhelming :smile: 
When this option new `condensedCompletionMenu` is enabled only the (key-) word and its kind is part of the completion menu item. The additional provided details from the LSP server are moved to the info popup, which is shown as soon as a completion item is selected.
This enhancement would address the issue:
 **Improve complete-popup on vertical screens #581**  https://github.com/yegappan/lsp/issues/581#issue-2748842834

Here is a screenshot of the default completion menu:
2024-12-29_condensedComplMenu-false

and here if this new option has been enabled:
2024-12-29_condensedComplMenu-true png

Instead of sprinkling the necessary modifications all over the source code, I decided to collect them in one place, which maybe would be less efficient but would improve the readability of this enhancement.

 Caveat: For completion LazyDoc the detailed information which is moved to the info popup will be automatically overwritten with the lazily delivered documentation from the LSP server.

#581

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -310,6 +310,26 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     endif  
 
     d.user_data = item
+
+    # Condense completion menu items to single words (plus kind)
+    # Move all additional details to the info popup
+    # Caveat: LazyDoc will override moved details!
+    if lspOpts.condensedCompletionMenu
+      const SEP = (s) => empty(s) ? "" : "\n- - -\n"
+      var infoText = ''
+      if d->has_key('abbr') && len(d.abbr) > len(d.word)
+        infoText ..= SEP(infoText) .. '    ' .. d.abbr
+        d.abbr = d.word
+      endif
+      if d->has_key('menu') && !empty(d.menu)
+        infoText ..= SEP(infoText) .. '    ' .. d.menu
+        d.menu = ''
+      endif
+      if !lspserver.completionLazyDoc && !empty(infoText)
+        d.info = infoText .. (d->has_key('info') ? SEP(infoText) .. d.info : '')
+      endif
+    endif
+
     completeItems->add(d)
   endfor
 
--- a/autoload/lsp/options.vim
+++ b/autoload/lsp/options.vim
@@ -143,6 +143,9 @@ export var lspOptions: dict<any> = {
 
   # Filter duplicate completion items
   filterCompletionDuplicates: false,
+
+  # Condenses the completion menu items to single (key-)words (plus kind)
+  condensedCompletionMenu: false,
 }
 
 # set the LSP plugin options from the user provided option values
--- a/doc/lsp.txt
+++ b/doc/lsp.txt
@@ -684,6 +684,12 @@ filterCompletionDuplicates |Boolean| option. If enabled, duplicate completion
 			items sent from the server will be filtered to only
 			include one instance of the duplicates. 
 
+						*lsp-opt-condensedCompletionMenu*
+condensedCompletionMenu |Boolean| option. If enabled, minimizes completion 
+			menu items to single (key-) words (plus kind).
+			Moves all additional details to info popup.
+			Caveat: LazyDoc will override moved details!
+
 For example, to disable the automatic placement of signs for the LSP
 diagnostic messages, you can add the following line to your .vimrc file: >
 


autoload/lsp/completion.vim
```diff
@@ -310,6 +310,26 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
...
+
+    # Condense completion menu items to single words (plus kind)
+    # Move all additional details to the info popup
+    # Caveat: LazyDoc will override moved details!
+    if lspOpts.condensedCompletionMenu
+      const SEP = (s) => empty(s) ? "" : "\n- - -\n"
+      var infoText = ''
+      if d->has_key('abbr') && len(d.abbr) > len(d.word)
+        infoText ..= SEP(infoText) .. '    ' .. d.abbr
+        d.abbr = d.word
```
Shane-XB-Qian, line 322: why change this?

autoload/lsp/completion.vim
```diff
@@ -310,6 +310,26 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
...
+      endif
+      if d->has_key('menu') && !empty(d.menu)
+        infoText ..= SEP(infoText) .. '    ' .. d.menu
+        d.menu = ''
```
Shane-XB-Qian, line 326: why change this?

Shane-XB-Qian:
i like your idea, but:
1. your example for golang seems good, but clangd its content in 'menu' seems was 'return type', so moving that to info popup seems odd.
2. to lazy doc servers, how reliable you made it work?

berggeist:
Hi & thank you for your review!

**Why did I make the changes like I did?**
First, I wanted to keep the changes as simple as possible in order to not damage existing source code.
Then, I wanted to make the completion menu item, as narrow as possible, in order to give much more display area to the info popup.

**How does Vim display the completion items:**
Each completion item, handed over to Vim, has this attributes:
- word (which contains typically only the short e.g. function name)
- abbr (which is typically more more verbose and contains e.g. the complete function signature)
- kind
- menu (which is handled very differently in each LSP server, e.g. `clangd` returns the return type of a function, whereas `gopls` returns the function signature, without the function name. Other LSP servers might leave this attribute even empty)

Vim concatenates the following attributes into the completion menu and displays the `info` attribute in the info-popup which is displayed when an item gets selected:
```
   +--- "word | abbr"
   |                 +--- "kind" 
   |                 |  +--- "menu"
   |                 |  |
   v                 v  v

  aaaaaaaaaaaaaaaaaa k mmmmmmmmmmmmmmmmmm
```
Unfortunately different LSP servers provide you with different data in their completion responses.

Therefore I decided to 
- take the smaller text of the `word` or the `abbrev`, which is typically more WYSIWYG (what you see is what you get). 
  In this case the `abbrev` will be moved to the info-popup.
- leave the `kind` untouched, as it is only one character short.
- move the `menu` to the info popup.

**Let's look into an example using the `clangd` LSP server:**

Here is the completion menu with the new option `condensedCompletionMenu` **disabled**:
2024-12-30_clangd_condensedComplMenu-false

and here with the new option `condensedCompletionMenu` **enabled**:
2024-12-30_clangd_condensedComplMenu-true-2

We can even compare the look of the completion menu to the default **hover** documentation popup:
2024-12-30_clangd_hover-doc

As you can see the completion menu with the new option `condensedCompletionMenu` **enabled** resembles much more the look of the hover documentation popup. Especially the type of the return value. Unfortunately the `clangd` LSP server does not (yet) provide the structured parameter info within its completion responses (nor does it provide the doxygen doc-strings).

autoload/lsp/completion.vim
berggeist: In order to make the completion menu as narrow as possible the shorter text of __abbr__ or __word__ is used. When __abbr__ is exchanged by __word__ (because __abbr__ is to verbose) then __abbr__ will be moved into the info-popup.

autoload/lsp/completion.vim
berggeist: __menu__ typically contains very verbose information provided by the LSP server. Therefore it is simply removed from the menu item (see also my explanation how Vim concatenates the display value for a menu item below). 
In this case this verbose information is __saved__ by moving it into the info-popup.

autoload/lsp/completion.vim
Shane-XB-Qian: i means d.word _maybe_ empty; 
and changing those value not sure had outage.

Shane-XB-Qian:
i knew your idea, but my concern:
1. d.word maybe empty, so you compared abbr's len with word's len, maybe wrong?
2. something cfg e.g 'filterCompletionDuplicates' if on, i am worried you changing those value made the result not reflect the truth.
3. still about lazy doc servers, how you deal with them?

-- 
shane.xb.qian

berggeist:
> i knew your idea, but my concern: 

Let's address your concerns:

> 1. d.word maybe empty, so you compared abbr's len with word's len, maybe wrong? 

`d.word` **must not** be empty at the location, where my pull request applies, otherwise the existing code before my patch is doing something wrong.

> 2. something cfg e.g 'filterCompletionDuplicates' if on, i am worried you changing those value made the result not reflect the truth.

This was also a concern I worried myself. At the end, all I'm doing, is replacing the attribute `d.abbr` by `d.word` and `d.word` is exactly what gets inserted into your code after you accept the completion item. So far with my pull request, you get far more WYSIWG (what you see is what you get) as before. E.g. before applying my pull request a LSP server would provide _"zzTop"_ in the attribute `d.word` and _"Bee Gees"_ in the attribute `d.abbr` with the consequence that _"Bee Gees"_ is displayed in the completion menu whereas _"zzTop"_ get inserted into your code. After applying my pull request _"zzTop"_ is displayed in the completion menu **and** will be inserted into your code.

>  3. still about lazy doc servers, how you deal with them?

I kept the handling of lazy doc servers __as is__. That means, if I recognize in line 328 
```
 if !lspserver.completionLazyDoc && !empty(infoText)
```
that the LSP server is providing its doc lazily, I do not touch the info-attribute at all. If it worked well up to now than it will work as well after applying my pull request. With the consequence that if e.g. the attribute `d.menu` is being emptied before it will not be shown at all, as it is not moved to the info-pop where it would be overwritten by the lazily supplied doc anyway. Most LSP servers I tried out do show the data provided in `d.menu` also in the (lazily provided) info doc.
But if you are worrying about you could just disable this option anyway. Which is the default, i.e. it is per default **disabled** in order to preserve the existing behavior.

Shane-XB-Qian:
> > 1. d.word maybe empty, so you compared abbr's len with word's len, maybe wrong? 
> 
> `d.word` **must not** be empty at the location, where my pull request applies, otherwise the existing code before my patch is doing something wrong.

well, who knows, at least, if d.word is empty, then d.abbr should not get from it.

> > 2. something cfg e.g 'filterCompletionDuplicates' if on, i am worried you changing those value made the result not reflect the truth.
> 
> This was also a concern I worried myself. At the end, all I'm doing, is replacing the attribute `d.abbr` by `d.word` and `d.word` is exactly what gets inserted into your code after you accept the completion item. So far with my pull request, you get far more WYSIWG (what you see is what you get) as before. E.g. before applying my pull request a LSP server would provide _"zzTop"_ in the attribute `d.word` and _"Bee Gees"_ in the attribute `d.abbr` with the consequence that _"Bee Gees"_ is displayed in the completion menu whereas _"zzTop"_ get inserted into your code. After applying my pull request _"zzTop"_ is displayed in the completion menu **and** will be inserted into your code.

but you changed d.menu to '' empty too, not only switched d.word and d.abbr

> >  3. still about lazy doc servers, how you deal with them?
> 
> I kept the handling of lazy doc servers __as is__. That means, if I recognize in line 328 
> ```
>  if !lspserver.completionLazyDoc && !empty(infoText)
> ```
> that the LSP server is providing its doc lazily, I do not touch the info-attribute at all. If it worked well up to now than it will work as well after applying my pull request. With the consequence that if e.g. the attribute `d.menu` is being emptied before it will not be shown at all, as it is not moved to the info-pop where it would be overwritten by the lazily supplied doc anyway. Most LSP servers I tried out do show the data provided in `d.menu` also in the (lazily provided) info doc.
> But if you are worrying about you could just disable this option anyway. Which is the default, i.e. it is per default **disabled** in order to preserve the existing behavior.

maybe let `!lspserver.completionLazyDoc` helped the d.word and d.abbr too, not only d.menu?

-- 
shane.xb.qian

autoload/lsp/completion.vim
berggeist: No ,`d.word` cannot be empty.
This is ensured by the specification of the language server protocol that mandates that at least `CompletionItem.label` must be not empty. And the _fallback_  (which might be in my eyes not straightforward regarding the handling of fuzzy-matching, see also #584 ) of this code (beginning with line 197) ensures this:
```
    if item->has_key('textEdit') &&
	lspOpts.completionMatcherValue != opt.COMPLETIONMATCHER_FUZZY
...
    elseif item->has_key('insertText')
...
    else
      d.word = item.label
    endif
```

`

berggeist:
> well, who knows, at least, if d.word is empty, then d.abbr should not get from it.

See also me remark from above:
No ,`d.word` cannot be empty. This is ensured by the specification of the language server protocol that mandates that at least `CompletionItem.label` must be not empty. And the _fallback_  (which might be in my eyes not straightforward regarding the handling of fuzzy-matching, see also #584 ) of this code (beginning with line 197) ensures this:
```
    if item->has_key('textEdit') &&
	lspOpts.completionMatcherValue != opt.COMPLETIONMATCHER_FUZZY
...
    elseif item->has_key('insertText')
...
    else
      d.word = item.label
    endif
```

> but you changed d.menu to '' empty too, not only switched d.word and d.abbr

Intentionally my modifications are inserted *after* the deduplication. Why: Because for e.g. the java language server `jdtls` provides in its completion list also all overloaded methods and I did not want to remove them. As for now you might see for overloaded methods multiple entries with the same `d.word` but as soon as you select any of them the info-popup will reveal their full method signature, so no information gets lost. 

> maybe let `!lspserver.completionLazyDoc` helped the d.word and d.abbr too, not only d.menu?

The worst thing that could happen is that the _lazily_ provided documentation by the LSP server will override the moved information, i.e. `d.abbr` and `d.menu`, but this cannot be fixed anyhow. If someone don't want this behavior, he just has to leave this option to its default value, i.e. *disabled*.

autoload/lsp/completion.vim
Shane-XB-Qian, line 326: here `d.menu = ''` maybe wrong
since checking following lazy-doc at next, that maybe false, but done d.menu = '' here already.

Shane-XB-Qian (commented):
though yegappan accepted your PR, but maybe you need to refine it in the future..

dezza:
Thanks for taking a look at this! I like the result! 👍

dezza:
It would be cool if this would be adaptive, when there is more `&lines` than `&columns` ([autocmd.txt#VimResized](https://vimhelp.org/autocmd.txt.html#VimResized)) in case you prefer the default, except when vertical.

PR #583, opened 2024-12-30
Fix completion item filtering condition

soywod:
I try to setup your LSP plugin for Rust using `rust-analyzer`, and I noticed that more results were shown in the completion menu. After diving into your code, I noticed that the following condition was not aligned with the comment just below. By removing the negation, it somehow filters properly the items. I am not sure about the consequences of such a change, but after the change it works as it was initially supposed to be.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -227,7 +227,7 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
       # snippet completion.  Needs a snippet plugin to expand the snippet.
       # Remove all the snippet placeholders
       d.word = MakeValidWord(d.word)
-    elseif !lspserver.completeItemsIsIncomplete || lspOpts.useBufferCompletion
+    elseif lspserver.completeItemsIsIncomplete || lspOpts.useBufferCompletion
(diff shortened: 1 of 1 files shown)


berggeist:
The provided information by the Rust language server is not straightforward, see also here ["Completion response says "isIncomplete": true even when the total number of completion items is two #7649"](https://github.com/rust-lang/rust-analyzer/issues/7649) 

According to the language server protocol specification `CompletionList.isIncomplete`
```
         /**
	 * This list is not complete. Further typing should result in recomputing
	 * this list.
	 *
	 * Recomputed lists have all their items replaced (not appended) in the
	 * incomplete completion sessions.
	 */
	isIncomplete: boolean;
```
shall provide a hint to the client that the provided completion list is only temporarily and will be soon replaced by another list so it is not necessary to apply any filtering to this completion list, because it will be lost anyway.
Unfortunately the Rust language server `rust-analyzer` does **always** set this flag `CompletionList.isIncomplete` to true. See also the link to the according github issue above.

soywod:
Interesting, thanks for the precision. What do you advise then?

PR #585, opened 2025-01-16
Support optional diagnostic severity

Tachi107:
The LSP specification defines the diagnostic severity as being optional, defaulting to Error. Properly handle such cases.

This patch also complements commit
5b94bf7635de39979e81f9dace85b94d248a67e8, which fixed the issue but only for when using the ALE integration. At the same time, it fixes it by using severity Error as the default, rather than Information.


Shane-XB-Qian:
@yegappan this sounds ok, and spec looks was default at 'error' level.

Issue #591, opened 2025-01-23
`aleSupport` sends incorrect end column


briancarbone:
It appears SendAleDiags is sending an extra column (you can see the extra highlight in the attached image) and it also appears the severity is incorrect (error indicated in the sign column where it should just be a hint).

![Image](https://github.com/user-attachments/assets/49ea4943-5dc6-40ac-ba18-cc95c26fcd9d)

briancarbone:
This is what it should look like... I'm not sure how that hint is becoming an error. I figure the column issue is an off-by-one error (i.e. if the end col is exclusive).

![Image](https://github.com/user-attachments/assets/39b4fc52-1075-4654-8ded-131e4c492382)

briancarbone:
The off-by-one error was easy to fix, but it appears ALE has no concept of "hint" as far as the `type` argument is concerned... fascinating.

briancarbone:
This needs further consideration; a PR to ALE seems appropriate.

PR #592, opened 2025-01-24
Fix ALE inline diagnostic end column and conform to their severity system

briancarbone:
LSP diagnostics funneled through ALE's diagnostic API:

1. Have an incorrect end column (as seen here: https://github.com/yegappan/lsp/issues/591);
2. Send unsupported 'Hint' diagnostics resulting in ALE defaulting to 'Error' (also seen at the link above).

This fixes both issues.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -348,15 +348,18 @@ def SendAleDiags(bnr: number, timerid: number)
     return
   endif
 
+  var typeMap: list<string> = ['E', 'W', 'I']
+
   # Convert to Ale's diagnostics format (:h ale-loclist-format)
   ale#other_source#ShowResults(bnr, 'lsp',
     diagsMap[bnr].sortedDiagnostics->mapnew((_, v) => {
+     var severity = get(v, "severity", 1)
      return {text: v.message,
              lnum: v.range.start.line + 1,
              col: util.GetLineByteFromPos(bnr, v.range.start) + 1,
              end_lnum: v.range.end.line + 1,
-             end_col: util.GetLineByteFromPos(bnr, v.range.end) + 1,
-             type: "EWIH"[get(v, "severity", 1) - 1]}
+             end_col: util.GetLineByteFromPos(bnr, v.range.end),
+             type: typeMap[(severity > 3 ? 3 : severity) - 1]}
     })
   )
 enddef


autoload/lsp/diag.vim
```diff
@@ -348,15 +348,23 @@ def SendAleDiags(bnr: number, timerid: number)
     return
   endif
 
+  def DiagSevToALESev(severity: number): string
+    var typeMap: list<string> = ['E', 'W', 'I']
+    if severity > 3
+      return 'I'
+    endif
+    return typeMap[severity - 1]
```
gcanat, line 356:
```suggestion
    return severity > 3 ? 'I' : typeMap[severity - 1]
```
This should be equivalent

briancarbone: I was keeping consistency with similar functions, e.g.:

```
def DiagSevToSignName(severity: number): string
  var typeMap: list<string> = ['LspDiagError', 'LspDiagWarning',
						'LspDiagInfo', 'LspDiagHint']
  if severity > 4
    return 'LspDiagHint'
  endif
  return typeMap[severity - 1]
enddef

def DiagSevToInlineHLName(severity: number): string
  var typeMap: list<string> = [
    'LspDiagInlineError',
    'LspDiagInlineWarning',
    'LspDiagInlineInfo',
    'LspDiagInlineHint'
  ]
  if severity > 4
    return 'LspDiagInlineHint'
  endif
  return typeMap[severity - 1]
enddef

def DiagSevToVirtualTextHLName(severity: number): string
  var typeMap: list<string> = [
    'LspDiagVirtualTextError',
    'LspDiagVirtualTextWarning',
    'LspDiagVirtualTextInfo',
    'LspDiagVirtualTextHint'
  ]
  if severity > 4
    return 'LspDiagVirtualTextHint'
  endif
  return typeMap[severity - 1]
enddef
```

autoload/lsp/diag.vim
```diff
@@ -348,15 +348,23 @@ def SendAleDiags(bnr: number, timerid: number)
...
   ale#other_source#ShowResults(bnr, 'lsp',
     diagsMap[bnr].sortedDiagnostics->mapnew((_, v) => {
      return {text: v.message,
              lnum: v.range.start.line + 1,
              col: util.GetLineByteFromPos(bnr, v.range.start) + 1,
              end_lnum: v.range.end.line + 1,
-             end_col: util.GetLineByteFromPos(bnr, v.range.end) + 1,
-             type: "EWIH"[get(v, "severity", 1) - 1]}
+             end_col: util.GetLineByteFromPos(bnr, v.range.end),
+             type: DiagSevToALESev(get(v, "severity", 4))}
```
Shane-XB-Qian, line 367: please check https://github.com/yegappan/lsp/pull/585

autoload/lsp/diag.vim
```diff
@@ -348,15 +348,23 @@ def SendAleDiags(bnr: number, timerid: number)
...
+    endif
+    return typeMap[severity - 1]
+  enddef
+
   # Convert to Ale's diagnostics format (:h ale-loclist-format)
   ale#other_source#ShowResults(bnr, 'lsp',
     diagsMap[bnr].sortedDiagnostics->mapnew((_, v) => {
      return {text: v.message,
              lnum: v.range.start.line + 1,
              col: util.GetLineByteFromPos(bnr, v.range.start) + 1,
```
Shane-XB-Qian, line 364: so here was this a 1-offset problem too?

autoload/lsp/diag.vim
```diff
@@ -348,15 +348,23 @@ def SendAleDiags(bnr: number, timerid: number)
     return
   endif
 
+  def DiagSevToALESev(severity: number): string
```
Shane-XB-Qian, line 351: looks no have to need a func

Shane-XB-Qian (commented):
seems was a ALE specific issue, was it confirmed:
1. 0-indexed or 1-indexed?
2. really no 'H'?
otherwise perhaps was a ale problem.

autoload/lsp/diag.vim
briancarbone: No, I do not see an issue with the start column. The rendering appears correct with just the end column adjusted.

briancarbone:
Updated in response to @Shane-XB-Qian's suggestions. Now defaults to severity 1 when not present and removed an unnecessary function for translating the severity.

vimpostor:
I think the change for `end_col` you did was still correct, it seems ale internally uses `matchaddpos()` to add the highlights, which expects as last argument the length of the highlight.

Ale computes the length incorrectly here by adding 1: https://github.com/dense-analysis/ale/blob/e319d489410df7705d72708ff9e44ae730721ac8/autoload/ale/highlight.vim#L51

I am pretty sure that this was originally an off-by-one error in Ale but it's hard to fix it in Ale now, because every usage site of the Ale API already works around that by subtracting 1 again, e.g.: https://github.com/dense-analysis/ale/blob/e319d489410df7705d72708ff9e44ae730721ac8/lua/ale/diagnostics.lua#L40

So indeed, we should subtract 1 as well. Do you want to reopen the PR with just that change?
I don't think the rest of the changes in this PR were correct, but that one could go in.

PR #597, opened 2025-02-07
Add support for range changes

pusewicz:
This adds support for incremental changes that are sent to the LSP server.

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -604,51 +604,109 @@ def TextdocDidClose(lspserver: dict<any>, bnr: number): void
   lspserver.sendNotification('textDocument/didClose', params)
 enddef
 
+# Compute character position in a line taking into account composing characters
+def GetCharPosition(line: string, bytePos: number): number
+  if bytePos <= 0
+    return 0
+  endif
+
+  # Use strcharpart to get substring up to bytePos
+  var substr = strpart(line, 0, bytePos)
+  # Count characters with skipcc=1 to handle composing chars separately
+  return strchars(substr, 1)
+enddef
+
+# Process line changes for incremental sync
+def ProcessIncrementalChanges(bnr: number, start: number,
+                             end: number, added: number): dict<any>
+  # Convert 1-based line numbers to 0-based for LSP
+  var startLnum = start - 1
+  var endLnum = end - 1
+
+  # Get start and end lines safely
+  var startLine = ''
+  var startLines = getbufline(bnr, start)
+  if !startLines->empty()
+    startLine = startLines[0]
+  endif
+
+  var endLine = ''
+  if end > 0
+    var endLines = getbufline(bnr, end)
+    if !endLines->empty()
+      endLine = endLines[0]
+    endif
+  endif
+
+  # Get full line content for changed range
+  var newText = getbufline(bnr, start, end + added - 1)->join("\n")
+  if len(newText) > 0
+    newText ..= "\n"  # Add trailing newline
+  endif
+
+  # Build LSP range
+  var changes = {
+    range: {
+      start: {
+        line: startLnum,
+        character: 0
+      },
+      end: {
+        line: endLnum,
+        character: len(endLine) > 0 ? GetCharPosition(endLine, len(endLine)) : 0
+      }
+    },
+    text: newText
+  }
+
+  return changes
+enddef
+
 # Send a file/document change notification to the language server.
 # Params: DidChangeTextDocumentParams
 def TextdocDidChange(lspserver: dict<any>, bnr: number, start: number,
-			end: number, added: number,
-			changes: list<dict<number>>): void
-  # Notification: 'textDocument/didChange'
-  # Params: DidChangeTextDocumentParams
-
-  var changeset: list<dict<any>> = []
-
-  # Compute the ranges and prepare the changeset
-  for change in changes
-    var lines: string
-    var start_lnum: number = change.lnum - 1
-    var end_lnum: number = change.end - 1
-    var start_col: number = 0
-    var end_col: number = 0
-
-    if change.added == 0
-      # lines changed
-      lines = getbufline(bnr, change.lnum, change.end - 1)->join("\n") .. "\n"
-    elseif change.added > 0
-      # lines added
-      lines = getbufline(bnr, change.lnum, change.lnum + change.added - 1)->join("\n") .. "\n"
-    else
-      # lines removed
-      end_lnum = change.lnum + (-change.added) - 1
-      lines = ''
+			    end: number, added: number, changes: list<dict<number>>): void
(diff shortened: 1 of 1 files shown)


pusewicz:
@yegappan Please let me know what you think!

yegappan:
Thanks for the contribution.

Did you run any tests to make sure that the correct contents are passed to the language server?  One of the problems that I ran into earlier when trying to implement this functionality was that under some conditions incorrect contents were passed to the language server.  After this, the language server will go out of sync with the buffer content in Vim.  Did you try making modifications to a somewhat large buffer and exercise some of the LSP functionality on the modified content?

pusewicz:
@yegappan I keep using this branch locally on a daily basis, on large code bases, and it _seems_ to be working fine. I do have extra logging added that I sometimes turn on and I spit out the lines that change to try to verify that this is correct. I looked at NeoVim's implementation and it seems similar, but yeah, I'm not sure what's the best way to verify the correctness. How about we move forward with this and fix as we go? Also, I'm not sure what's the issue with the tests and why are they so slow, even without the change.

DanielViberg:
I looked in to this a while ago, see https://github.com/yegappan/lsp/discussions/550

mmrwoods:
When support for incremental did change notifications was added to Neovim's builtin lsp client it was initially an opt-in experimental feature. Maybe that approach would be suitable here?

yegappan:
Yes. I agree.  I will see how this PR can be merged.

yegappan:
@mmrwoods I have now added the support for incrementally syncing the buffer changes to the language server (53b3cfa9fba3f2a08d4921424ec6b4623b047997).  Currently this is disabled by default.  You can enable it by setting the `incrementalSync` option to `true`.  Please try this out and let me know if you run into any issues.

mmrwoods:
@yegappan This breaks diagnostics and completion when using typescript language server, even when `incrementalSync` is `false`. I'm seeing messages like this in the debug log: 

```04/15/26 06:27:04: Got notification {"method":"window/logMessage","jsonrpc":"2.0","params":{"message":"Unexpected resource file:///Users/mwoods/code/foo.js","type":1}}```

At a glance Jedi seems fine (maybe doesn't support incremental did change?), sorry, no time to investigate further right now, but will check in some more detail later.

EDIT: it is specifically this commit that causes the problem, if I revert the commit all is well

**Comment by yegappan** [comment]
@mmrwoods I tried to reproduce this issue with the typescript language server.  I couldn't reproduce it.  Which version of the typescript language server are you using?  How are you registering the server with the LSP client?  Can you send a sample file that shows the issue?

**Comment by mmrwoods** [comment]
My config is here: https://github.com/mmrwoods/dotfiles/blob/main/vim/lsp-settings.vim

`typescript-language-server` is version 5.1.3

Even with a file this simple I get no completions or diagnostics:

```js
const fs = require('fs');

fs.readFile()
```

I should get these diagnostic messages, and do if I revert the commit to support incremental sync:

```
foo.js|1 col 12-25 note| File is a CommonJS module; it may be converted to an ES module.
foo.js|3 col 4-12 error| Expected 2-3 arguments, but got 0.
```

I don't think it's anything to do with ts config, if I save `foo.js` in my home dir I get the same issue.

Sorry, like yesterday, I'm tight on time again today, but I'll try to find time to investigate a little further.

mmrwoods:
@yegappan The problem is that typescript language server reports the `textDocumentSync` capability using an integer value representing the `TextDocumentSyncKind` number, e.g. `"textDocumentSync":2`.

While your code allows for this, it also only sets `lspserver.supportsDidOpenClose` to `true` when `textDocumentSync` is a dict, and thereafter no `textDocument/didOpen` message is sent to the server so LSP is effectively never enabled for the buffer.

I don't really understand the conditional `supportsDidOpenClose`, I thought this support was mandatory.

Here is a diff of a one-line code change that fixes the problem for me (not that I'd expect this to be committed):

```diff
diff --git a/autoload/lsp/capabilities.vim b/autoload/lsp/capabilities.vim
index 9501bca..57e936f 100644
--- a/autoload/lsp/capabilities.vim
+++ b/autoload/lsp/capabilities.vim
@@ -36,7 +36,7 @@ export def ProcessServerCaps(lspserver: dict<any>, caps: dict<any>)
 
   # textDocumentSync capabilities
   lspserver.supportsDidSave = false
-  lspserver.supportsDidOpenClose = false
+  lspserver.supportsDidOpenClose = true
   # Default to TextDocumentSyncKind.None
   lspserver.textDocumentSync = 0
   if lspserver.caps->has_key('textDocumentSync')
```

This is the code from typescript language server that generates the "Unexpected resource" error I reported earlier:

```ts
    public toOpenDocument(textDocumentUri: DocumentUri, options: { suppressAlertOnFailure?: boolean; } = {}): LspDocument | undefined {
        const filepath = this.toTsFilePath(textDocumentUri);
        const document = filepath && this.documents.get(filepath);
        if (!document) {
            const uri = URI.parse(textDocumentUri);
            if (!options.suppressAlertOnFailure && !fileSchemes.disabledSchemes.has(uri.scheme)) {
                console.error(`Unexpected resource ${textDocumentUri}`);
            }
            return undefined;
        }
        return document;
    }
```

I haven't gone through the code in enough detail to be sure, but `this.documents` here seems to be the collection of opened files, presumably updated when handling `textDocument/didOpen` notifications from the client, which are no longer being sent.

yegappan:
@mmrwoods Thanks for investigating this issue and finding the root cause. The commit [cd48a9d](https://github.com/yegappan/lsp/commit/cd48a9d1c67701c37cd2bff0201a060e10a29af3) ensures that the supportsDidOpenClose capability is set correctly when the textDocumentSync server capability is a non-zero value. This change was pushed after the initial incremental sync support was added. Could you verify if you are using the latest version of the plugin that includes this commit?

mmrwoods:
@yegappan Yep, that fixes it, thanks 👍

yegappan:
>
> @yegappan Yep, that fixes it, thanks 👍
>

@mmrwoods Thanks for the confirmation.  Please try out the new incremental sync support and let me know if you see any issues.

Shane-XB-Qian:
> ```
> foo.js|1 col 12-25 note| File is a CommonJS module; it may be converted to an ES module.
> foo.js|3 col 4-12 error| Expected 2-3 arguments, but got 0.
> ```

After so, your original example, you now got 2 err or just the first one?

-- 
shane.xb.qian

mmrwoods:
> After so, your original example, you now got 2 err or just the first one?

Dependent on config, I get two errors as before when I save it in the same directory with a `jsconfig.json`, but only one (the first one) if I save it in my home dir.

arp242:
Looks like this was added in 53b3cfa, which supersedes this PR?

yegappan:
The incremental sync support is now added to the plugin.

Issue #601, opened 2025-02-18
Add yaml-language-server to the wiki


isaiahfisher:
Could we add an entry to the wiki for the yaml-language-server? Here is my proposed markdown for such an entry:

````
## YAML Language Server
**Language**: YAML

**Home Page**: https://github.com/redhat-developer/yaml-language-server

Sample code to add the YAML Language Server to the LSP plugin:
```
call LspAddServer([#{
            \   name: 'yaml-language-server',
            \   filetype: 'yaml',
            \   path: '/usr/local/node_modules/.bin/yaml-language-server',
            \   args: ['--stdio'],
            \   workspaceConfig: #{
            \       yaml: #{
            \           schemas: {
            \               "https://raw.githubusercontent.com/compose-spec/compose-spec/master/schema/compose-spec.json" : [
            \                   "**/*docker-compose*.yaml"
            \               ],
            \              "https://json.schemastore.org/chart.json": [
            \                   "**helm/values*.yaml"
            \               ]
            \           }
            \       }
            \   }
            \ }])
```
**Note**: The workspaceConfig `yaml.schemas` object is a key-value pair where 
the key is a link or path to a yaml schema and the value is an array of 
file path patterns to enforce the schema on. There can be any number of 
enforced schemas based on a user's need, but I have given examples
based on enforcing the compose spec and adding helm chart suggestions for configuration.
````

Apologies in advance if there is some way to contribute a pull request directly to the wiki page, I did not see the option anywhere. Thank you!

Edited based on suggestions from @Shane-XB-Qian

Shane-XB-Qian:
>             \   workspaceConfig: #{
>             \       yaml: #{
>             \           schemas: {
>             \               "https://json.schemastore.org/composer" : [
>             \                   "**directory/file*.yaml"
>             \           }
>             \       }
>             \   }

May i ask what is this config meaning here, would you give some explanation or this was just some example?

BTW: do you know (i'd like to ask since i have no idea) how to judge yaml-language-server vs ansible-language-server?

> Apologies in advance if there is some way to contribute a pull request directly to the wiki page, I did not see the option anywhere. Thank you!

No, you cannot, user cannot directly PR repo wiki page as i know.

-- 
shane.xb.qian

isaiahfisher:
> \   workspaceConfig: #{
>              \       yaml: #{
>              \           schemas: {
>              \               "https://json.schemastore.org/composer" : [
>              \                   "**directory/file*.yaml"
>              \           }
[...]

PR #603, opened 2025-02-20
move Wiki to doc/ for possible PRs on setting up LSs

Konfekt:
Since the Wiki is locked, add a document documenting LS setups similar to https://github.com/neovim/nvim-lspconfig/blob/master/doc/configs.md to allow curated contributions (as in https://github.com/yegappan/lsp/issues/601);
also point out loc.cit. as fallback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,8 @@ call LspAddServer([#{
 	\  }])
 ```
 
-The above lines register the language servers for C/C++, Javascript/Typescript, Go and Rust file types.  Refer to the [Wiki](https://github.com/yegappan/lsp/wiki) page for various language server specific configuration.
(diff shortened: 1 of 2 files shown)


Konfekt:
This won't change, but now there's a standard way to contribute whereas for the Wiki these requests are scattered in issues and require manual pasting by the repo owner

Shane-XB-Qian:
> This won't change, but now there's a standard way to contribute whereas for the Wiki these requests are scattered in issues and require manual pasting by the repo owner 

I remember there is link in doc said about repo wiki already, so maybe that's good enough.
And you can git clone the wiki like general repo also, actually, maybe not worthy to separate such things too much.

-- 
shane.xb.qian

PR #609, opened 2025-03-07
Adds verible-ls config setting to the doc

tarikgraba:


tarikgraba:
`verible-ls` is a language server for verilog/systemverilog HDL.

https://github.com/chipsalliance/verible

PR #610, opened 2025-03-08
open buffers from special windows in preview window

Konfekt:
For example in the window shown after :LspIncomingCalls the openend references quickly become unreadable due to smaller and smaller splits; instead use the preview window to peek at them.

Prospectively use different mappings, say p for preview, o to open, to edit the buffer or open in a new split similar to @yssl 's' [0]

[0]: https://github.com/yssl/QFEnter?tab=readme-ov-file#usage


mmrwoods:
Might a better solution to this be to move to the next "usable" window?

e.g. https://github.com/vim-fuzzbox/fuzzbox.vim/blob/main/autoload/fuzzbox/internal/utils.vim#L66-L78

Konfekt:
Yes, for example stay in the same window, say by checking it to be a fleeting window of the same file type

PR #626, opened 2025-05-05
Adds the option ignoreCompleteItemsIsIncomplete.

berggeist:
Optional list of LSP server names for which
'IsIncomplete' messages should be ignored and
filtering of received completion item lists
is enforced. This is helpful for LSP servers
that are misbehaving and always send this
kind of 'IsIncomplete' messages.

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -178,7 +178,11 @@ export def CompletionReply(lspserver: dict<any>, cItems: any)
     items = cItems
   else
     items = cItems.items
-    lspserver.completeItemsIsIncomplete = cItems->get('isIncomplete', false)
+    if opt.lspOptions.ignoreCompleteItemsIsIncomplete->index(lspserver.name) >= 0
+      lspserver.completeItemsIsIncomplete = v:false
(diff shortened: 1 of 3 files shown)


berggeist:
This is an attempt to fix #583  
 [Fix completion item filtering condition](https://github.com/yegappan/lsp/pull/583)

berggeist:
As an example for a LSP server for which this new option might be helpful:

See this issue of the  Rust language server  https://github.com/rust-lang/rust-analyzer/issues/7649

According to the language server protocol specification CompletionList.isIncomplete

         /**
	 * This list is not complete. Further typing should result in recomputing
	 * this list.
	 *
	 * Recomputed lists have all their items replaced (not appended) in the
	 * incomplete completion sessions.
	 */
	isIncomplete: boolean;

shall provide a hint to the client that the provided completion list is only temporarily and will be soon replaced by another list so it is not necessary to apply any filtering to this completion list, because it will be lost anyway.
Unfortunately the Rust language server rust-analyzer does always set this flag CompletionList.isIncomplete to true. See also the link to the according github issue above.

PR #627, opened 2025-05-06
Feat/refactor popup configuration

nickspoons:
After discussing the new popup configuration with @jclsn and others in #vim on libera, we realised that the implementation was not very consistent with the normal lsp configuration, via `SetLspOptions(...)`.

This PR makes the configuration more consistent, and simpler, as there is now only a single option required to enable or disable popup borders (`{popupBorder: false}`, defaults to `true`), set popup highlights (`{popupHighlight: 'Pmenu'}`, defaults to `'LspPopup'` which links to `'Normal'`) and set popup border highlights (`{popupBorderHighlight: 'Title'}`, defaults to `'LspPopupBorder'` which links to `'Normal'`).

Individual popup types may then be overridden, e.g. disable peek borders with `{popupBorderPeek: false}`.

@jclsn feedback welcome


nickspoons:
I'm not sure how to read or fix these failing tests

jclsn:
I am very happy with it, but a few things:

- In clangd the documentation popup right to the Pmenu is not styled (is this a bubble menu?)
- I believe borders should be off by default for consistency
- When the borders are off: 
    - a padding should be inserted again as a border between the popup and buffer text
    - the document symbol and symbol references (peek references) still shows the default double border
    - ``popupHighlight`` and ``popupBorderHighlight`` should both fall back to ``Pmenu` for consistency
    
There is no way to add border to the ``Pmenu``, is there?

berggeist:
I would assume it is worth to postpone any rework here and to first wait for the completion of this open PR in Vim: [feat: add pumborder #17091](https://github.com/vim/vim/pull/17091) and afterwards create an overall consistent solution which fits to the out-of-box configuration and look&feel of  Vim.

nickspoons:
@berggeist that's a good point, although I'm not sure we need to wait. Whatever outcome that PR has, we will still be able to use this configuration for highlight selection etc., even if we end up dropping the borderchars parts.

jclsn:
@berggeist Thanks for pointing that out. I think we should still merge this, as the current merged state is already inconsistent. As soon as the Vim PR is merged, we can default to borders everywhere.

Shane-XB-Qian:
> @berggeist that's a good point, although I'm not sure we need to wait. Whatever outcome that PR has, we will still be able to use this configuration for highlight selection etc., even if we end up dropping the borderchars parts.

perhaps better impl was making those options be uniformed to single one 'switch' (those 4 cfg), then `popup_id->set_option()` to apply to all popup;
even latter vim had native impl of border, then you can just change that 'switch';
but now those seems a bit spread out, and too much options which perhaps no require to 4 options for 6 places?
maybe a bit burden to maintain, it was designed to be simple UI (and cfg of course), but now just 4*6 options~ :smile:

-- 
shane.xb.qian

nickspoons:
@Shane-XB-Qian  Are you saying there are too many options for the user to configure, or too many locations in code that they are configured? I had been thinking of refactoring all the code locations to something like this:

```vim
# autoload/lsp/options.vim
export def PopupConfigure(type: string, opts: dict<any>)
  opts.highlight = get(lspOptions, 'popupHighlight' .. type, lspOptions.popupHighlight)
  if get(lspOptions, 'popupBorder' .. type, lspOptions.popupBorder)
    opts.border = []
    opts.borderchars = lspOptions.popupBorderChars
    opts.borderhighlight = [get(lspOptions, 'popupBorderHighlight' .. type, lspOptions.popupBorderHighlight)]
  endif
  return opts
enddef

# autocmd/lsp/hover.vim
    var popupAttrs = opt.PopupConfigure('Hover', {
      moved: 'any',
      close: 'click',
      fixed: true,
      maxwidth: 80,
      filter: HoverWinFilterKey
    })
    var winid = hoverText->popup_atcursor(popupAttrs)
```

jclsn:
Awesome! Works like a charm :)

Shane-XB-Qian:
> @Shane-XB-Qian<https://github.com/Shane-XB-Qian> Are you saying there are too many options for the user to configure, or too many locations in code that they are configured?

I meant both. those options can be simply same value, or need not to be configurable.
make it be KISS

--
shane.xb.qian

nickspoons:
@Shane-XB-Qian If we really want to keep it simple, we don't need to allow _any_ of this configuration, but it is nice to have the option to be able to configure your editor in the way you like, and this gives people flexibility to do that. Is there something specific you think is unnecessary here?

Shane-XB-Qian:
I think it had been answered why less cfg here. high weight lsp client there are already many ones.

--
shane.xb.qian

jclsn:
Thanks for inviting me to the repo! I just saw that you already took care of the documentation part though :) 

@Shane-XB-Qian: I see your point, but none of this is going to affect the performance of the plugin. Yes, other LSP plugins are bloated, but they are mainly because they many different programming languages to realize the LSP, whereas this one is native and blazing fast. This plugin is more or less the native/default LSP plugin for Vim and imo it should be just as configurable as Vim itself.

Shane-XB-Qian:
you just have answered the point: some cfg can leave it to vim itself.

P.S: if you noticed, I have been in this repo when it just a baby, so I think this better be KISS, tho all not on my decision.

--
shane.xb.qian

Shane-XB-Qian:
ok, let me repeat:
those almost same borderchar and borderhighlight, vs 'popup highlight' and 'border' maybe a bit different (specially 'signature'), which NO have to 4*7 = 28 options plus 4 option switches.
// and as @berggeist said above, vim may native have cfg to setup such also latter
so i think really no need to make it such complex, tho all not on my decision

Shane-XB-Qian:
> Can you then please make any concrete proposals for improvement! So far you are only commenting with buzzwords like KISS and it is hard to understand what you actually have in mind. This is not constructive!

'constructive' didnot mean 'you do not need to read carefully',
and it's not a job, please donot expect too much or play like you have paid something.. :smile:

nickspoons:
@jclsn I've taken your advice and reset the defaults to `Pmenu` highlighting and no borders, and signature help is no longer special-cased. I've also refactored the configuration into a function in options.vim as described in an earlier comment. This results in much simpler code, and brings all the generic popup configuration into one place.

I think this is pretty good. The popups are nice and configurable without excessive options. Users _may_ choose to use different highlights for different types of popups but those options are never named in code and don't add any maintenance burden.

If vim now introduces standardised popup border options, it will be simple to remove some of the options from this configuration and keep the highlight group configuration.

At the very least, this PR now brings the plugin into a cleaner state than it currently is in, dropping the `g:LspCompletionPopupBorderhighlight` options etc. and will be easier to clean up further if future vim options make the whole configuration aspect obsolete.

jclsn:
Only thing missing was the padding for the hover. I went ahead and added it back! Apart from that it looks good on my end!

nickspoons:
I made a couple more small changes, ensuring that when unconfigured the `:LspDocumentSymbol` and `:LspPeekImpl` layouts still worked correctly with their default borders, and that `:LspDocumentSymbol` padding is removed when borders _are_ enabled, keeping the layout consistent.

yegappan:
Thanks for working on this.

Shane-XB-Qian:
@nickspoons following patch may would make the option more strong, and kept its original value (if there), and later if there vim had native such option, maybe can do similar..
have not verified/tested it, please think if ok for you.

```diff
  export def PopupConfigure(type: string, popupAttrs: dict<any>): dict<any>
-  popupAttrs.highlight = PopupOptionGet(type, 'popupHighlight')
+  popupAttrs.highlight = popupAttrs->get('highlight') ?? PopupOptionGet(type, 'popupHighlight')
   if PopupOptionGet(type, 'popupBorder')
-    popupAttrs.border = []
-    popupAttrs.borderchars = lspOptions.popupBorderChars
-    popupAttrs.borderhighlight = [PopupOptionGet(type, 'popupBorderHighlight')]
+    popupAttrs.border = popupAttrs->get('border') ?? []
+    popupAttrs.borderchars = popupAttrs->get('borderchars') ?? lspOptions.popupBorderChars
+    popupAttrs.borderhighlight = popupAttrs->get('borderhighlight') ?? [PopupOptionGet(type, 'popupBorderHighlight')]
   endif
   return popupAttrs
 enddef
 ```

**Comment by nickspoons** [comment]
@Shane-XB-Qian I did try the same thing in a different way while developing:

```patch
 export def PopupConfigure(type: string, popupAttrs: dict<any>): dict<any>
-  popupAttrs.highlight = PopupOptionGet(type, 'popupHighlight')
+  var attrs = {
+    highlight: PopupOptionGet(type, 'popupHighlight')
+  }
   if PopupOptionGet(type, 'popupBorder')
-    popupAttrs.border = []
-    popupAttrs.borderchars = lspOptions.popupBorderChars
-    popupAttrs.borderhighlight = [PopupOptionGet(type, 'popupBorderHighlight')]
+    attrs.border = []
+    attrs.borderchars = lspOptions.popupBorderChars
+    attrs.borderhighlight = [PopupOptionGet(type, 'popupBorderHighlight')]
   endif
-  return popupAttrs
+  return extendnew(popupAttrs, attrs, 'keep')
 enddef
```
However it actually makes it impossible to add overrides for things that are set up by type.

For example, the "SymbolMenu" (`:LspDocumentSymbol`) is actually 2 popups, an "input" and a "names" list. They default to having a menu on the input but not on the names list, but when setting `{ popupBorder: true }` they now both get a border.
If we use either of our patches, then we can't override the `popupAttrs` options, so we can never remove the "input" border, or add a "names" popup border, as these are both set in the initial settings.

Shane-XB-Qian:
yes. I remember there two popups for symbol, perhaps it should be two label names for them too: e.g symbolinput, symbol names.

--
shane.xb.qian

nickspoons:
There _are_ 2 names for them, one is "SymbolMenuInput" and the other is "SymbolMenu". So we can style them individually, that's fine, you can see a screenshot in #629 where I have them displayed with different highlights.

But if we use either of the patches to let popups define their own base settings, we won't be able to add a border to the "SymbolMenu" (which has a `[0, 0, 0, 0]` border by default), or remove the default border from "SymbolMenuInput".

I suppose if we really wanted to, we could flip the conditions and _first_ check the user configuration in `opt.lspOptions`, and if not configured there, _then_ check for existing `popupAttrs` options:

```vim
  popupAttrs.highlight = get(lspOptions,
    'popupHighlight' .. type,
    get(popupAttrs, 'highlight',
      get(lspOptions, 'popupHighlight')))
```

But there doesn't seem to a need for it right now, I think it's better to leave it as it is, until there is a need for this.

Shane-XB-Qian:
"border" actually should work like popup "highlight" which *not controlled by label switch.
// I have said above which regarding the "specially" above comment.

--
shane.xb.qian

Shane-XB-Qian:
something like this:

```diff
export def PopupConfigure(type: string, popupAttrs: dict<any>): dict<any>
-  popupAttrs.highlight = PopupOptionGet(type, 'popupHighlight')
+  popupAttrs.highlight = popupAttrs->get('highlight') ?? PopupOptionGet(type, 'popupHighlight')
+  # popupAttrs.border = popupAttrs->get('border') ?? []
+  # popupAttrs.border = popupAttrs->get('padding') ?? []
   if PopupOptionGet(type, 'popupBorder')
-    popupAttrs.border = []
-    popupAttrs.borderchars = lspOptions.popupBorderChars
-    popupAttrs.borderhighlight = [PopupOptionGet(type, 'popupBorderHighlight')]
+    popupAttrs.borderchars = popupAttrs->get('borderchars') ?? lspOptions.popupBorderChars
+    popupAttrs.borderhighlight = popupAttrs->get('borderhighlight') ?? [PopupOptionGet(type, 'popupBorderHighlight')]
   endif
   return popupAttrs
 enddef
```

simple and clear~

Shane-XB-Qian:
@nickspoons *IF* https://github.com/vim/vim/pull/18487 merged eventually, perhaps this needed redo something per that.

PR #628, opened 2025-05-08
PR#627 lacks popup configuration support for CodeAction

berggeist:
This PR adds the popup configuration of border
and highlight to the CodeAction popup.


yegappan:
Thanks for the patch.

PR #629, opened 2025-05-08
Combine LspSymbol prop types

nickspoons:
This change allows base syntax-highlighting to bleed through the `LspSymbolName` and `LspSymbolName` highlights when using `:LspDocumentSymbol`:

```vim
colorscheme retrobox
hlset([
  {name: 'LspSymbolName', cterm: {bold: true}, guibg: '#303030'},
  {name: 'LspSymbolRange', cterm: {bold: true}, guibg: '#3c3836'}
])
```

Before:
image

After:
image

I can't think of any downside to using `combine: true` but I may have missed something?


yegappan:
Thanks for the patch.

PR #632, opened 2025-05-18
WIP: Add LSP pull diagnostic commands

berggeist:
LSP specification has introduced with version 3.17 new commands, which allow the client to pull diagnostics from the server. 
Earlier versions of the LSP specification just allowed the notification of diagnostics from the server to the client.
This methods are increasingly implemented by new LSP servers. 
In this first commit just the minimum set of changes has been added for : 'textDocument/diagnostic'.
Further commands like 'workspace/diagnostic' and 'workspace/diagnostic/refresh' would be work in progress.

Links to the LSP specification:
- [x] ['textDocument/diagnostic'](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_diagnostic)
- [x] ['workspace/diagnostic'](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_diagnostic)
- [x] ['workspace/diagnostic/refresh'](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic_refresh) 

What has been implemented so far:
- [x] Taking care whether a LSP server has the feature `diagnostic`, i.e. is able to act as `isDiagnosticProvider`
- [x] A new command `:LspDiagPull` that requests the set of diagnostics for the current buffer from all active LSP servers. This command might be used as part of an `autocmd`.
- [x] The asynchronously requested diagnostics are provided to the exisiting implementation within the file `autoload/lsp/diag.vim` for merging with other known diagnostics.
- [x] Related documents are now being processed, e.g. if a change in one file generates diagnostics in another file.
- [x] Workspace diagnostics are now implemented by processing all  diagnostics for affected files within the project
- [x] Two new commands have been introduced:
  - LspDiagPull = Pull document diagnostics from the server for the currently active buffer
  - LspDiagPullWorkspace = Pull workspace diagnostics from the server
 - [x] A new user event "LspWorkspaceDiagnosticRefresh" has been introduced, which is raised when a workspace/refresh is being notified from the server to the client

What would be still work in progress:
- [ ] Decide whether it is needed and feasible to obtain a cache for all versions of received diagnostics
      (this would be an improvement for an efficient reply of unchanged diagnostics by the server)
- [ ] Improve handling of related documents
- [ ] Better error handling
- [ ] Documentation

It is still not clear how to cope with the fact that the same server could provide diagnostics either by pull request and at the same time as notification.


autoload/lsp/capabilities.vim
```diff
@@ -319,6 +319,14 @@ export def ProcessServerCaps(lspserver: dict<any>, caps: dict<any>)
       endif
     endif
   endif
+
+  # diagnosticProvider
+  if lspserver.caps->has_key('diagnosticProvider')
+    lspserver.isDiagnosticProvider = true
+  else
+    lspserver.isDiagnosticProvider = false
+  endif
```
pusewicz, line 328: Would this not be the same and also shorter?

```suggestion
  lspserver.isDiagnosticProvider = lspserver.caps->has_key('diagnosticProvider')
```

yegappan:
The `foreach` function is available only in recent versions of Vim.  The older 9.x version doesn't have this function.
Can you rewrite the code without using the `foreach` function?

berggeist:
> The `foreach` function is available only in recent versions of Vim. The older 9.x version doesn't have this function. Can you rewrite the code without using the `foreach` function?

Thank you for your reply, I will fix this accordingly in the next draft version.

yegappan:
@berggeist I have added the support for pull diagnostics in 50d4b60c59c623a1ba7745f36f2f8d4912455e3f.  This is triggered automatically if the server supports it.  Can you try the latest version of the plugin and see whether you see any problems with this? Thanks

yegappan:
Thanks for the contribution.  The support for pull diagnostics has been added to the plugin now.  So closing this PR.

PR #640, opened 2025-07-17
Support sending custom requests to the language server

jclsn:
I am currently adding [vim support for the pkl-neovim plugin](https://github.com/apple/pkl-neovim/pull/35). It also exposes commands for these requests for nvim-lsp

https://pkl-lang.org/lsp/current/integration.html

Currently this is not possible in lsp and I would like to have the feature. Suggestions for improvement welcome!

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -26,6 +26,7 @@ import './callhierarchy.vim' as callhier
 import './typehierarchy.vim' as typehier
 import './inlayhints.vim'
 import './semantichighlight.vim'
+import './buffer.vim' as buf
 
 # LSP server standard output handler
 def Output_cb(lspserver: dict<any>, chan: channel, msg: any): void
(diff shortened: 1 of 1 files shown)


berggeist:
Hi, 
I don't want to argue, but out of interest isn't that what the language server protocols

-  [Code Lens Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeLens)
-  [Code Action Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction)

are supposed to solve in a standard way instead of implementing custom requests?

The LSP servers provide sets of CodeLense or CodeAction, which can be triggered via the above linked standard way.

I'm using CodeActions a lot and this mechanism works like a charm without any customization.

Here examples for `gopls` provided CodeLense:


image



And here examples for CodeAction:


image



Even if you need some custom handling of the replies of a CodeAction ,this Vim lsp server provides a mechanism to register custom CodeAction handlers. Just  take a look at `:h lsp-custom-commands`.

jclsn:
> Hi, I don't want to argue, but out of interest isn't that what the language server protocols
> 
>     * [Code Lens Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeLens)
>     * [Code Action Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction)
> 
> are supposed to solve in a standard way instead of implementing custom requests?

Sure, but some developers of language servers choose to implement non-default requests. In this case, the request is not dependent on the code, but he project folder. You could argue that this is nothing that the language server should handle, but rather the command-line client and I would agree.

For the pkl-lsp above, to download a package for example you need to send

```vim
call g:LspCustomRequst("pkl/downloadPackage", "package://pkg.pkl-lang.org/pkl-pantry/icalendar@0.1.1")
```

which is a string not a dictionary. In some cases you send nothing, a text document identifier or dictionaries. 

```vim
call g:LspCustomRequst("pkl/syncProjects", 0)
```

This is why I have chosen type any as payload here. Furthermore the message needs to be flexible, which I don't see in the Code Lense or Code Action Requests. I may be wrong. If you can show me how to implement the above commands otherwise, I am an open for it!

PR #646, opened 2025-08-15
Add full diagnostics to location-list

berggeist:
This is just a small enhancement that adds the complete information of diagnostic messages
as `user_data` to location list entries. 
It works only in combination with the lsp-option `autoPopulateDiags`.

Of course this comes at the expense of additional memory usage!

Background: From time to time I'm experimenting with alternative UI's, i.e. which are not built
into this lsp library, for displaying diagnostic messages and for this I need access to most of the
information provided by the diagnostic messages which are unfortunately not included within
the standard location list entries.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -550,7 +550,8 @@ def DiagsUpdateLocList(bnr: number, calledByCmd: bool = false): bool
 		    end_lnum: d_end.line + 1,
                     end_col: util.GetLineByteFromPos(bnr, d_end) + 1,
 		    text: text,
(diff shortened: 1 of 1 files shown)


berggeist:
@yegappan Please let me know if this PR is something that is worth to be merged into the main branch, then I would also add some documentation to the existing option `autoPopulateDiags`, e.g.

> The complete information of diagnostic messages received for the lsp server
> can be access on each location list entry via its attribute `user_data`.
> A dictionary `{ 'diagnostic': <complete diag message> }` is being added to the
> location list attribute `user_data`

PR #652, opened 2025-08-27
Explicitly clear text properties before deleting lines to prevent Vim E340, E685

berggeist:
Vim has issues when buffer lines are deleted, which have text properties assigned to them.

See:

- [internal errors after removing multiline text property](github.com/vim/vim/issues/12568)
- [Internal errors E340 E685 when buffer lines are deleted, which have text properties assigned](https://github.com/vim/vim/issues/18136)

This might even lead to segfaulting Vim.

Solution:
In order to prevent such an issue the text properties are explicitly cleared before the lines are deleted.

References:

[internal errors after removing multiline text property](github.com/vim/vim/issues/12568)
[Internal errors E340 E685 when buffer lines are deleted, which have text properties assigned](https://github.com/vim/vim/issues/18136)


berggeist:
See also:

https://github.com/vim/vim/issues/16448

PR #662, opened 2025-09-15
Update dictionaries to fix syntax errors in LSP configs

liamgines:
Added a few missing commas and hashes in dictionaries passed to LspAddServer.


Konfekt:
Thank you. One might wonder why this config follows the more cumbersome legacy Vim script syntax instead of that of Vim 9 (since this is a Vim 9 plugin)

yegappan:
Thanks for the patch.

> Thank you. One might wonder why this config follows the more cumbersome legacy Vim script syntax instead of that of Vim 9 (since this is a Vim 9 plugin)

A user may add the language server configuration in the .vimrc file which is a legacy vimrc file.
To support this, the examples are using the legacy Vim script.

PR #663, opened 2025-09-21
feat: Add custom callback variant of LspRequestCustom

blami:
This change allows to provide callback function to custom requests.

Builds on #640 to accommodate scenarios where response is needed from server but cannot be handled by custom handler (as there's no `method` field).

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -1037,18 +1037,26 @@ def FindLocations(lspserver: dict<any>, peek: bool, method: string, args: dict<a
   symbol.ShowLocations(lspserver, reply.result, peek, 'Symbol Locations')
 enddef
 
-# send a custom request to the server
+# send a custom async request to the server
 # Name: name of the server
 # Request: any
 # Params: any
-def g:LspRequestCustom(name: string, msg: string, params: any): string
+def g:LspRequestCustom(name: string, msg: string, params: any): void
+  g:LspRequestCustomCb(name, msg, params, WorkspaceExecuteReply)
+enddef
+
+# send a custom async request to the server with custom callback
+# Name: name of the server
+# Request: any
+# Params: any
+# Cbfunc: callback function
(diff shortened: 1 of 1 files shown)


jclsn (changes requested):
Afaik a funcref needs to exist. For every call without a callback, you would have to create and empty function to pass as argument. So I think it would be better to have a separate function for this. Not a 100% sure though.

blami:
@jclsn Thanks for review, I agree with you. I like approach with `*Cb()` callback variant. Also I changed the non-callback to just call callback variant providing original `WorkspaceExecuteReply` so that if logic ever changes fix will be in single place. 

The only other option I can think of is to have variadic argument and check for emptyness but that feels too fragile and obscure to me.

Let me know what do you think now.

autoload/lsp/lspserver.vim
```diff
@@ -1037,18 +1037,26 @@ def FindLocations(lspserver: dict<any>, peek: bool, method: string, args: dict<a
...
+def g:LspRequestCustom(name: string, msg: string, params: any): void
+  g:LspRequestCustomCb(name, msg, params, WorkspaceExecuteReply)
+enddef
+
+# send a custom async request to the server with custom callback
+# Name: name of the server
+# Request: any
+# Params: any
+# Cbfunc: callback function
+def g:LspRequestCustomCb(name: string, msg: string, params: any, Cbfunc: func): void
```
yegappan, line 1053: Can you update the help text with some details about how to use this new function?

blami: @yegappan  @jclsn  Sure I can. I actually did bit of investigation and learning and feel in my specific usecase this should not be needed and it is language server for which I do this being wrong and anything that edits workspace should be really code action and not custom request.

Anyway I think there's still some use for this and `g:LspRequestCustom`. As I got more familiar with codebase I now think that rather than making both these `g:` it would be actually better to export in `lsp.vim` as in case of other functions that are used by commands.

Issue #623, opened 2025-05-04
Signature help mapping conflicts with other mappings to ( or {


jclsn:
I use this old [vim-PairTools](https://github.com/MartinLafreniere/vim-PairTools) plugin for per-filetype automatic pair insertion.

For some reason the plugin breaks pair insertion for ``(`` and ``{`` characters. Any idea why?

jclsn:
Seems like it is the ``LspShowSignature`` mapping. Would be nice to have both actually.

jclsn:
I asked the coc.nvim developers how they handle it. Seems they check if the last inserted character matches the trigger char and the do an ``InsertCharPre``.

Issue #638, opened 2025-07-13
Cant able to use autopairs


Omkamble172:
I am using clangd and whenever i edit my .c files autopairs dont work. what can be issue.

LunarWatcher:
See #623

PR #668, opened 2025-09-26
signature: Fix conflict with autopair plugins

jclsn:
Since autopair plugins usually map the trigger chars as well, using imap does create a conflict here. This addresses this by emulating the functionality with autocmds.

Fixes https://github.com/yegappan/lsp/issues/623 and https://github.com/yegappan/lsp/issues/638

There might be better solutions. Please give your two cents!

### DIFF
--- a/autoload/lsp/signature.vim
+++ b/autoload/lsp/signature.vim
@@ -59,21 +59,35 @@ export def BufferInit(lspserver: dict<any>)
     return
   endif
 
+  # Add and clear the augroup for the signature help
   augroup lspShowSignatureHelp
     au!
+  augroup END
+
+  # Use a mapping for versions that don't support KeyInputPre yet'
+  if v:version < 901 || (v:version == 901 && !has('patch0563'))
+    # map characters that trigger signature help
+    for ch in lspserver.caps.signatureHelpProvider.triggerCharacters
+      var mapChar = ch
+      if ch =~ ' '
+	mapChar = '<Space>'
+      endif
+      exe $"inoremap <buffer> <silent> {mapChar} {mapChar}<C-R>=g:LspShowSignature()<CR>"
+    endfor
+  else
     # detect the trigger chars and show the signature
     autocmd_add([{bufnr: bufnr(),
 		  event: 'KeyInputPre',
 		  group: 'lspShowSignatureHelp',
 		  cmd: $'if index({lspserver.caps.signatureHelpProvider.triggerCharacters}, v:char) != -1
 			\ | call timer_start(1, function("LspShowSignatureCb")) | endif'}])
+  endif
 
-    # close the signature popup when leaving insert mode
-    autocmd_add([{bufnr: bufnr(),
-		  event: 'InsertLeave',
-		  group: 'lspShowSignatureHelp',
-		  cmd: 'CloseCurBufSignaturePopup()'}])
-  augroup END
+  # close the signature popup when leaving insert mode
+  autocmd_add([{bufnr: bufnr(),
+		event: 'InsertLeave',
+		group: 'lspShowSignatureHelp',
+		cmd: 'CloseCurBufSignaturePopup()'}])
 enddef
 
 # process the 'textDocument/signatureHelp' reply from the LSP server and


autoload/lsp/signature.vim
```diff
@@ -55,15 +74,22 @@ export def BufferInit(lspserver: dict<any>)
...
   endif
 
+  var triggerChars = lspserver.caps.signatureHelpProvider.triggerCharacters
+
   # map characters that trigger signature help
-  for ch in lspserver.caps.signatureHelpProvider.triggerCharacters
+  for ch in triggerChars
     var mapChar = ch
     if ch =~ ' '
       mapChar = '<Space>'
```
yegappan, line 81: The conversion of the space character to `<Space>` is needed to map the space character below using `inoremap`.
This is not needed if `InsertCharPre` is used to check for the character.

autoload/lsp/signature.vim
```diff
@@ -55,15 +74,22 @@ export def BufferInit(lspserver: dict<any>)
...
     endif
-    exe $"inoremap <buffer> <silent> {mapChar} {mapChar}<C-R>=g:LspShowSignature()<CR>"
+    # triggers without mappings to not conflict with automatic pair-closing
+    # plugins like autopairs.vim or vim-PairTools
+    exe $'autocmd InsertCharPre <buffer> if v:char ==# "{mapChar}" | call timer_start(1, function("LspShowSignatureCb")) | endif'
```
yegappan, line 85: I am concerned about using multiple `InsertCharPre` autocmds for every buffer (one per trigger character).  
Is it possible to use a single autocmd to check whether `v:char` is in `triggerChars`?

autoload/lsp/signature.vim
```diff
@@ -41,6 +41,25 @@ export def InitOnce()
...
+enddef
+
+# Checks if the cursor moved over a trigger character to
+# not break autopair plugins
+def DetectJumpOverTriggerChar(triggerChars: list<string>)
+  var col = col('.')
+  var lineStr = getline('.')
+
+  for ch in triggerChars
+    if ch == ','
```
yegappan, line 55: What is the need for checking the `,` character here?

autoload/lsp/signature.vim
```diff
@@ -41,6 +41,25 @@ export def InitOnce()
...
+      break
+    elseif lineStr[col - 2] == ch
```
yegappan, line 57: Can you retrieve the character from `lineStr[col - 2]` outside the loop and use it here?

autoload/lsp/signature.vim
jclsn: Without it the signature help will close when you move over the `,` character with the arrow keys. 

See this example:

```c
// while the signature help is open and you are inside the parentheses, after moving over the , 
// from left to right, the signature help would close
printf("val = %d\n, myval);
```

autoload/lsp/signature.vim
jclsn: Removed the portion

autoload/lsp/signature.vim
jclsn: I am honestly not sure what's the better solution here:

Yes, `autocmd InsertCharPre` does register autocommands for all the trigger characters for every buffer, but also just fires, when those trigger characters are detected.

`autocmd CursorMovedI` does fire on every cursor movement in insert mode (also typing), which is also not ideal.

There doesn't seem to be an autocommand that triggers on a range a list of characters. I also don't think this would be much different from the single autocommands.

Furthermore, I think this would need an `augroup` for robustness wouldn't it?

autoload/lsp/signature.vim
jclsn: Removed

jclsn:
The `KeyInputPre` autocmd doesn't seem to be recognized by the vim version in the pipeline according to Copilot

jclsn:
I see! I can also change this to use a `CursorMovedI` that challs a function that checks for trigger chars under the cursor and then calls `LspShowSignature()`. The current version is way cleaner though. I could also add a version check and implement both or keep your old version and use the new one if greater than 9.1.0563.

Let me know what you prefer!

PR #672, opened 2025-09-30
signature: Show signature help only in insert mode

jclsn:
Without this check, the signature help is also triggered when pressing K when hovering.

### DIFF
--- a/autoload/lsp/signature.vim
+++ b/autoload/lsp/signature.vim
@@ -42,7 +42,10 @@ export def InitOnce()
 enddef
 
 def LspShowSignatureCb(timer: number)
-  call g:LspShowSignature()
+  # Show signature only in insert mode
+  if mode() == 'i'
+    call g:LspShowSignature()
+  endif
 enddef
 
 # Initialize the signature triggers for the current buffer


jclsn:
```c
#include <stdio.h>

int main(int argc, char *argv[])
{
	int myint = 0;

	printf("test %d\n", myint);
	return 0;
}
```

grafik

jclsn:
It seems like this is a problem with `KeyInputPre`. It happens when you type :( in the command line. The mapping causes `:call SmartHover()` to be entered in the command line and therefore the signature help is triggered by the parentheses. I do think this fix is needed.

yegappan:
Can we use the `InsertCharPre` autocmd instead of the `KeyInputPre` autocmd to avoid this problem?

jclsn:
If we use `InsertChartPre`  we won't catch if a autopair plugin jumps over a character. They usually filter e.g. the closing parenthesis and jump over it. So no trigger char is recognized. So we would have to use this in combination with `CursorMovedI` and a filter function or just use `CursorMovedI`. I do think `KeyInputPre` is cleaner, but up to you.

jclsn:
Sure, I will. Thanks.

PR #691, opened 2025-10-28
Set correct width for rendering markdown

balki:
`winwidth()` includes signcolumn, numbercolumn, foldcolumns etc. This is especially bad in rust and using preview window. All the horizontal lines (`---`) are wrapped.

### DIFF
--- a/ftplugin/lspgfm.vim
+++ b/ftplugin/lspgfm.vim
@@ -57,7 +57,8 @@ def RenderGitHubMarkdownText()
     endif
     # parse the github markdown content and convert it into a list of text and
     # list of associated text properties.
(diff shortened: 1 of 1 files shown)


balki:
### Before
image

### After
image

### vimrc
```
source $VIMRUNTIME/defaults.vim

set number signcolumn=yes foldcolumn=1

packadd lsp

let lspOpts = #{hoverInPreview: v:true}
call LspOptionsSet(lspOpts)

" Rust language server
call LspAddServer([#{
        \    name: 'rustlang',
        \    filetype: ['rust'],
        \    path: 'rust-analyzer',
        \    args: [],
        \    syncInit: v:true
        \  }])

```

PR #695, opened 2025-11-01
add option (off by default) to fall back to built-in formatter

Konfekt:
completes [the previous PR](https://github.com/yegappan/lsp/pull/689)

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -1033,8 +1033,10 @@ export def TextDocFormat(range_args: number, line1: number, line2: number)
       if canFallback
 	util.WarnMsg('Formatting unsupported; falling back to built-in.')
 	execute $'keepjumps normal! {line1}Ggq{line2}G'
+      elseif !&filetype->empty()
+	util.ErrMsg($'Language server for "{&filetype}" file type supporting "documentRangeFormatting" feature is not found')
       else
-	util.ErrMsg($'Language server for "{&filetype}" file type supporting documentRangeFormatting feature is not found')
+	util.ErrMsg('No language server supporting "documentRangeFormatting" feature is found')
       endif
     else
       lspserver.textDocFormat(fname, true, line1, line2)
@@ -1045,8 +1047,10 @@ export def TextDocFormat(range_args: number, line1: number, line2: number)
       if canFallback
 	util.WarnMsg('Formatting unsupported; falling back to built-in.')
 	execute 'keepjumps normal! 1GgqG'
-      else
-	util.ErrMsg($'Language server for "{&filetype}" file type supporting documentFormatting feature is not found')
+      elseif !&filetype->empty()
+	util.ErrMsg($'Language server for "{&filetype}" file type supporting "documentFormatting" feature is not found')
+      # else
+	# util.ErrMsg('No language server supporting "documentRangeFormatting" feature is found')
       endif
     else
       lspserver.textDocFormat(fname, false, 0, 0)


autoload/lsp/lspserver.vim
```diff
@@ -1159,7 +1159,18 @@ def TextDocFormat(lspserver: dict<any>, fname: string, rangeFormat: bool,
 				start_lnum: number, end_lnum: number)
   # Check whether LSP server supports formatting documents
   if !lspserver.isDocumentFormattingProvider
-    util.ErrMsg('LSP server does not support formatting documents')
+    if &formatexpr !=# 'lsp#lsp#FormatExpr()' && opt.lspOptions.formatFallback
```
yegappan: Instead of duplicating this code in lsp.vim and lspserver.vim, can you move this to a function in util.vim and use that?

Konfekt:
Unsure whether making that test pass is the way to go; the idea is to avoid two warning messages in case fallback is enabled

autoload/lsp/hover.vim
```diff
@@ -107,7 +107,7 @@ export def HoverReply(lspserver: dict<any>, hoverResult: any, cmdmods: string):
   if hoverText->empty()
     if &keywordprg !=# ':LspHover' && !empty(&l:keywordprg) && opt.lspOptions.hoverFallback
       if cmdmods !~ 'silent'
-        util.WarnMsg($'No documentation found for current keyword; falling back to built-in.')
+        util.WarnMsg('No documentation found for current keyword; falling back to built-in.')
```
mmrwoods: Maybe remove from this PR, not directly related

autoload/lsp/lsp.vim
```diff
@@ -331,7 +331,7 @@ export def GotoDefinition(peek: bool, cmdmods: string, count: number)
   if lspserver->empty()
     if &tagfunc !=# 'lsp#lsp#TagFunc' && opt.lspOptions.definitionFallback
       if cmdmods !~ 'silent'
-      	util.WarnMsg($'definition lookup unsupported; falling back to tags file')
+      	util.WarnMsg('definition lookup unsupported; falling back to tags file')
```
mmrwoods: Maybe remove from this PR, not directly related

autoload/lsp/lsp.vim
```diff
@@ -875,7 +875,7 @@ export def Hover(cmdmods: string)
   if lspserver->empty()
     if &keywordprg !=# ':LspHover' && !empty(&l:keywordprg) && opt.lspOptions.hoverFallback
       if cmdmods !~ 'silent'
-      	util.WarnMsg($'Hovering unsupported; falling back to built-in.')
+      	util.WarnMsg('Hovering unsupported; falling back to built-in.')
```
mmrwoods: Maybe remove from this PR, not directly related

autoload/lsp/lsp.vim
```diff
@@ -1019,16 +1019,18 @@ export def TextDocFormat(range_args: number, line1: number, line2: number)
     return
   endif
 
-  var lspserver: dict<any> = buf.CurbufGetServerChecked('documentFormatting')
+  silent var lspserver: dict<any> = buf.CurbufGetServerChecked('documentFormatting')
```
mmrwoods: Using `:silent` will suppress all errors, suggest using `buf.BufLspServerGet(bufnr(), 'documentFormatting')` without `:silent` instead

mmrwoods:
btw, branch rebased on `main` is here: https://github.com/mmrwoods/lsp/tree/fallbacks

I reverted the commit to "remove redundant format string" as it was unrelated to this PR, and squashed the remaining commits before rebasing.

autoload/lsp/util.vim
```diff
@@ -363,4 +363,22 @@ export def FindNearestRootDir(startDir: string, files: list<any>): string
...
+import './options.vim' as opt
+
+export def TextDocFormatFallback(range: bool, line1: number, line2: number): bool
+  if &formatexpr !=# 'lsp#lsp#FormatExpr()' && opt.lspOptions.formatFallback
+    WarnMsg('Formatting unsupported; falling back to built-in.')
+    const line_start = range ? 1 : line1
+    const line_end = range ? line('$') : line2
+    try
+      execute 'normal!' line_start .. 'Ggq' .. line_end .. 'G'
+    catch /.*/
```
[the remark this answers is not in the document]
Konfekt: You're right here; again I can't remember

autoload/lsp/lsp.vim
```diff
@@ -1019,16 +1019,18 @@ export def TextDocFormat(range_args: number, line1: number, line2: number)
...
   if lspserver->empty()
-    return
-  endif
-
-  var fname: string = @%
-  if range_args > 0
-    lspserver.textDocFormat(fname, true, line1, line2)
+    if !util.TextDocFormatFallback(range_args == 0, line1, line2)
```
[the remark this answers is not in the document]
Konfekt: I split the function up in check and format

Konfekt:
Since https://github.com/yegappan/lsp/actions/runs/20327947245/job/58396885730#step:5:815 calls for action, should https://github.com/yegappan/lsp/blob/87d179a5e3c89fe7838a58724c6ba916b34685fa/test/clangd_tests.vim#L92 be satisfied or adapted?

Konfekt:
> The behaviour is a little strange when document formatting is supported, but range formatting is not. It works as expected using motions in normal mode, writing a warning message and falling back to formatprg if set or internal formatting otherwise, but in visual mode just writes a warning message to say falling back to built-in, and then does nothing.

This should be addressed, thank you for pointing that out. One can imagine that one should check for mode and, say, instead work on the visual selection `'<,'>` instead at 

https://github.com/yegappan/lsp/pull/695/changes#diff-89400e0c68cc41fcdae808ce36f81438f12b87b4e17c2210b5980f9fd9b9fa15R375

autoload/lsp/lsp.vim
mmrwoods: It was necessary to hide the warning message from `CurbufGetServerChecked()` when the feature is not supported

autoload/lsp/lsp.vim
```diff
@@ -1025,16 +1025,31 @@ export def TextDocFormat(range_args: number, line1: number, line2: number)
...
-  if lspserver->empty()
-    return
-  endif
-
   var fname: string = @%
   if range_args > 0
-    lspserver.textDocFormat(fname, true, line1, line2)
+    var lspserver: dict<any> = buf.BufLspServerGet(bufnr(), 'documentRangeFormatting')
+    if lspserver->empty()
+      if util.TextDocFormatFallbackSupported()
```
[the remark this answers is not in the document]
mmrwoods: Or a `const` ;-)

mmrwoods:
@Konfekt Indentation seems to have gotten messed up, do you have modelines enabled?

Here's a patch to fix it anyway...

```diff
diff --git a/autoload/lsp/lsp.vim b/autoload/lsp/lsp.vim
index 25f0156..10eb3fd 100644
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -1031,10 +1031,10 @@ export def TextDocFormat(range_args: number, line1: number, line2: number)
     var lspserver: dict<any> = buf.BufLspServerGet(bufnr(), 'documentRangeFormatting')
     if lspserver->empty()
       if canFallback
-        util.WarnMsg('Formatting unsupported; falling back to built-in.')
-  	execute 'keepjumps normal!' line1 .. 'Ggq' .. line2 .. 'G'
+	util.WarnMsg('Formatting unsupported; falling back to built-in.')
+	execute 'keepjumps normal!' line1 .. 'Ggq' .. line2 .. 'G'
       else
-        util.ErrMsg($'Language server for "{&filetype}" file type supporting documentRangeFormatting feature is not found')
+	util.ErrMsg($'Language server for "{&filetype}" file type supporting documentRangeFormatting feature is not found')
       endif
     else
       lspserver.textDocFormat(fname, true, line1, line2)
@@ -1043,10 +1043,10 @@ export def TextDocFormat(range_args: number, line1: number, line2: number)
     var lspserver: dict<any> = buf.BufLspServerGet(bufnr(), 'documentFormatting')
     if lspserver->empty()
       if canFallback
-        util.WarnMsg('Formatting unsupported; falling back to built-in.')
-  	execute 'keepjumps normal! 1GgqG'
+	util.WarnMsg('Formatting unsupported; falling back to built-in.')
+	execute 'keepjumps normal! 1GgqG'
       else
-        util.ErrMsg($'Language server for "{&filetype}" file type supporting documentFormatting feature is not found')
+	util.ErrMsg($'Language server for "{&filetype}" file type supporting documentFormatting feature is not found')
       endif
     else
       lspserver.textDocFormat(fname, false, 0, 0)
```

plugin/lsp.vim
```diff
@@ -76,8 +76,8 @@ command! -nargs=0 -bar LspDocumentSymbol lsp.ShowDocSymbols()
 command! -nargs=0 -bar LspFold lsp.FoldDocument()
 
 command! -nargs=0 -bar -range=% LspFormat lsp.TextDocFormat(<range>, <line1>, <line2>)
-def g:LspFormatFunc(type: string, visual_mode = v:false)
-  if visual_mode
+def g:LspFormatFunc(type: string)
+  if type ==# 'block'
```
yegappan: This change is already merged in 958a1457b4f1eb197c544946145bc4faa017be74.

Konfekt:
So I read addressing a test failure rather as making it pass by changing the tested code than the test itself, and regarding

```
command line..script /home/runner/work/lsp/lsp/test/runner.vim[48]..function <SNR>1_LspRunTests[20]..Test_LspFormat line 91: Expected '' but got '\nError: Language server for "" file type supporting documentFormatting feature is not found'
```

there be given no error message at all? I commented it out accordingly.

Regarding the now appearing

```
command line..script /home/runner/work/lsp/lsp/test/runner.vim[48]..function <SNR>1_LspRunTests[20]..Test_LspFormat line 58: Expected [''] but got ['', '', '', '']
Test_LspFormat(): FAIL
```

I am a bit at loss where that came from as the code should not touch this test branch and I cannot reproduce

autoload/lsp/lsp.vim
```diff
@@ -1025,16 +1025,36 @@ export def TextDocFormat(range_args: number, line1: number, line2: number)
...
-    lspserver.textDocFormat(fname, true, line1, line2)
+    var lspserver: dict<any> = buf.BufLspServerGet(bufnr(), 'documentRangeFormatting')
+    if lspserver->empty()
+      if canFallback
+	util.WarnMsg('Formatting unsupported; falling back to built-in.')
+	execute $'keepjumps normal! {line1}Ggq{line2}G'
+      elseif !&filetype->empty()
+	util.ErrMsg($'Language server for "{&filetype}" file type supporting "documentRangeFormatting" feature is not found')
+      else
+	util.ErrMsg('No language server supporting "documentRangeFormatting" feature is found')
```
mmrwoods: This seems a possible improvement to me, but it is an isolated one, so this PR might not be the place to make it. Many LSP commands do nothing when no language servers are configured for the current buffer, e.g. `:LspHover`, `:LspFold`. Some commands print a warning though, e.g. `:LspDiagCurrent`, so behaviour is not fully consistent, but adding another behaviour would make it even less consistent, so maybe best to make no changes to existing behaviour, i.e. when no language servers found supporting required formatting feature, and fallback is not enabled, do nothing, as before.

mmrwoods:
@yegappan I checked out this PR locally and ran the tests and I also cannot reproduce the failure on CI. Tested with Vim 9.1.850 on MacOS and 9.1.697 on Ubuntu (docker container, noble image).

autoload/lsp/lsp.vim
yegappan: If the user invokes a command and the language server that provides the feature is not available, it is helpful to give the user feedback about the error. However, if the command is triggered by an autocmd or is executed frequently, showing an error or warning every time can become noisy, and users may find it undesirable.

yegappan:
> @yegappan I checked out this PR locally and ran the tests and I also cannot reproduce the failure on CI. Tested with Vim 9.1.850 on MacOS and 9.1.697 on Ubuntu (docker container, noble image).

In my local system also the tests are passing with this PR.  I will look into the CI failure.

autoload/lsp/lsp.vim
mmrwoods: I'm with you on autocmds, I would not be suggesting any plugin defined autocmds write warning messages when a feature is not supported. But when a user directly invokes an `:Lsp...` command, it would be good to consistently provide feedback to the user when the requested feature is not supported, right?

I'm not sure how you can determine whether a command is executed frequently without actually recording data about invocations and maybe only showing a message the first time, and that sounds unnecessarily complex. TBH, I'm struggling to understand why a user would choose to execute a command frequently, after being informed it is not supported, other than by accident, in which case a warning still seems sensible to me.

For user defined autocmds, there is always `:silent` which will silence lsp warnings, but not actual errors.

Shane-XB-Qian:
> @yegappan commented on this pull request.
> 
> >    if range_args > 0
> -    lspserver.textDocFormat(fname, true, line1, line2)
> +    var lspserver: dict<any> = buf.BufLspServerGet(bufnr(), 'documentRangeFormatting')
> +    if lspserver->empty()
> +      if canFallback
> +	util.WarnMsg('Formatting unsupported; falling back to built-in.')
> +	execute $'keepjumps normal! {line1}Ggq{line2}G'
> +      elseif !&filetype->empty()
> +	util.ErrMsg($'Language server for "{&filetype}" file type supporting "documentRangeFormatting" feature is not found')
> +      else
> +	util.ErrMsg('No language server supporting "documentRangeFormatting" feature is found')
> 
> If the user invokes a command and the language server that provides the feature is not available, it is helpful to give the user feedback about the error. However, if the command is triggered by an autocmd or is executed frequently, showing an error or warning every time can become noisy, and users may find it undesirable.

Another case: if `FormatExpr()` return `1`, it looks would re-try, so such `warning` would keep (some dup) echoing.
// and `!&filetype->empty()` is it required? Maybe warning of DocFormat() and DocRangeFormat() should be uniformed?

-- 
shane.xb.qian

Issue #688, opened 2025-10-26
How to show autocomplete doc on preview window instead of popup?


813ethan:
There's an option to show the documentation of `:LspHover` on a preview window instead of a popup,

is there a way to achieve that with the documentation popup shown when selecting autocompletion keywords?

Personally im using this plugin in Android Termux, the autocompletion doc seems to be truncated even after enabling condensed autocomplete menu

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/4b3ce642-f38a-4ea5-8ca2-cbbad187dd8f" />

putting the doc on a separate panel will be a good option, if there's a way to accomplish it

Konfekt:
See `:help 'completeopt'` which should default to include `preview`. If this plug-in overrides it, a PR could take care of a corresponding option

PR #696, opened 2025-11-02
feat: add completionInPreview option and preview filetype handling

Konfekt:
A stab at https://github.com/yegappan/lsp/issues/688 ; if @813ethan could test and comment

### DIFF
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -648,25 +648,26 @@ def LspSetPopupFileType()
   endif
 enddef
 
+def ClosePreviewWindow()
+  try
+    :pclose
+  catch /E441/ # No preview window
+  endtry
+enddef
+
 # complete done handler (LSP server-initiated actions after completion)
 def LspCompleteDone(bnr: number)
   var lspserver: dict<any> = buf.BufLspServerGet(bnr, 'completion')
   if lspserver->empty()
     if opt.lspOptions.completionInPreview && opt.lspOptions.closePreviewOnComplete
-      try
-      	:pclose
-      catch /E441/ # No preview window
-      endtry
+      ClosePreviewWindow()
     endif
     return
   endif
 
   if v:completed_item->type() != v:t_dict
     if opt.lspOptions.completionInPreview && opt.lspOptions.closePreviewOnComplete
-      try
-      	:pclose
-      catch /E441/ # No preview window
-      endtry
+      ClosePreviewWindow()
     endif
     return
   endif
@@ -675,10 +676,7 @@ def LspCompleteDone(bnr: number)
   if completionData->type() != v:t_dict
       || !opt.lspOptions.completionTextEdit
     if opt.lspOptions.completionInPreview && opt.lspOptions.closePreviewOnComplete
-      try
-      	:pclose
-      catch /E441/ # No preview window
-      endtry
+      ClosePreviewWindow()
     endif
     return
   endif
@@ -701,10 +699,7 @@ def LspCompleteDone(bnr: number)
 
   # Close the preview window automatically after completion
   if opt.lspOptions.completionInPreview && opt.lspOptions.closePreviewOnComplete
-    try
-      :pclose
-    catch /E441/ # No preview window
-    endtry
+    ClosePreviewWindow()
   endif
 enddef
 


Konfekt:
I have no idea why the preview window height is not sufficiently high; by default it should be (=12) and Vim respect it?

813ethan:
One more problem, there's this error showing up when im using preview with clangd?

Screenshot_20251105-151838_Termux

this is my lsp config:
```
{
          'name': 'clangd',
          'filetype': ['c', 'cpp'],
          'path': TERMUX_PREFIX .. '/usr/bin/clangd',
          'args': ['--background-index', '--clang-tidy']
}
```

All good otherwise, the error didn't happen with other clients like `python-language-server`

813ethan:
https://groups.google.com/g/vim_dev/c/UX9-Dgw86xA
maybe this is related?

813ethan:
> Can you try to make sense of the solution [justmao945/vim-clang#140](https://github.com/justmao945/vim-clang/pull/140) ?

it tried to move the problematic calls away from the autocomplete function?

I'm not an expert about the internals of this plugin (lsp), so i don't know if there's a way to adapt that here, not counting that this plugin is not clangd specific 

yet something I noticed is that in `python-language-server`, `LspPopupFileType()` which causes the error was never called,

while `clangd` never uses lazy loading (`lspserver.completionLazyDoc`) and renders the documentation directly instead, maybe it's something about it?

813ethan:
my understanding is `LspSetPopupFileType()` is called to setup the completion popup's markdown highlighting, while it works for popup windows, it fails with preview windows bc of the error,

while lazy loading uses `ShowCompletionDocumentation()`, which also setups the markdown but didn't get any errors

maybe something with the implementation or timing  of those functions? I have no idea

Konfekt:
Thank you, so doing it the roundabout `ShowCompletionDocumentation()` way

```vim
#  autoload/lsp/completion.vim (lines 453-460)
  :wincmd P
  :setlocal modifiable
  bufnr()->deletebufline(1, '$')
  infoText->append(0)
  [1, 1]->cursor()
 setlocal ft=lspgfm
 :wincmd p
catch /E441/ # No preview window
```

solves that problem?

813ethan:
> ok nvm `try ... catch` E565 here silenced the errors, but markdown isn't applied to the preview window
> 
> https://github.com/yegappan/lsp/blob/5ded99487aab03513120a758fd5222d413284bf4/autoload/lsp/completion.vim#L631-L635
> 
>  screen-20251105-181844.2.mp4 
> the error still exists after removing `setlocal ft=lspgfm`

won't, the error will still happen with `:wincmd P`..

Konfekt:
Here's the corresponding fix [for php complete](https://github.com/vim/vim/pull/7799/files)

Konfekt:
For short, get the popup window id and then `call win_execute(..)` to change it, say `&filetype`. Could you look into that?

813ethan:
> > Thank you, so doing it the roundabout `ShowCompletionDocumentation()` way
> > ```viml
> > #  autoload/lsp/completion.vim (lines 453-460)
> >   :wincmd P
> >   :setlocal modifiable
> >   bufnr()->deletebufline(1, '$')
> >   infoText->append(0)
> >   [1, 1]->cursor()
> >  setlocal ft=lspgfm
> >  :wincmd p
> > catch /E441/ # No preview window
> > ```
> 
> How come then this doesn't give any errors ?

It wasn't run on `CompleteChanged`, it's a separate process called (during the completion?) from `lspserver.vim`

813ethan:
> > > Changing windows while inside a completion function is forbidden now.
> > 
> > 
> > I think changing means less ambiguously switching here; so we can still change, say the filetype, but have to stay inside the window
> 
> the buffer is still being changed via the markdown renderer (`lspgfm.vim`)
> 
> tried setting the file type indirectly via
> 
> ```
> setbufvar(winbufnr(winnr('#')), '&ft', 'lspgfm')
> ```
> 
> but it gives another error Screenshot_20251107-193318_Termux

This changes the `&filetype` of the preview buffer, without doing the `wincmdP/p` dance.

how can I get the window id for `win_execute(...)`?

813ethan:
setting modifiable upstream
```viml
setbufvar(bnr, '&modifiable', 1)
```
at `UpdatePreviewWindowContents()` gives another E565
Screenshot_20251107-204552_Termux

Konfekt:
Is only the latter or both patches needed?

813ethan:
both, the latter is for resizing the preview window

PR #697, opened 2025-11-06
Register autocmd only once per buffer

Ja-sonYun:
When restarting the LSP server with `LspServer restart`, some features such as autoimport are triggered multiple times.

This PR ensures `BufferInit` runs only once per buffer to prevent duplicate registrations.


Ja-sonYun:
```vim script
acmds->add({bufnr: bnr, replace: true, ...})
```

`once: true` didn't work, but `replace: true` work. And seems like this is better approach to do register once

PR #720, opened 2025-12-12
Check for documentRangeFormatting capability in FormatExpr()

mmrwoods:
I want to set `formatexpr=lsp#lsp#FormatExpr()`, but only if range formatting is supported, otherwise I want to fallback to use `formatprg` (if set)

This change allows me to check if range formatting is supported on LspAttached:

```vim
    autocmd User LspAttached
      \ if !lsp#buffer#BufLspServerGet(bufnr(), 'documentRangeFormatting')->empty() |
      \   setlocal formatexpr=lsp#lsp#FormatExpr() |
      \ endif
```

This is probably the wrong way to achieve this outcome, the functions exported from `autoload/lsp/buffer.vim` probably aren't intended as a public API, but it does work. Is there a better way?

### DIFF
--- a/autoload/lsp/buffer.vim
+++ b/autoload/lsp/buffer.vim
@@ -43,6 +43,7 @@ var SupportedCheckFns = {
   declaration: (lspserver) => lspserver.isDeclarationProvider,
   definition: (lspserver) => lspserver.isDefinitionProvider,
   documentFormatting: (lspserver) => lspserver.isDocumentFormattingProvider,
+  documentRangeFormatting: (lspserver) => lspserver.isDocumentRangeFormattingProvider,
   documentHighlight: (lspserver) => lspserver.isDocumentHighlightProvider,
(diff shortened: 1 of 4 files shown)


mmrwoods:
> @mmrwoods can we somehow converge here, as in first fallback yours, second my PR #695

Yes, I do really like the idea of simply configuring a fallback, but a check for `documentRangeFormatting` capability also seems necessary as it's required for `lsp#lsp#FormatExpr()` to work.

If I tidy this up so it's ready to merge, and restores the check for `documentRangeFormatting` in `lsp#lsp#FormatExpr()` such that it returns 1 when range formatting is not supported, I can try rebasing your PR on top and see how it goes.

yegappan:
> BTW, I have not yet said how delightful it is to use this plugin, it just works, the code is easy to follow, thank you :-)

Thank you.  Happy to hear that the plugin is working well for you.  It is always encouraging to hear feedback like this.

PR #741, opened 2026-01-07
Improve symbol search when needOffsetEncoding

### LINKED ISSUE #721: Symbol search unusable on large project without `forceOffsetEncoding: 'utf-32'`
Hi @yegappan,

Using typescript language server on a relatively sizeable project, I found that workspace symbol search was unusable as it was so slow to open the popup and when the popup eventually opens it appears to hang initially (keyboard input does not get handled by the filter function). 

This is because locations for all symbols returned by the server are decoded, before even trying to render the workspace symbol popup (which does not actually display the locations), and decoding locations is very costly as the code ends up calling `bufadd()` for each location.

Reducing the max number of symbols, by hacking the code to slice the results, e.g. `var symInfo: list<dict<any>> = reply.result->slice(0, 50)`, does improve things, but not hugely, decoding locations is just costly, and I think should ideally only be done at the point where it is necessary, in the case of workspace symbol search, when you want to jump to the location. 

I haven't made that change because I see there are a number of other places where results are iterated over and the locations decoded, e.g. `TagFunc`, so I guess addressing this needs a bit more thought rather than a quick fix only applied to workspace symbol search.

Happy to submit a PR if you can provide some guidance on how you think this might be refactored to be more efficient.

For now, forcing 'utf-32' encoding is enough to make symbol search perform adequately, and this will likely work ok for me in most places.

mmrwoods:
When the language server does not support utf-32 position encoding, workspace symbol search is painfully slow because locations for all symbols returned by the server are decoded before opening the popup, even though locations are only needed when jumping to a location.

Fix by delaying decoding the position until jumping to the location.

Fixes #721

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -1644,23 +1644,18 @@ def WorkspaceQuerySymbols(lspserver: dict<any>, query: string, firstCall: bool, 
 
   var symInfo: list<dict<any>> = reply.result
 
-  if lspserver.needOffsetEncoding
-    # Decode the position encoding in all the symbol locations
-    symInfo->map((_, sym) => {
-      if sym->has_key('location')
-	lspserver.decodeLocation(sym.location)
-      endif
-      return sym
-    })
-  endif
-
   if firstCall && symInfo->len() == 1
     # If there is only one symbol, then jump to the symbol location
     var symLoc: dict<any> = symInfo[0]->get('location', {})
     if !symLoc->empty()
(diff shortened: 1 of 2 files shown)


yegappan:
@mmrwoods Thanks for the patch.  I will test it out over the weekend.

yegappan:
@mmrwoods I have tested this change locally and it looks good to me.  Can you remove draft from this PR? Thanks.

mmrwoods:
Thanks for testing, now marked as ready

Issue #745, opened 2026-01-18
Trigger servers on explicit filenames


jclsn:
The docker language server should be triggered on `Dockerfile` and `docker-compose.yml`. There is a filetype `dockerfile`, but[ Christian is opposed to adding a duplicate filetype](https://github.com/vim/vim/pull/19200) `dockercompose` into upstream vim. So in this case it would be nice if triggering on certain filenames was possible

Example:

```vim
let docker_ls =
			\ #{name: 'docker-language-server',
			\   path: 'docker-language-server',
			\   args: ['start', '--stdio'],
			\   filetype: ['dockerfile'],
			\   filename: ['docker-compose.yaml', 'docker-compose.yml']
			\ }
```

PR #746, opened 2026-01-19
Only send the base filetype when a file has multiple filetypes

jclsn:
In case of a dockercompose file, which is technically yaml, it would make sense to set it's filetype to `dockercompose.yaml`. This commit makes sure only the basename is sent as languageId to the server.

Fixes https://github.com/yegappan/lsp/issues/745

This strictly needs the `languageId` to be the first filetype. So `dockercompose.yaml` works, but not `yaml.dockercompose`.

@yegappan Let me know if you would like a more flexible solution! If this is okay, I would add another commit to document this.


jclsn:
I realized that overriding the filetype would break the yaml language server. Since currently multiple servers can't run for the same filetype anyway, this would need to get fixed first.

PR #752, opened 2026-01-25
Add option for disabling syntax overriding for `LspDiagInline*` when 'cursorline' is set

LunarWatcher:
The fix introduced in #216 has been partly cancelled out by `override` now being dependent on whether or not cursorline is set. This is breaking for me, as it's currently impossible to tell whether or not I've correctly selected text if it's in a `LspDiagInline*` highlight group:

image

This is fixed by `override: false`, but #216 and what I assume the motivation for the new `override` value is, suggests doing this isn't universally desirable. Therefore, this PR instead introduces an option that allows `override` to be forced `false` for  the `LspDiagInline*` options even when `'cursorline'` is set. When `diagNoOverrideSyntaxHighlighting: true`:

image

With my particular theme, the errors are still visible with `'cursorline'` due to `gui=undercurl`:

image

Losing the background is not as bad as losing all visual reference when doing a selection in an error group. And when the variable is false (default), it works as it always has, and the selection is impossible to see when `'cursorline'` is set.

It'll likely still hide all error highlighting if `gui` isn't set, but that's a theme option, and as long as this isn't a default requirement, I see this as a perfectly acceptable tradeoff.

Also for the record, the theme I use set `LspDiagInlineError` to:
```
:hi LspDiagInlineError
LspDiagInlineError xxx cterm=underline gui=undercurl guifg=#444444 guibg=#ffafd7 guisp=#d70000
```


LunarWatcher:
Due to [AI slop in vim](https://hachyderm.io/@AndrewRadev/116175986749599825), I have switched to emacs and no longer care about this PR. I'll leave the fork up in case you want to cherry-pick the commit, but I won't be able to follow up on reviews, so I'm closing the PR

PR #753, opened 2026-01-29
Add possibility to overwrite the languageId that is sent to the server

jclsn:
I am putting this PR up for discussion:

I have tried to copy a feature that nvim-lsp has, which is overwriting the `languageId` that is sent to the server. See https://github.com/neovim/nvim-lspconfig/blob/238583bb00770b079c68c69a860d65e5d1d8acf9/lsp/docker_language_server.lua#L14

The docker language-server supports Dockerfile and docker-compose.yml. In case a yaml is detected, you need to send a different languageId to the server. This commit provides a way to customize the languageId via a function.

This is only needed though, because in Neovim they call their sub-filetypes `yaml.docker-compose`, `yaml.gitlab` etc. and then the wrong `languageId` would get sent to the server. You could however just do

```vim
au BufEnter,BufRead docker-compose.yaml,docker-compose.yml set filetype=dockercompse
au BufEnter,BufRead docker-compose.yaml,docker-compose.yml set syntax=yaml
```

and forget about it. You would then just have to add the `dockercompose` filetype also to the YAML language server, so it will still pick it up, which the Neovim people also still need to do: https://github.com/neovim/nvim-lspconfig/blob/238583bb00770b079c68c69a860d65e5d1d8acf9/doc/configs.md?plain=1#L14415

The advantage of a `yaml.docker-compose` filetype is merely

1. It still sources the yaml syntax file
2. It provides a good mental model of categorizing filetypes

I wonder if we should do this. I already made the patch, so we can add the feature, but it is not strictly needed for the functionality.

It would be used like this

```vim
vim9script

augroup dockercompose_yaml
    autocmd!
    autocmd BufEnter,BufReadPre docker-compose.yaml,docker-compose.yml setfiletype dockercompose.yaml
augroup END

def GetLanguageId(): string
    if expand('%:t') =~# 'docker-compose' && &filetype =~# 'yaml'
        return 'dockercompose'
    endif
    return &filetype
enddef

g:LspAddServer ([{
    name: 'docker-language-server',
    path: 'docker-language-server',
    args: ['start', '--stdio'],
    filetype: ['dockerfile', 'dockercompose.yaml'],
    debug: v:true,
    languageId: GetLanguageId,
}])
```

### DIFF
--- a/doc/lsp.txt
+++ b/doc/lsp.txt
@@ -362,6 +362,72 @@ Additionally the following configurations can be made:
 			encoded and decoded in every LSP message and will
 			incur some overhead.
 
+	languageId				*lsp-cfg-languageId*
+			(Optional) A possibility to override tha languageId
+			that is sent to the language server by specifying a
+			function or lambda for cases where the languageId is
+			not identical to the filetype. This could be useful
+			for server like:
+
+			https://github.com/docker/docker-language-server
+			https://github.com/alesbrelih/gitlab-ci-ls
+
+			Example:
+>
+		vim9script
+
+		augroup dockercompose-yaml
+		    autocmd!
+		    autocmd BufEnter,BufReadPre docker-compose.yaml,docker-compose.yml setfiletype dockercompose.yaml
+		augroup END
+
+		def GetLanguageId(): string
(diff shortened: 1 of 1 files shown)


yegappan:
Thanks

PR #754, opened 2026-01-30
fix: hardcoded log dir

slicklash:
### Problem

`/tmp` directory is not always present or writable e.g. on Termux `LspServer debug on` fails.

### Solution

Respect and use `$TMPDIR` when present.

### DIFF
--- a/autoload/lsp/util.vim
+++ b/autoload/lsp/util.vim
@@ -24,7 +24,7 @@ enddef
 # Lsp server trace log directory
 var lsp_log_dir: string
 if has('unix')
-  lsp_log_dir = '/tmp/'
+  lsp_log_dir = (exists('$TMPDIR') && !empty($TMPDIR) ? $TMPDIR : '/tmp')->fnamemodify(':p')
 else
   lsp_log_dir = $TEMP .. '\\'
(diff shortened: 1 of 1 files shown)


mmrwoods:
This is problematic for MacOS users as `$TMPDIR` is by default automatically set to something like `/var/folders/57/kqhq8g895tsbtx6j3nd7xqbw0000gn/T/`, and the LSP debug logs are no longer easily located.
 
This could do with being documented, and it would be nice if it was also configurable, e.g. `lspDebugLogDir: '/tmp'`

I'll open a PR if I get some time over the next few days, but if anyone wants to beat me to it... ;-)

mmrwoods:
I wonder if the original default behaviour of just using `/tmp/` should be restored, but a `lspDebugLogDir` option added, so anyone who wants it can just set `lspDebugLogDir: $TMPDIR`?

yegappan:
@mmrwoods I have committed 0df1c5327067c7420cc80fd937f4619e39e3c2df now to prefer the /tmp directory for storing the log messages over $TMPDIR.

mmrwoods:
Thanks :-)

PR #763, opened 2026-03-01
Do not attempt LSP response to request with str id

### LINKED ISSUE #761: Non-numerical request/response ids incorrectly converted to number (via str2nr)
The LSP server `pylsp` (for Python, and `lake serve` for Lean) sends LSP requests with non-numerical IDs.
Here are some examples from the debug log:
```
03/01/26 02:33:16: Received {"id":"8799aaa9-7ba9-4ffe-932a-2fca9bb83507","jsonrpc":"2.0","method":"window/workDoneProgress/create","params":{"token":"ea27c5ca-d2cd-4618-8ee4-4e56383e12c7"}}
03/01/26 03:18:29: Received {"id":"787ef272-ca6b-429b-82c5-90fc2a0216ea","jsonrpc":"2.0","method":"window/workDoneProgress/create","params":{"token":"d7120c8d-4d65-4e07-a8c7-c52cdf66f2f4"}}
03/01/26 03:18:30: Received {"id":"1c72c3e4-d128-465d-9cc9-275b795f7215","jsonrpc":"2.0","method":"window/workDoneProgress/create","params":{"token":"62fdb7d6-1bac-43c0-93e8-73b56128e227"}}
03/01/26 03:18:33: Received {"id":"f2574a79-fb83-4187-939c-026c0843907e","jsonrpc":"2.0","method":"window/workDoneProgress/create","params":{"token":"59908bd9-a390-4345-9fa4-05d0ae360605"}}
```

I believe the `string` type in the spec for `id`:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage
means literally a string, including non-numerical strings, can be an identifier, as an alternative to numbers.

The responses sent back by this plugin for the above are:
```
03/01/26 02:33:16: Sent {"id":8799,"jsonrpc":"2.0","result":null}
03/01/26 03:18:29: Sent {"id":787,"jsonrpc":"2.0","result":null}
03/01/26 03:18:30: Sent {"id":1,"jsonrpc":"2.0","result":null}
03/01/26 03:18:33: Sent {"id":0,"jsonrpc":"2.0","result":null}
```
which I think is due to using `str2nr` in this line:
https://github.com/yegappan/lsp/blob/0f7203e71680438056d3ba77c8c24b6fd36d8940/autoload/lsp/lspserver.vim#L336
These response id's are not correct, they should be the same string received by the plugin from the LSP server.

I suspect a fix is simply reverting back to
```
var resp: dict<any> = lspserver.createResponse(request.id)
```
which was changed in this PR
https://github.com/yegappan/lsp/pull/55/changes
based on the incorrect assumption that request/reponse ids should be numerical.

In the current main branch I see a pre-check
```
if (type(request.id) == ...
  ...
  return
endif
```
at
https://github.com/yegappan/lsp/blob/0f7203e71680438056d3ba77c8c24b6fd36d8940/autoload/lsp/lspserver.vim#L330
before the call to `lspserver.createResponse(` and that seems OK to me.

If you would like I can make and test this change locally and make a PR.

castedo:
- Stop sending responses with invalid integer ids due to string ids
- Calls to vim ch_sendexpr only support integer message ids https://github.com/vim/vim/issues/14091

Fixes #761

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -327,13 +327,17 @@ enddef
 
 # send a response message to the server
 def SendResponse(lspserver: dict<any>, request: dict<any>, result: any, error: dict<any>)
-  if (request.id->type() == v:t_string && request.id->trim()->empty())
-    || (request.id->type() != v:t_string && request.id->type() != v:t_number)
+  if request.id->type() == v:t_string
+    # give up on response due to https://github.com/vim/vim/issues/14091
+    if lspserver.debug
+      lspserver.traceLog($'Response for request with string id "{request.id}" not supported')
+    endif
+    return
+  elseif request.id->type() != v:t_number
     util.ErrMsg('request.id of response to LSP server must be a number or a string')
     return
   endif
-  var resp: dict<any> = lspserver.createResponse(
-	    request.id->type() == v:t_string ? request.id->str2nr() : request.id)
+  var resp: dict<any> = lspserver.createResponse(request.id)
   if error->empty()
     resp->extend({result: result})
   else


yegappan:
> I have not yet figured out how to run the unit tests (and maybe tweaks unit tests).
> 
> This PR makes a limited fix in the absence of resolving the following high-level question in my mind:
> 
> * Why does this section of vim9script code manage to get an id of type string but `ch_sendexpr` does not support string ids?
> 
> I image that there is _some_ vim function that can receive JSON messages with an `id` member with a JSON string value. So why isn't there a vim function that can _send_ a JSON message with an `id` member as a JSON string? Why can't this plugin use such a vim function?
> 
> Can vim9script send an LSP notification message with a string id? If so, can the same approach be used to send responses? Why is `ch_sendexpr` needed to match integer ids?

Thanks for the patch! Currently, Vim’s channel functions only support using numbers
as message identifiers; string identifiers aren't supported yet. We’ll need to enhance the
channel functions to support strings.

castedo:
After a quick view details of this failing GitHub action, it does not seem like it has anything to do with the change in this PR. Should I be attempting to reproduce this failure in a local environment? This seems like a CI unit test failure that you should look at.

PR #768, opened 2026-03-01
Respond to LSP requests with valid string id

### LINKED ISSUE #728: Asking for support of non-numeric IDs for language server protocol (LSP) channels
This is a follow up issue to the corresponding enhancement filed against Vim: https://github.com/vim/vim/issues/19003
After the Vim codebase gets (hopefully) enhanced, the Vim9 source code for this LSP client needs also be adapted.

castedo:
Fixes #728

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -306,7 +306,7 @@ def CreateRequest(lspserver: dict<any>, method: string): dict<any>
 enddef
 
 # create a LSP server response message
-def CreateResponse(lspserver: dict<any>, req_id: number): dict<any>
+def CreateResponse(lspserver: dict<any>, req_id: any): dict<any>
   var resp = {
     jsonrpc: '2.0',
     id: req_id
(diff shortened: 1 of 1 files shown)


castedo:
Some commentary:

This vim plugin does not need vim to manage an RPC response to an LSP response. This vim plugin just needs vim to send an LSP response as a message, nothing more. It seems to me that the handling of ids of an RPC request/response cycle by `ch_sendexpr` (or `ch_evalexpr`) is not needed when this vim plugin *receives* an LSP request and just needs to respond back.

Also of note is that serializing a JSONRPC message is trivial whereas parsing one is trickier.

Although it would be a nice feature for `ch_sendexpr` to do JSONRPC message serialization with string ids, it's really not a blocker. It seems more of a nicety given that `ch_sendraw` is available.

castedo:
Thank you for helping me stick to vim with simple LSP support and continue avoiding all the newfangled editors! :rofl:

PR #774, opened 2026-03-02
Don't highlight symbols with the same name as a title in the outline

arp242:
It would highlight identifiers that happen to be the same name as a title, such as String or or Event:

cap-2026-03-02T15:02:35_border

In Go especially String() is a very common method name.

To fix it, the titles are now inserted with a concealed @ at the end (Title@). I don't think there's any language which has these kind of symbols?


autoload/lsp/outline.vim
```diff
@@ -73,10 +73,10 @@ def AddSymbolText(bnr: number,
       lnumMap->extend([{}])
     endif
     if children
-      text->extend([prefix .. symType])
+      text->extend([prefix .. symType .. '@'])
       prefix ..= '  '
     else
-      text->extend([symType])
+      text->extend([symType .. '@'])
```
yegappan: Can you use string interpolation here `$"{symType}@"`?

Issue #767, opened 2026-03-01
Feature request: option to disable LSP diagnostics in insert mode


iam28th:
I found only an option to disable diagnostics completely - `features: {'diagnostics': v:false}` when adding LSP server.

Would be nicer to disable diagnostics in inset mode only, as otherwise there's a lot of noise about missing bracket or a semicolon and general errors caused by syntax when typing, at least in my experience with clangd. 

For example, vim.lsp provides:
[g:lsp_diagnostics_highlights_insert_mode_enabled](https://github.com/prabirshrestha/vim-lsp/blob/master/doc/vim-lsp.txt#L604)

iam28th:
I'll try to move these lines around....
https://github.com/yegappan/lsp/blob/main/autoload/lsp/diag.vim#L376C1-L379C1

PR #778, opened 2026-03-05
Check g:ale_lint_on_text_changed before SendAleDiags

iam28th:
Addresses https://github.com/yegappan/lsp/issues/767

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -377,7 +377,7 @@ export def ProcessNewDiags(bnr: number)
 
   var lspOpts = opt.lspOptions
   if lspOpts.aleSupport
-    if textChangedMode && !g:ale_lint_on_text_changed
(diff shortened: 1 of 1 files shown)


autoload/lsp/diag.vim
```diff
@@ -372,17 +372,23 @@ enddef
 export def ProcessNewDiags(bnr: number)
   DiagsUpdateLocList(bnr)
 
+  var curmode: string = mode()
+  var textChangedMode: bool = (curmode == 'i' || curmode == 'R' || curmode == 'Rv')
+
   var lspOpts = opt.lspOptions
   if lspOpts.aleSupport
-    SendAleDiags(bnr, -1)
+    if textChangedMode && !g:ale_lint_on_text_changed
```
yegappan: If the ale_lint_on_text_changed global variable is not set, then this will result in an error.
Instead can you use the following?

```
    if textChangedMode && !get(g:, 'ale_lint_on_text_changed', 0)
```

PR #790, opened 2026-04-01
Do not restore cursor after applying edits

cskr:
`ApplyTextEdits` [handles changes such that the cursor need not be restored explicitly](https://github.com/cskr/lsp/blob/main/autoload/lsp/textedit.vim#L178-L202). When `:LspFormat` tries to restore the cursor's position, it places it on the incorrect line. This is particularly jarring when `LspFormat` is configured to run automatically on `BufWritePre`, as it disrupts the editing flow.

### Steps to reproduce bug on main

1. Add `ruff server` as an LSP server for Python.
2. Open a Python file with the following contents and place the cursor on the last line.
```
def main():
    pass
def foo():
    pass
def bar():
    pass
```
3. Run `:LspFormat` 
4.  Observe that the cursor jumps from `bar()` to `foo()` when two blank lines are added after each function.

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -1277,9 +1277,7 @@ def TextDocFormat(lspserver: dict<any>, fname: string, rangeFormat: bool,
 
   # interface TextEdit
   # Apply each of the text edit operations
-  var save_cursor: list<number> = getcurpos()
   textedit.ApplyTextEdits(bnr, reply.result)
-  save_cursor->setpos('.')
 enddef
 
 def DecodeCallHierarchyItem(lspserver: dict<any>, item: dict<any>)


cskr:
These are the LSP-related lines in my .vimrc, in case it helps you reproduce the behaviour. I doubt this is relevant, but sharing anyway.

```
packadd lsp

var lspOptions = {
    noNewlineInCompletion: true,
    popupBorder: true,
}
g:LspOptionsSet(lspOptions)

var lspServers = [
    {
        name: 'rust-analyzer',
        filetype: 'rust',
        path: 'rust-analyzer',
    },
    {
        name: 'ty',
        filetype: 'python',
        path: 'ty',
        args: ['server'],
    },
    {
        name: 'ruff',
        filetype: 'python',
        path: 'ruff',
        args: ['server'],
    },
]
g:LspAddServer(lspServers)

def OnLspAttached()
    autocmd! BufWritePre <buffer> LspFormat

    nmap <buffer> <silent> gd <cmd>LspGotoDefinition<cr>
    nmap <buffer> <silent> gf <cmd>LspCodeAction<cr>
    nmap <buffer> <silent> gs <cmd>LspSymbolSearch<cr>
    nmap <buffer> <silent> gr <cmd>LspShowReferences<cr>
    nmap <buffer> <silent> gi <cmd>LspGotoImpl<cr>
    nmap <buffer> <silent> gt <cmd>LspGotoTypeDef<cr>
    nmap <buffer> <silent> <F2> <cmd>LspRename<cr>
    nmap <buffer> <silent> [g <cmd>LspDiag prev<cr>
    nmap <buffer> <silent> ]g <cmd>LspDiag next<cr>
    nmap <buffer> <silent> K <cmd>LspHover<cr>
enddef

autocmd! User LspAttached OnLspAttached()
```

yegappan:
The cursor position is still not being correctly restored after formatting. Even with this change, the cursor remains at the previous line/column coordinates rather than following the text it was originally on.

Examples:
1. Cursor on line 3 (`def foo`): After `:LspFormat`, the cursor stays on line 3, which is now an empty line.
2. Cursor on line 4 (`pass`): After formatting, the cursor stays on line 4, which is now an empty line.
3. Cursor on line 5 (`def bar`): After formatting, the cursor stays on line 5, which now contains `def foo`.

Do you see the same behavior?

cskr:
Yes, I see those behaviour. This change does not handle cases where new lines are added both above and on or below the line the cursor is on. It only helps when lines are modified entirely above the current line.

Handling the case where lines are added both above and below the current line requires a more involved change. I am trying a possible fix in [cskr/lsp@restore-cursor](https://github.com/cskr/lsp/tree/restore-cursor), but it is not fully functional yet.

I think there is value in fixing only the behaviour for line additions/removals before the current line. For example, LSPs tend to reorder and regroup imports, which changes lines at the top of the file.

cskr:
Any thoughts on whether it makes sense to merge this change, even though it addresses a narrow issue?

yegappan:
I tried implementing cursor restoration via marks and text properties, but both are lost when the LSP replaces the buffer text during formatting. I'll merge this as-is for now, acknowledging that we still need a more robust solution for cursor positioning after an edit.

PR #801, opened 2026-05-16
Add g:lsp_options to conveniently set opts from vimrc

mmrwoods:
## 📌 Summary

Adds `g:lsp_options` variable to allow convenient setting of plugin options without needing an autocmd in a user's vimrc

## 🧪 What This PR Improves

It seems unusual to require an autocmd to set configuration options for a plugin, so this adds a convenience global variable to allow options to be set in a vimrc without having to add an autocmd.

In my case this allows me to neatly split out plugin config in vimrc from plugin customisation in after/plugin, but in future there may be an equivalent of vim-lsp-settings for this plugin, and a user might not even need to configure servers directly, and therefore not need to set any autocmds themselves, just config options and maybe custom mappings.

## 🛠️ Changes Made

Check if `g:lsp_options` exists when loading plugin, if true automatically create `LspSetup` autocmd to call `LspOptionsSet()` with `g:lsp_options` as argument.

## 🧪 Testing

Tested locally removing my `LspSetup` autocmd to call `LspOptionsSet()` and replaced with `g:lsp_options` in vimrc. Tested with both vim9script and legacy script.  

## 📦 Breaking Changes

- [ ] Yes  
- [X] No  

## 📚 Documentation

- [ ] Yes  
- [X] No  

## ✔️ Checklist

- [ ] I have added tests or examples where appropriate.

### DIFF
--- a/plugin/lsp.vim
+++ b/plugin/lsp.vim
@@ -170,6 +170,11 @@ if has('gui_running')
   endif
 endif
 
+# Convenience option to avoid needing to add autocmd to vimrc
+if exists('g:lsp_options') && g:lsp_options->type() == v:t_dict
+  g:LspOptionsSet(g:lsp_options)
+endif
+
 # Invoke autocmd to register LSP servers and to set LSP options
 if exists('#User#LspSetup')
   :doautocmd <nomodeline> User LspSetup
--- a/doc/lsp.txt
+++ b/doc/lsp.txt
@@ -2013,6 +2013,22 @@ diagnostic messages, you can add the following line to your .vimrc file: >
 
 	call g:LspOptionsSet({'autoHighlightDiags': false})
 <
+                                                 *g:lsp_options*
+For convenience the options above can also be set using this global variable
+in your vimrc. The value must be a dictionary, and supports the same options
+as calling LspOptionsSet() directly. LspOptionsSet() is automatically called
+on plugin load when g:lsp_options is set. For example, in your vimrc, this: >
+
+    vim9script
+    g:lsp_options = {autoHighlightDiags: true}
+<
+Is equivalent to this: >
+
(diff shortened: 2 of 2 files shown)


plugin/lsp.vim
```diff
@@ -170,6 +170,11 @@ if has('gui_running')
   endif
 endif
 
+# Convenience option to avoid needing to add autocmd to vimrc
+if exists('g:lsp_options') && type(g:lsp_options) == 4
+  autocmd LSPAutoCmds User LspSetup g:LspOptionsSet(g:lsp_options)
```
yegappan: Instead of adding to the autocmd, why can't the options be directly set using:
```
g:LspOptionsSet(g:lsp_options)
```

plugin/lsp.vim
```diff
@@ -170,6 +170,11 @@ if has('gui_running')
   endif
 endif
 
+# Convenience option to avoid needing to add autocmd to vimrc
+if exists('g:lsp_options') && type(g:lsp_options) == 4
```
yegappan: Can you change this code to the following?
```
if exists('g:lsp_options') && g:lsp_options->type() == v:t_dict
    g:LspOptionsSet(g:lsp_options)
endif
```

yegappan (changes requested):

plugin/lsp.vim
mmrwoods: Honestly, I hadn't thought about it, yes, that would be neater, but it will also force the options autoload script to be loaded at plugin load time. This happens already, but it shouldn't have to and it would be nice to minimise the impact on vim startup time.

mmrwoods: Done, also updated docs to reflect this change

PR #804, opened 2026-05-18
Handle push diagnostics even if pull is enabled

### LINKED ISSUE #798: [Bug]: Diagnostics are not displayed with rust-analyzer
### Description

50d4b60 breaks diagnostics when used with rust-analyzer. When I open a `.rs` file with known errors, the errors are not highlighted in the buffer. The following error appears in the status line.

```
Error: request textDocument/diagnostic failed (server cancelled the request, error = ServerCancelled, data = {'retriggerRequest': true}) 
```

### Vim Version

Vim 9.1.1230

### Operating System

Debian 13

### Plugin Commit Hash

159de71

### Language Server

rust-analyzer

### Language Server Version

1.95.0

### Minimal Vim Configuration

```vim
var lspServers = [
    {
        name: 'rust-analyzer',
        filetype: 'rust',
        path: 'rust-analyzer',
    },
]
g:LspAddServer(lspServers)
```

### Steps to Reproduce

```text
1. Open a Rust source file with `use` lines for unknown crates.
2. Observe that no diagnostic/error is shown on the line with the incorrect import.
3. Wait for a few seconds and observe the following error on the status line.


Error: request textDocument/diagnostic failed (server cancelled the request, error = ServerCancelled, data = {'retriggerRequest': true})
```

### Expected Behavior

Diagnostics should be displayed automatically, without any error on the status line.

### Actual Behavior

Diagnostics are not shown and an error is displayed on the status line.

### Logs

```shell

```

### Final Checklist

- [x] I have verified this issue persists with a minimal configuration.
- [x] I am using the latest version of the plugin.
- [x] I am using a supported Vim version (9.0+).

cskr:
## 📌 Summary

Some LSP servers, like rust-analyzer, send diagnostic items through both push and pull channels. While rust-analyzer sends a disjoint set of diagnostics through the two channels, the LSP specification doesn't appear to mandate this. Hence the same diagnostic item may be sent through push and pull by some LSP servers.

This change handles pushed diagnostic items, even if the server supports pull. Additionally, it maintains separate lists for diagnostic items receive through push and pull. It deduplicates items across the two channels, using start line, start character, code, and message of the items.

---

## 🔍 Related Issues

Fixes #798

---

## 🧪 What This PR Improves

Before this change diagnostics sent through push were not visible, if the server supported `diagnosticProvider` capability.

---

## 🛠️ Changes Made

- Handle pushed diagnostic items, even if the server supports pull, by maintaining separate lists for push and pull.
- Deduplicate items across push and pull, using start line, start character, code, and message of the items.

---

## 🧪 Testing

With `rust-analyzer` configured as an LSP server, test the following scenarios.

### Push Diagnostics

Add the following lines in main.rs, write it with `:w`, and observe that a diagnostic item is displayed for missing import of `HashMap` on line #2.

```
fn main() {
    HashMap::new();
}
```

### Pull Diagnostics

Add the following lines in main.rs and observe that a diagnostic item is displayed for missing semicolon on line #2.

```
use std::collections::HashMap;

fn main() {
    let _map: HashMap<i64, i64> = HashMap::new()
    println!("Hello, world!");
}
```

---

## 📦 Breaking Changes

- [ ] Yes  
- [x] No 

---

## 📚 Documentation

- [ ] Yes  
- [x] No  

If yes, describe what needs updating.

---

## ✔️ Checklist

- [x] I have tested this with a minimal Vim9script configuration.
- [x] I have run the plugin against the relevant LSP server(s).
- [ ] I have updated documentation if needed.
- [x] I have followed the project’s coding style and conventions.
- [ ] I have added tests or examples where appropriate.

### DIFF
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -166,7 +166,12 @@ def DeduplicateDiags(diags: list<dict<any>>): list<dict<any>>
   var result = []
   var seen = {}
   for d in diags
-    var key = $"{d.range.start.line}:{d.range.start.character}:{d->get('code', '')->string()}:{d.message}"
+    var key = string([
+      d.range.start.line,
+      d.range.start.character,
+      d->get('code', ''),
+      d.message
+    ])
     if !seen->has_key(key)
       seen[key] = true
       result->add(d)
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -440,7 +440,18 @@ export def DiagNotification(lspserver: dict<any>, uri: string, diags_arg: list<d
   endif
   var bnr: number = fname->bufnr()
 
-  var newDiags: list<dict<any>> = diags_arg->slice(0, opt.lspOptions.maxDiagnostics)
+  var serverId = lspserver.id
+  var serverDiags: dict<dict<list<any>>> = diagsMap->has_key(bnr) ?
+      diagsMap[bnr].serverDiagnostics : {}
+  var kindDiags: dict<list<any>> = serverDiags->has_key(serverId) ?
+      serverDiags[serverId] : {}
+
+  # Count existing diagnostics from this server for other sync kinds
+  var otherDiagCount = kindDiags
+      ->keys()
+      ->filter((_, k) => k != sync_kind)
+      ->reduce((acc, k) => acc + kindDiags[k]->len(), 0)
+  var newDiags: list<dict<any>> = diags_arg->slice(0, opt.lspOptions.maxDiagnostics - otherDiagCount)
 
   if lspserver.needOffsetEncoding
     # Decode the position encoding in all the diags
@@ -456,12 +467,6 @@ export def DiagNotification(lspserver: dict<any>, uri: string, diags_arg: list<d
 
   # TODO: Is the buffer (bnr) always a loaded buffer? Should we load it here?
   var lastlnum: number = bnr->getbufinfo()[0].linecount
-
-  var serverId = lspserver.id
-  var serverDiags: dict<dict<list<any>>> = diagsMap->has_key(bnr) ?
-      diagsMap[bnr].serverDiagnostics : {}
-  var kindDiags: dict<list<any>> = serverDiags->has_key(serverId) ?
-      serverDiags[serverId] : {}
   kindDiags[sync_kind] = newDiags
   serverDiags[serverId] = kindDiags
 
--- a/autoload/lsp/diag.vim
+++ b/autoload/lsp/diag.vim
@@ -8,8 +8,14 @@ import './util.vim'
 
 # [bnr] = {
 #   serverDiagnostics: {
-#     lspServer1Id: [diag, diag, diag]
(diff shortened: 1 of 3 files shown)


autoload/lsp/diag.vim
```diff
@@ -154,6 +160,24 @@ def SortDiags(diags: list<dict<any>>): list<dict<any>>
   return diags->sort(DiagsSortFunc)
 enddef
 
+# Deduplicate diagnostics, if the same dignostic is sent in
+# both push and pull channels
+def DeduplicateDiags(diags: list<dict<any>>): list<dict<any>>
+  var result = []
+  var seen = {}
+  for d in diags
+    var key = printf('%d:%d:%s:%s',
```
yegappan: Thanks for the patch.

Calling printf() to generate the key has higher overhead than using string conversion and concatenating the values.  Will the following work?
```
var key = $"{d.range.start.line}:{d.range.start.character}:{d->get('code', '')->string}:{d.message}"
```

autoload/lsp/diag.vim
```diff
@@ -401,7 +425,7 @@ enddef
 # process a diagnostic notification message from the LSP server
 # Notification: textDocument/publishDiagnostics
 # Param: PublishDiagnosticsParams
-export def DiagNotification(lspserver: dict<any>, uri: string, diags_arg: list<dict<any>>): void
+export def DiagNotification(lspserver: dict<any>, uri: string, diags_arg: list<dict<any>>, channel: string): void
```
yegappan: The name `channel` may refer to the Vim channel feature.  To avoid confusion, can you use `sync_kind` for the argument name instead of `channel`?

yegappan (changes requested):

cskr:
Thanks for the review. I have pushed a commit to address the comments.

yegappan:
Thanks for addressing the review comments.  Can you incorporate the suggestions 1 and 2 from the latest Copilot review? Thanks

cskr:
I have addressed the suggestion for key construction and the missed deduplication before collecting into `sortedDiagnostics`.

The suggestion for application of `maxDiagnostics` needs more thought. The documentation says, "Maximum number of diagnostics to process when receiving diagnostic notifications from the server.". Do we want to limit the number of diagnostics we process or the number of diagnostics displayed to the user?

yegappan:
> I have addressed the suggestion for key construction and the missed deduplication before collecting into `sortedDiagnostics`.
> 

Thanks.

> The suggestion for application of `maxDiagnostics` needs more thought. The documentation says, "Maximum number of diagnostics to process when receiving diagnostic notifications from the server.". Do we want to limit the number of diagnostics we process or the number of diagnostics displayed to the user?

The maximum number of diagnostics displayed should be limited. But if the diagnostics are not going to be displayed to the user, then there is no need to process them either. So the `maxDiagnostics` setting controls the number of diagnostics that will be processed.

cskr:
Updated to account for diagnostics from other sync kind, before limiting the number of diagnostics to be processed.

yegappan:
Thanks for addressing the review comments.  Looks good to me.

PR #807, opened 2026-05-23
Backwards compatibility for async RPC callback

mmrwoods:
The signature for the callback function was changed to require an error argument in commit f396fe75b7e7be4cc3e703641905c90d41c767b4, but this is a breaking change not reflected in the lsp-custom-requests documentation

This restores compatibility with plugins like fuzzbox-lsp.vim and scope.vim, and anyone who has followed the docs for custom requests.

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -505,7 +505,16 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
   try
     RpcCb(lspserver, result, error)
   catch
-    lspserver.errorLog($'Callback for {method} raised exception: {v:exception}')
+    # backwards compatibility for old callback signature
+    if v:exception =~# '\v:(E118):' && v:throwpoint =~# 'AsyncRpcCb'
+      try
+        RpcCb(lspserver, result)
+      catch
+        lspserver.errorLog($'Callback for {method} raised exception: {v:exception}')
+      endtry
+    else
+      lspserver.errorLog($'Callback for {method} raised exception: {v:exception}')
+    endif
   endtry
 enddef
 
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -506,7 +506,7 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
     RpcCb(lspserver, result, error)
   catch
     # backwards compatibility for old callback signature
-    if v:exception =~# '\v:(E118):' && v:throwpoint =~# 'AsyncRpcCb'
+    if v:exception =~# '\v:(E118):' && v:throwpoint =~# 'AsyncRpcCb,\s\+line\s\+\d'
       try
         RpcCb(lspserver, result)
       catch


autoload/lsp/lspserver.vim
```diff
@@ -504,8 +504,13 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
   # Pass both result AND error to callback
   try
     RpcCb(lspserver, result, error)
-  catch
-    lspserver.errorLog($'Callback for {method} raised exception: {v:exception}')
+  catch /\v:(E118):/
```
yegappan: Thanks for the patch.

This will not catch other exceptions that may occur when the RPC callback function is invoked.
Can you modify this to catch all the exceptions and then check for E118 specifically below?

mmrwoods: Ah yes, sorry, I just threw this together quickly when I identified the problem, will update PR now

autoload/lsp/lspserver.vim
```diff
@@ -504,6 +504,17 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
   # Pass both result AND error to callback
   try
     RpcCb(lspserver, result, error)
+  catch /\v:(E118):/
```
yegappan: Can you remove the error message pattern from this catch statement?

autoload/lsp/lspserver.vim
```diff
@@ -504,6 +504,17 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
...
+    # backwards compatibility for old callback signature
+    if v:throwpoint =~ '\<AsyncRpcCb\>'
```
yegappan: Can you change this to `v:exception =~ "E118:"`?

mmrwoods: Are you expecting a conditional statement in a catch-all try/catch to work here? I don't think it will.

mmrwoods: Something like this:

```vim
  try
    RpcCb(lspserver, result, error)
  catch
    # backwards compatibility for old callback signature
    if v:exception =~# '\v:(E118):'
      try
        RpcCb(lspserver, result)
      catch
        lspserver.errorLog($'Callback for {method} raised exception: {v:exception}')
      endtry
    endif
    lspserver.errorLog($'Callback for {method} raised exception: {v:exception}')
  endtry
```

???

This will always log an error when falling back to the old signature, is that desired?

yegappan: You can move the second errorLog() call to an `else` statement:

```
  try
    RpcCb(lspserver, result, error)
  catch
    # backwards compatibility for old callback signature
    if v:exception =~# '\v:(E118):'
      try
        RpcCb(lspserver, result)
      catch
        lspserver.errorLog($'Callback for {method} raised exception: {v:exception}')
      endtry
    else
       lspserver.errorLog($'Callback for {method} raised exception: {v:exception}')
    endif
  endtry
```

mmrwoods: Yeah, sorry, got that in my latest commit, seems to work well

mmrwoods: I think I still need to check for the throwpoint though, I'll do some more testing

mmrwoods: I'll just ask Copilot :-)

PR #809, opened 2026-05-27
Improve backwards compat for async RPC callback

mmrwoods:
## 📌 Summary

Use `v:stacktrace` if available to check call site when handling "Too many arguments" error and support previously async RPC callback function signature.

---

## 🔍 Related Issues

- Related to #807

---

## 🧪 What This PR Improves

More reliably checks that the "Too many arguments" error occurred within the LSP server code rather than in the callback code itself when falling back to using the old function signature.

---

## 🛠️ Changes Made

- If `v:stacktrace` exists, confirm that the current script path matches the last filepath in the stacktrace before falling back to the old callback function signature
- If `v:stacktrace` does not exist, continue to pattern match on `v:throwpoint`, as before

---

## 🧪 Testing

Manually tested locally with fuzzbox-lsp.vim

---

## 📦 Breaking Changes

- [ ] Yes  
- [x] No  

If yes, explain:

---

## 📚 Documentation

- [ ] Yes  
- [X] No  

---

## ✔️ Checklist

- [X] I have added tests or examples where appropriate.

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -473,6 +473,13 @@ def Rpc(lspserver: dict<any>, method: string, params: any, opts: dict<any> = {})
   return {}
 enddef
 
+# legacy script function to get v:stacktrace to avoid function compile errors on
+# Vim versions that do not include it (only added to Vim in version 9.1.0984)
+# get(v:, 'stacktrace') in a vim9 function throws E1075: Namespace not supported: v:, 'stacktrace'
+function GetStackTrace()
+  return get(v:, 'stacktrace')
+endfunction
+
 # LSP server asynchronous RPC callback
 def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel, reply: dict<any>)
   if lspserver.debug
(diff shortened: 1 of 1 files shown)


mmrwoods:
Fails in vim 9.0.000, will investigate.

EDIT: function compile failure, should have thought of that, will see if it's worth a workaround

mmrwoods:
I'm not sure there is any way around this, I tried using `get(v:, 'stacktrace')` which I expected to work fine, and does work fine in a vim9script at script scope, but does not work inside a function, e.g.

This works fine:
```vim
echow get(v:, 'stacktrace')
```

But this throws error "E1075: Namespace not supported: v:, 'stacktrace')"
```vim
def Foo()
  echow get(v:, 'stacktrace')
enddef
Foo()
```

mmrwoods:
Fixed using legacy function to get `v:stacktrace`, it's not pretty, but it should be reliable

mmrwoods:
This requires further work to avoid falling back to the old throwpoint check if the filepath in the last item in the stack is not the current script. Honestly, I'm just not sure it's worth the mess in the code to support it.

PR #810, opened 2026-05-28
Improve backwards compat for async RPC callback

mmrwoods:
## 📌 Summary

Use `v:stacktrace` if available to check call site when handling "Too many arguments" error and support previously async RPC callback function signature.

---

## 🔍 Related Issues

- Related to #807

---

## 🧪 What This PR Improves

More reliably checks that the "Too many arguments" error occurred within the LSP server code rather than in the callback code itself when falling back to using the old function signature.

---

## 🛠️ Changes Made

- If `v:stacktrace` exists, confirm that the current script path matches the last filepath in the stacktrace before falling back to the old callback function signature
- If `v:stacktrace` does not exist, continue to pattern match on `v:throwpoint`, as before

---

## 🧪 Testing

Manually tested locally with fuzzbox-lsp.vim using Vim 9.0.0000 and 9.2.0421

---

## 📦 Breaking Changes

- [ ] Yes  
- [x] No  

---

## 📚 Documentation

- [ ] Yes  
- [X] No  

---

## ✔️ Checklist

- [X] I have added tests or examples where appropriate.

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -506,7 +506,15 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
     RpcCb(lspserver, result, error)
   catch
     # backwards compatibility for old callback signature
-    if v:exception =~# '\v:(E118):' && v:throwpoint =~# 'AsyncRpcCb,\s\+line\s\+\d'
+    var retry: bool = false
+    if v:exception =~# '\v:(E118):'
+      if exists_compiled('v:stacktrace')
+	retry = expand('<script>:p') == v:stacktrace[-1]['filepath']
+      else
+	retry = v:throwpoint =~# 'AsyncRpcCb,\s\+line\s\+\d'
+      endif
+    endif
+    if retry
       try
         RpcCb(lspserver, result)
       catch


mmrwoods:
@yegappan I decided to re-open this (well, create a new issue as I don't have permission to re-open) after having a second look this morning. The code does add complexity, so I am not fully convinced it is worth it for the likely small number of users or collaborating plugins affected, but I'll let you be the judge.

autoload/lsp/lspserver.vim
```diff
@@ -506,7 +513,15 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
     RpcCb(lspserver, result, error)
   catch
     # backwards compatibility for old callback signature
-    if v:exception =~# '\v:(E118):' && v:throwpoint =~# 'AsyncRpcCb,\s\+line\s\+\d'
+    var retry: bool
```
yegappan: Can you initialize this to false?

autoload/lsp/lspserver.vim
```diff
@@ -506,7 +513,15 @@ def AsyncRpcCb(lspserver: dict<any>, method: string, RpcCb: func, chan: channel,
...
+    if v:exception =~# '\v:(E118):'
+      if exists('v:stacktrace')
```
yegappan: The existence of the v:stacktrace variable is checked here and then the get() function is used in GetStackTraceVar() to get the variable value again.  Instead can you use the following here?

```
var strace: list<dict<any>> = get(v:, 'stacktrace', [])
if strace->empty()
   retry = v:throwpoint =~# 'AsyncRpcCb,\s\+line\s\+\d
else
   retry = expand('<script>:p') == strace[-1]['filepath']
endif
```

autoload/lsp/lspserver.vim
mmrwoods: Will do

autoload/lsp/lspserver.vim
mmrwoods: This won't work, it will result in error "E1075: Namespace not supported: v:, 'stacktrace'"

mmrwoods: You can confirm this quickly by creating any compiled vim9 function that tries to `get(v:, 'stacktrace')`

This throws E1075:

```vim
def Foo()
  echow get(v:, 'stacktrace')
enddef
Foo()
```

This works fine (code at script scope):
```vim
echow get(v:, 'stacktrace')
```

In Vim 9.2.0421 anyway

mmrwoods: Note: just tested Vim 9.0.000, same error with `get(v:, 'stacktrace')` inside a vim9 function, but works fine at script scope.

yegappan: I just realized Vim currently doesn't support namespaces other than g: in a def function.  Can you use this instead?

```
if exists_compiled('v:stacktrace')
   retry = expand('<script>:p') == v:stacktrace[-1]['filepath']
else
   retry = v:throwpoint =~# 'AsyncRpcCb,\s\+line\s\+\d
endif
```

mmrwoods: Fab! I didn't know about `exists_compiled()`, works perfectly in both Vim 9.0.000 and 9.2.0421 👍

mmrwoods:
@yegappan Commits squashed and rebased on latest `main`

PR #814, opened 2026-06-14
fix(diag): diagnostic messages shouldn't have CR

mao-yining:
Especially in virtual text and quickfix window.

## 📌 Summary

Previously, things with '\n' will also being shown in location list and virtual text. Which is ugly and unhelpful. Other lsp plugins like nvim also didn't show it in virtual text.
图片

Now it only show one single line.
图片

Still show complete message in `:LspDiagCurrent`
图片

## 🧪 Testing

I didn't know whether this is the right behaviour.
---

## 📦 Breaking Changes

- [x] Yes


mao-yining:
BTW I think the interface of `:LspDiagCurrent` can be improve. This is what nvim provide:

图片

@yegappan if you think it is good. Maybe I can implement it.

PR #827, opened 2026-07-28
fix: some var errs

Shane-XB-Qian:
## 📌 Summary

somehow these 2 var in lsp.vim would popup errs, looks one is var was deleted, another was not found,
not sure if was a vim9 script itself bug, but for now just trying to workaround it.

## 🧪 What This PR Improves

suspend the errs.

### DIFF
--- a/autoload/lsp/lsp.vim
+++ b/autoload/lsp/lsp.vim
@@ -116,7 +116,8 @@ enddef
 # score, LSP servers with the same score are being returned.
 # Returns an empty list if the servers is not found.
 def LspGetServers(bnr: number, ftype: string): list<dict<any>>
-  if !ftypeServerMap->has_key(ftype)
+  # E1302 var maybe deleted already
+  if !exists('ftypeServerMap') || !ftypeServerMap->has_key(ftype)
     return []
   endif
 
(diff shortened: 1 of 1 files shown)


yegappan:
Thanks for the patch.  I am not able to reproduce this issue though.  Are you still seeing these errors with the latest version of the plugin and the latest version of Vim?

Shane-XB-Qian:
> Are you still seeing these errors with the latest version of the plugin and the latest version of Vim?

yes, v9.2.1046 the latest today i still saw this err;
or anyway, kept this open to whom encountered this issue to workaround by himself

-- 
shane.xb.qian

Shane-XB-Qian:
Or in code level, in `AddFile()`,
// i have not tried to workaround this, since no sense to me,

do you know why this e1302 (var was deleted) possibly happened?

```vim
  for lspserver in lspservers
    if lspserver.stoppedByUser
      continue
    endif

    if attachedServerIds->index(lspserver.id) != -1
      # This server is already attached to the buffer
      continue
    endif

    if !lspserver.running     # <~~~ Here, sometimes e1302 <~~~
      if !lspInitializedOnce
        LspInitOnce()
      endif
      lspserver.startServer(bnr)
    endif

    buf.BufLspServerSet(bnr, lspserver)
    attachedServers->add(lspserver)
  endfor
```

-- 
shane.xb.qian

Shane-XB-Qian:
deepcopy can workaround this e1302 err, but logic looks going to be wrong if so.

-- 
shane.xb.qian

PR #829, opened 2026-07-30
feat(tagfunc): support 'kind'

### LINKED ISSUE #824: [Feature]: Support "kind" in tagfunc
### Problem description

Support kind in tagfunc

### Proposed solution

```diff
diff --git a/autoload/lsp/completion.vim b/autoload/lsp/completion.vim
index 8f2b5ad..518db7e 100644
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -53,7 +53,7 @@ export def OmniComplSet(ftype: string, enabled: bool)
 enddef
 
 # Map LSP complete item kind to a character
-def LspCompleteItemKindChar(kind: number): string
+export def LspCompleteItemKindChar(kind: number): string
   var kindMap: list<string> = [
     '',
     'Text',
diff --git a/autoload/lsp/lspserver.vim b/autoload/lsp/lspserver.vim
index 2d57443..1b1db5f 100644
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -2379,6 +2379,9 @@ def TagFunc(lspserver: dict<any>, pat: string, flags: string, info: dict<any>):
 	continue
       endif
       taglocations->add(sym.location)
+      if sym->has_key('kind')
+	taglocations[-1]['kind'] = completion.LspCompleteItemKindChar(sym.kind)
+      endif
     endfor
 
     if taglocations->empty()
diff --git a/autoload/lsp/symbol.vim b/autoload/lsp/symbol.vim
index a9c6c70..bbb4859 100644
--- a/autoload/lsp/symbol.vim
+++ b/autoload/lsp/symbol.vim
@@ -567,7 +567,11 @@ export def TagFunc(lspserver: dict<any>,
     var bnr = util.LspUriToBufnr(uri)
     var rstart = range.start
     var startByteIdx = util.GetLineByteFromPos(bnr, rstart)
-    tagitem.cmd = $"/\\%{rstart.line + 1}l\\%{startByteIdx + 1}c"
+    tagitem.cmd = $"/\\%{rstart.line + 1}l\\%{startByteIdx + 1}c/"
+
+    if tagloc->has_key('kind')
+      tagitem.kind = tagloc.kind
+    endif
 
     retval->add(tagitem)
   endfor
```

### Alternatives considered

```text

```

### Example usage (Vim9script)

```vim
vim9script
command! -nargs=_ -complete=customlist,TagComplete Tag tag <args>
def TagComplete(arglead: string, _, _): list<dict<any>>
	if empty(arglead)
		return null_list
	else
		return arglead->taglist(expand('%:p'))->map((_, v) => ({
			word: v.name, kind: v.kind, menu: v.filename
		}))
	endif
enddef
```

### Related Issues or Discussions

_No response_

### Additional context

What's more the current tagfunc only return name is equal to pattern. Witch is not good for searching.

### Final Checklist

- [ ] I have searched existing issues to avoid duplicates.
- [ ] I have explained why this feature is important.
- [ ] I have provided enough detail for maintainers to evaluate the request.

mao-yining:
Support 'kind' for tagunc

closes #824

### DIFF
--- a/autoload/lsp/lspserver.vim
+++ b/autoload/lsp/lspserver.vim
@@ -2378,6 +2378,9 @@ def TagFunc(lspserver: dict<any>, pat: string, flags: string, info: dict<any>): 
       if sym->get('name', '') != pat
 	continue
       endif
+      if sym->has_key('kind')
+	sym.kind = completion.LspCompleteItemKindChar(sym.kind)
+      endif
       taglocations->add(sym.location)
     endfor
 
--- a/autoload/lsp/symbol.vim
+++ b/autoload/lsp/symbol.vim
@@ -569,6 +569,10 @@ export def TagFunc(lspserver: dict<any>,
     var startByteIdx = util.GetLineByteFromPos(bnr, rstart)
     tagitem.cmd = $"/\\%{rstart.line + 1}l\\%{startByteIdx + 1}c/"
 
+    if tagloc->has_key('kind')
+      tagitem.kind = tagloc.kind
+    endif
+
     retval->add(tagitem)
   endfor
 
--- a/autoload/lsp/completion.vim
+++ b/autoload/lsp/completion.vim
@@ -53,7 +53,7 @@ export def OmniComplSet(ftype: string, enabled: bool)
 enddef
 
 # Map LSP complete item kind to a character
-def LspCompleteItemKindChar(kind: number): string
+export def LspCompleteItemKindChar(kind: number): string
   var kindMap: list<string> = [
     '',
     'Text',


autoload/lsp/symbol.vim
```diff
@@ -569,6 +569,10 @@ export def TagFunc(lspserver: dict<any>,
     var startByteIdx = util.GetLineByteFromPos(bnr, rstart)
     tagitem.cmd = $"/\\%{rstart.line + 1}l\\%{startByteIdx + 1}c/"
 
+    if tagloc->has_key('kind')
+      tagitem.kind = tagloc.kind
```
yegappan, line 573: Can you change this to the following?
```
   tagitem.kind = tagloc->get('kind', '')
```

autoload/lsp/lspserver.vim
```diff
@@ -2378,6 +2378,9 @@ def TagFunc(lspserver: dict<any>, pat: string, flags: string, info: dict<any>):
       if sym->get('name', '') != pat
 	continue
       endif
+      if sym->has_key('kind')
+	sym.kind = completion.LspCompleteItemKindChar(sym.kind)
```
yegappan, line 2382: Have you tested the returned tag kind and see whether it makes any difference?

autoload/lsp/symbol.vim
mao-yining: Emm If we do that, all item will have 'kind', which require more process for `tagfunc`. Or you mean has_key() is costly and get() is much faster?

autoload/lsp/lspserver.vim
mao-yining: 'kind' is number originally. Use `LspCompleteItemKindChar` to change it into char.

autoload/lsp/symbol.vim
yegappan: In general, for optimization reasons, the plugin uses the get() function with a default value to retrieve the value of a key in a Dict instead of first checking whether the key is present using has_key().  But in this case, if the kind key is not present, then an empty kind string will be used.  This may confuse the tag processing code in Vim.  You can leave the has_key() check in this case.

mao-yining: For current vim core of `tagfunc` if the return value has extra values, it will do a lot of things for 'kind' and check other things.

However, if we assume that LSP will return values with 'kind' in a larger probability. Maybe use `get()` is a better solution.

PR #835, opened 2026-08-12
refactor: use tr() instead of substitute()

mao-yining:
Some performance enhancements, no logic changes.


yegappan:
Thanks for the patch.

